PR #3, opened 2015-09-23
Added support for JsonHttpLogger

whiskeysierra:


src/main/java/org/zalando/springframework/web/logging/DefaultHttpLogger.java
```diff
@@ -28,7 +28,7 @@
 import java.util.List;
 import java.util.Map;
 
-public class DefaultHttpLogger implements HttpLogger {
+public final class DefaultHttpLogger implements HttpLogger {
```
lukasniemeier-zalando: Why final?

whiskeysierra: > Why final?

Because it's not intended to be subclassed, is it? If somebody needs stuff from it he/she is free to use composition.

lukasniemeier-zalando: You never know what crazy things clients of the library want to do, maybe they want to extend this class - so why prevent it?

jhorstmann: In that case they should open a pull request and give a good reason for the change :)

Issue #5, opened 2015-09-28
Split up formatting and writing


whiskeysierra:
Right now the `HttpLogger` implementations do both, formatting the request/response data into some kind of text format and they also choose a concrete target for the data, usually a Slf4j Logger.

If we split those responsibilities up into two interfaces we could more easily add support for one of those concerns without changing the other.

As correctly mentioned by @lukasniemeier-zalando this could also mean we extract the condition (`shouldLog`) from the interface.

lukasniemeier-zalando:
:+1: maybe we should take the chance and easen the creation of the creation of a fully configured ``LoggingFilter``

Issue #4, opened 2015-09-26
Get rid of Spring dependency


whiskeysierra:
Right now the only reason to have Spring as a direct dependency is that we use the `OncePerRequestFilter` which in turn is just an implementation detail.

Other than that we are just providing a servlet filter that, in theory, would work with any other framework just as good, as long as it's running in a servlet container.

That would mean we change the scope (and the name) to something more generic, e.g `servlet-logging`. Which got me thinking - *we're probably not the first ones to do something like this*.

I've found:

- http://logback.qos.ch/access.html
- https://github.com/librucha/servlet-logging-filter
- http://stackoverflow.com/questions/8933054/how-to-log-response-content-from-a-java-web-server
- ...

Most of them are either just a proof of concept or too tightly coupled to a certain logging framework. But I think we can steal some ideas from them.

For example logback-access by default logs full requests/responses like this:

```
GET /logback-demo/index.jsp HTTP/1.1
Host: localhost:8080
User-Agent: Mozilla/5.0 (Windows; U; Windows NT 5.1; en-US; rv:1.8) Gecko/20070312 Firefox/1.5.0
Accept: text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8
Accept-Language: en-us,en;q=0.5
Accept-Encoding: gzip,deflate
Accept-Charset: ISO-8859-1,utf-8;q=0.7,*;q=0.7
Keep-Alive: 300
Connection: keep-alive
Referer: http://localhost:8080/logback-demo/login.jsp
Cookie: JSESSIONID=15c7tqi9ehlwk;  OID324nkzcmr=null; OID32862zgoa=null; 



HTTP/1.1 200 OK
Content-Type: text/html; charset=iso-8859-1
Expires: Thu, 01 Jan 1970 00:00:00 GMT
Set-Cookie: JSESSIONID=bgebt99ce9om;path=/logback-demo


<html>
<head>
        <LINK REL=StyleSheet HREF="css/pk.css" />
</head>
<body>

<h2>Logback demo center</h2>

[snip, so that text is reasonably sized]
```

If you combine this with color-coded console output, one could e.g. produce log output that looks like [httpie](https://github.com/jkbrzt/httpie#httpie-a-cli-curl-like-tool-for-humans) which would be pretty cool for local development.

whiskeysierra:
What about *Logbook* as a project name?

lukasniemeier-zalando:
[...]

Issue #2, opened 2015-09-22
Servlet 3 asynchronous requests not correctly handled


c0nscience:
Hi,

we just tried your logging filter implementation, but it breaks all of our requests. Because we use exclusively servlet 3 async requests in our application, every request mapping method returns a callable.

Example method declaration:
```java
public Callable<ResponseEntity<Entity>> getEntity(@RequestBody final dto)
```

The response is always missing the content type and body, with the logging filter enabled.

Any suggestions how to fix this problem?

lukasniemeier-zalando:
I haven't had the chance to look into a filter chain with async requests. Could you provide a failing test case?

c0nscience:
Thanks @lukasniemeier-zalando for the quick response.

I just created a small example project with a simple rest endpoint and the logging filter configured.
Have a look at https://github.com/c0nscience/async-logging-test

The junit test though succeeds, but you can see in the logs that the httpLogger of the filter does not log a content type nor the body. If you run the example with the filter enabled the response is empty and if disabled the response is correct.

Sorry for the form of the example, I didn't know a faster and easier solution.
I hope it helps a bit.

lukasniemeier-zalando:
Thank's for the test case. I think @whiskeysierra already tried out something promising :clock1:

rzimmer:
Uh, that's great! Really looking forward to have async responses having logged! :+1:

whiskeysierra:
@rzimmer @c0nscience Can you try with `org.zalando:logbook-servlet:0.1.0`? We completely refactored the library and added support for async dispatch in the meantime.

rzimmer:
We had the chance to take a deeper look at the refactored library yesterday and so far, it looks really good. Async responses are logged correctly now with body messages. Until now, nothing seems broken.

Thanks for your effort, your quick reaction time and your greatly designed library, makes a lot of fun!

whiskeysierra:
Glad to hear that it works for you.

We still have some TODOs on our wishlist for this, e.g. #17 (special setup for secure services) and I'm not 100% happy with the current separation of formatting and writing, but nothing concrete, more like a gut feeling.

PR #6, opened 2015-09-29
Refactored library completely

whiskeysierra:
- got rid of Spring dependency (see #4)
- fixed async dispatch support (see #2)
- refactored API to split up formatting and writing (see #5)

I renamed the project to *Logbook* in the process. Still missing:

- support for obfuscators (headers, parameters, body)
- ~50% code coverage
- logo for readme


lukasniemeier-zalando:
* I'd stick with obfuscate the body as a String
* You need to insert you as a (or the) maintainer :stuck_out_tongue:

whiskeysierra:
> I'd stick with obfuscate the body as a String

Absolutely, I'd just think about having a different interface for the body (without the *fake* key):

    String obfuscate(final String body);

vs.

    String obfuscate(final String key, final String value);

whiskeysierra:
> I'd just think about having a different interface for the body

Otherwise I have the fear that clients, without reading the docs properly, would just look at the method having a key and a value and would assume that a body obfuscator would get called for every property + value (e.g. for json).

lukasniemeier-zalando:
:+1: obfuscation of the body will surely be also very different from obfuscating a header/parameter implementation-wise

whiskeysierra:
Any idea whether this is safe to use as a wallpaper?
https://en.wikipedia.org/wiki/File:Grand_Turk(34).jpg

src/main/java/org/zalando/logbook/Markable.java
```diff
@@ -0,0 +1,39 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+
+import javax.servlet.ServletRequest;
+
+interface Markable extends Named {
```
jhorstmann: I don't see the need for having this as an interface with default methods instead of protected methods in `OnceFilter`. Default methods are nice if a subclass might have a more optimized implementation, but here it just makes the code harder to follow.

src/main/java/org/zalando/logbook/Named.java
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+
+interface Named {
+
+    default String getName() {
+        // TODO this will use the system hash (hopefully), is this correct?
```
jhorstmann: As long as the class implementing this interface does not override `toString`. Maybe you want something like

    getClass().getName() + "@" + Integer.toHexString(System.identityHashCode(this))

Again, I would prefer to put this in a method of `OnceFilter`, this should really be a hidden implementation detail of the filter.

src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public String format(final TeeHttpServletRequest request) throws IOException {
+        final ImmutableMap<String, Object> content = ImmutableMap.<String, Object>builder()
+                .put("sender", request.getRemoteAddr())
+                .put("method", request.getMethod())
+                .put("path", request.getRequestURI())
+                .put("headers", getHeaders(request))
+                .put("params", request.getParameterMap())
+                .put("body", request.getBodyAsString())
```
jhorstmann: I wonder if this roundtrip to string can be replaced by some jackson magic.

src/main/java/org/zalando/logbook/Obfuscator.java
```diff
@@ -24,6 +24,8 @@
 import java.util.function.BiPredicate;
 import java.util.function.Predicate;
 
+// TODO implement
+// TODO find a nice way for body obfuscation
 @FunctionalInterface
 public interface Obfuscator {
```
jhorstmann: The api with the check for `(!Objects.equals(replacement, value))` looks a bit awkward. Is there a usecase where the obfuscator needs to decide on the value? I would prefer a separate method for checking whether an obfuscator applies to a header name.

README.md
```diff
@@ -1,62 +1,100 @@
...
+
+```xml
+<filter>
+    <filter-name>LogbookFilter</filter-name>
+    <filter-class>org.zalando.logbook.LogbookFilter</filter-class>
+</filter>
+<filter-mapping>
+    <filter-name>LogbookFilter</filter-name>
+    <url-pattern>/*</url-pattern>
+    <dispatcher>FORWARD</dispatcher>
```
jhorstmann: Would we actually need the once per request logic if the filter is not registered for FORWARD / INCLUDE?

whiskeysierra: > Would we actually need the once per request logic if the filter is not registered for FORWARD / INCLUDE?

No clue, tbh

src/main/java/org/zalando/logbook/Named.java
whiskeysierra: > Again, I would prefer to put this in a method of OnceFilter, this should really be a hidden implementation detail of the filter.

Already moved it internally.

src/main/java/org/zalando/logbook/Markable.java
whiskeysierra: > but here it just makes the code harder to follow.

I inlined most of them now (not pushed yet).

src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
whiskeysierra: > I wonder if this roundtrip to string can be replaced by some jackson magic.

What else do you suggest?

jhorstmann: > What else do you suggest?

It should be possible to give the `byte[]` to jackson and let it serialize that as string. By default it would probably use base64, I'm not sure how that could be changed.

logbook-core/pom.xml
```diff
@@ -0,0 +1,123 @@
...
+            <artifactId>guava</artifactId>
+            <version>18.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>1.7.12</version>
+        </dependency>
+        <dependency>
+       		<groupId>org.projectlombok</groupId>
```
jhorstmann: Are you using lombok just for one `SneakyThrows` or did I miss any more usages?

logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogFormatter.java
```diff
@@ -0,0 +1,131 @@
...
+        if (!body.isEmpty()) {
+            lines.add("");
+            lines.add(body);
+        }
+
+        return join(lines);
+    }
+
+    private String formatRequestLine(final HttpRequest request) {
+        return String.format("%s %s HTTP/1.1", request.getMethod(), getRequestURI(request));
```
jhorstmann: I'd prefer to keep the [original request line][1], based on `HttpServletRequest.getRequestURI()` and `.getQueryString()`. That should also avoid confusion with form parameters, which would still be logged as the request body.

 [1]: http://stackoverflow.com/a/2222268/139595

logbook-core/pom.xml
whiskeysierra: > Are you using lombok just for one SneakyThrows or did I miss any more usages?

This is the only usage right here. I haven't check if it actually helps to get full coverage.

logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogFormatter.java
whiskeysierra: I talked to @lukasniemeier-zalando about this already and I think it hard for certain implementations to provide the request uri, query string (or parameters) and form parameters.

The Servlet API e.g. combines both kinds of parameters into one single multimap. I'm not even sure if POST parameters in the body are available to filters or whether the Container already consumed them.

Spring's RestTemplate interceptors don't provide parameters at all.

logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogFormatter.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @Override
+    public String format(final Precorrelation<HttpRequest> precorrelation) throws IOException {
+        final HttpRequest request = precorrelation.getRequest();
+        final List<String> lines = new ArrayList<>();
+
+        lines.add(formatRequestLine(request));
+        lines.addAll(formatHeaders(request.getHeaders()));
+
+        final String body = request.getBodyAsString();
```
bocytko: Would it make sense to extract lines 45-51 as they're equal to 66-72?

logbook-core/src/test/java/org/zalando/logbook/ObfuscatorTest.java
```diff
@@ -0,0 +1,66 @@
...
+        final Obfuscator unit = Obfuscator.authorization();
+
+        assertThat(unit.obfuscate("Accept", "text/plain"), is(equalTo("text/plain")));
+    }
+
+    @Test
+    public void compoundShouldObfuscateMultipleTimes() {
+        final Obfuscator unit = Obfuscator.compound(
+                Obfuscator.obfuscate((key, value) -> "XXX".equals(value), "YYY"),
+                Obfuscator.obfuscate((key, value) -> "password".equals(key), "<secret>"),
```
bocytko: It's not necessarily needed, right?

logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogFormatter.java
whiskeysierra: :+1:

logbook-core/src/test/java/org/zalando/logbook/ObfuscatorTest.java
whiskeysierra: Indeed. I wanted to have one obfuscator in there that is actually not used to illustrate the usage. Maybe I should add a comment.

bocytko: Both a comment and/or dedicated test would do the trick! :+1:

logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogFormatter.java
whiskeysierra: I changed it as follows:

- `HttpRequest.getRequestUri()` returns the original URI from the request line, including query paramters
- `HttpRequest.getParameters` no longer exists
- The parameter obfuscator works on query parameters alone
- There is no builtin support for form parameters in POST bodies (yet?)
- although that could be implemented using the existing body obfuscator interface and delegating to a key-value based obfuscator

PR #12, opened 2015-10-08
Fixed IllegalStateException when accessing AsyncContext

whiskeysierra:
Turns out that `ServletRequest.getAsyncContext()` by contract is supposed to throw an `IllegalStateException` if no async process was started. I just used `request.getAsyncContext() == null` to see if the current filter pass was a part of an async dispatch. Coincidentally Tomcat doesn't adhere to the spec here and actually returns null, which made it look like it worked. Testing it with Undertow though showed the problem.

Additionally Spring MockMvc doesn't generate a *true* async dispatch request; the dispatcher type was still set to `REQUEST`. Fun stuff...


lukasniemeier-zalando:
Sounds like you had fun :+1:

PR #14, opened 2015-10-12
Fixed invalid json in case of empty json body

### LINKED ISSUE #13: Support empty bodies for application/json messages
The `JsonHttpLogFormatter` silently assumes a non-empty body in case the content type is JSON. Not sure if an empty body is technically valid JSON (I think it's not), but we could juse fallback to `null` here in our logged json string.

whiskeysierra:
This fixes #13

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
@@ -100,7 +100,7 @@ public final class JsonHttpLogFormatter implements HttpLogFormatter {
         final String body = request.getBodyAsString();
 
         if (isJson(request.getContentType())) {
-            builder.put("body", new JsonBody(compactJson(body)));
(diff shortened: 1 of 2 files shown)


logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
```diff
@@ -100,7 +100,7 @@ private void addBody(final HttpMessage request, final ImmutableMap.Builder<Strin
         final String body = request.getBodyAsString();
 
         if (isJson(request.getContentType())) {
-            builder.put("body", new JsonBody(compactJson(body)));
+            builder.put("body", body.isEmpty() ? JsonBody.NULL : new JsonBody(compactJson(body)));
```
jhorstmann: Should this be an empty string instead? Otherwise you can't distinguish between an empty body and a body containing the string "null".

whiskeysierra: True. I changed it to an empty string.

PR #19, opened 2015-11-04
Fixed flush issue with buffering print writers

### LINKED ISSUE #18: Servlet filter is swallowing response body
I don't have a reproducible test case, but when enabling the Jolokia endpoint in a Spring Boot application (http://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-jmx.html#production-ready-jolokia) it neither logs nor writes the original response body:

Issuing this request:

```http
GET /jolokia/read/java.lang:type=Memory/HeapMemoryUsage/used HTTP/1.1
Cookie: JSESSIONID=PErsV8g5P12cKvCmMZ3L29fR
Authorization: XXX
Cache-Control: no-cache
Accept: */*
Connection: keep-alive
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Ubuntu Chromium/45.0.2454.101 Chrome/45.0.2454.101 Safari/537.36
Host: localhost:8080
Postman-Token: d7248d0c-1e58-409e-003d-d9fdd8c6850c
Accept-Encoding: gzip, deflate, sdch
Accept-Language: en-US,en;q=0.8
```

will only log this response:

```http
HTTP/1.1 200
Strict-Transport-Security: max-age=31536000 ; includeSubDomains
Cache-Control: no-cache
Pragma: no-cache
Expires: Tue, 03 Nov 2015 11:11:06 GMT
X-Application-Context: application
Date: Tue, 03 Nov 2015 12:11:06 GMT
Content-Type: text/plain;charset=utf-8
```

When I disable the logbook filter the response is returned to the browser.

```http
{
  "request": {
    "path": "used",
    "mbean": "java.lang:type=Memory",
    "attribute": "HeapMemoryUsage",
    "type": "read"
  },
  "value": 195264032,
  "timestamp": 1446552834,
  "status": 200
}
```

whiskeysierra:
This fixes #18

### DIFF
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LogbookFilter.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LogbookFilter.java
@@ -58,6 +58,7 @@ public final class LogbookFilter implements HttpFilter {
 
             try {
                 chain.doFilter(request, response);
+                response.getWriter().flush();
             } finally {
                 logResponse(correlator.get(), request, response);
             }
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/TeeResponse.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/TeeResponse.java
@@ -48,15 +48,20 @@ final class TeeResponse extends HttpServletResponseWrapper implements RawHttpRes
     private final HttpServletRequest request;
     private final ByteArrayDataOutput output = ByteStreams.newDataOutput();
 
+    private final TeeServletOutputStream stream;
+    private final PrintWriter writer;
+
     /**
      * Null until we a) capture it ourselves or b) retrieve it from {@link Attributes#RESPONSE_BODY}, which
      * was previously captured by another filter instance.
      */
     private byte[] body;
 
-    TeeResponse(final HttpServletRequest request, final HttpServletResponse response) {
+    TeeResponse(final HttpServletRequest request, final HttpServletResponse response) throws IOException {
         super(response);
         this.request = request;
+        this.stream = new TeeServletOutputStream();
+        this.writer = new TeePrintWriter();
     }
 
     @Nullable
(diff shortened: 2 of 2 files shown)


lukasniemeier-zalando:
:+1:

jhorstmann:
:+1: This also seems to fix the issues I had with using Logbook in a jsp application.

Is it really required to call `response.getWriter().flush()` in `LogbookFilter`? I could imagine that could cause problems if the response is already committed. Per the spec you are also not supposed to call `getWriter` if `getOutputStream` was called before.

whiskeysierra:
> This also seems to fix the issues I had with using Logbook in a jsp application.

That was my hope.

> Is it really required to call response.getWriter().flush() in LogbookFilter?

Excellent question. I was kind of puzzled as to why *nobody* was calling flush on the response ever. I placed a breakpoint on it and called `flush` manually and that worked. So I'm definitely missing one crucial piece of information here. Who is flushing/closing the response output stream under normal conditions?

I looked at Spring's [ContentCachingResponseWrapper](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/util/ContentCachingResponseWrapper.html) and it calles flush on the underlying stream [on every write](https://github.com/spring-projects/spring-framework/blob/d5ee787e1e6653257720afe31ee3f8819cd4605c/spring-web/src/main/java/org/springframework/web/util/ContentCachingResponseWrapper.java#L228)!

```java
@Override
public void write(char buf[], int off, int len) {
    super.write(buf, off, len);
    super.flush();
}

@Override
public void write(String s, int off, int len) {
    super.write(s, off, len);
    super.flush();
}

@Override
public void write(int c) {
    super.write(c);
    super.flush();
}
```

> Per the spec you are also not supposed to call getWriter if getOutputStream was called before.

Correct, but in thise case we are **not** calling those methods on anyone's response, but on our own class. And since we implemented the stream on top of the writer we can be sure it works either way.

PR #23, opened 2015-11-26
Added security strategy support

### LINKED ISSUE #17: Add support for security setups to logbook-servlet
In our common setup we would like to have a filter chain setup like this:

1. logging, 1st pass
2. security
2. logging, 2nd pass

The first logging filter should log both, request and response, if and only if the response is a 401 (or 403?) and it should ignore the request body. This is desired because unauthorized requests can potentially be DoS attacks with very large requests to take our service down.

The second logging filter logs requests as they come in and responses as they go out - business as usual. The only important thing is, that any request/response is only **logged exactly once**.

whiskeysierra:
Still **missing**:
- [x] some description in the readme

Fixes #17


lukasniemeier-zalando:
:+1:

PR #27, opened 2015-12-02
Removed buffer-once semantic in multi-filter setups

### LINKED ISSUE #24: Remove "buffer once" semantic
It doesn't make sense this it doesn't allow to log requests/responses at different stages of the pipeline. A common setup could be to log responses when the application returned and before writing them, to log differences, e.g. automatic embedding of resources etc.

whiskeysierra:
This fixes #24


lukasniemeier-zalando:
:+1:

PR #34, opened 2015-12-03
Requests will now contain an absolute request uri

### LINKED ISSUE #30: Log full URI on request
Currently protocol and host are omitted when logging the request `URI`. It would be cool if we could have the full `URI` there.

	{
	  "correlation": "40e13d1e-3b57-4a94-a954-90df38bbc12c",
	  "remote": "172.17.0.1",
	  "method": "GET",
	  "uri": "/oauth2/tokeninfo?access_token=XXX",
	  "headers": {
	    "Accept": [
	      "application/json, application/*+json"
	    ],
	    "User-Agent": [
	      "Apache-HttpClient/4.5.1 (Java/1.8.0_66-internal)"
	    ],
	    "Connection": [
	      "Keep-Alive"
	    ],
	    "Host": [
	      "rhino.example.org"
	    ],
	    "Accept-Encoding": [
	      "gzip,deflate"
	    ]
	  }
	}

whiskeysierra:
This fixes #30

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,10 @@
 [![Release](https://img.shields.io/github/release/zalando/logbook.svg)](https://github.com/zalando/logbook/releases)
 [![Maven Central](https://img.shields.io/maven-central/v/org.zalando/logbook.svg)](https://maven-badges.herokuapp.com/maven-central/org.zalando/logbook)
 
-*Logbook* is an extensible library to enable request and response logging for different client- and server-side technologies. It comes with a core module `logbook-core` and specific modules per framework, e.g. [`logbook-servlet`](#servlet) for Servlet 3.0 environments and [`logbook-httpclient`](#http-client) for applications using Apache's `HttpClient`.
+*Logbook* is an extensible library to enable request and response logging for different client- and server-side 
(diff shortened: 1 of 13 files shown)


lukasniemeier:
:+1:

whiskeysierra:
For the servlet filter this is easy, as it provides [`HttpServletRequest.getRequestURL`](http://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletRequest.html#getRequestURL--).
The HTTP Client on the other hand unfortunately wraps the request inside an [`HttpRequestWrapper`](http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/client/methods/HttpRequestWrapper.html) which rewrites the URI to `/` (or whatever the path was).

whiskeysierra:
I modified it again and even though it's pretty ugly it should work now.

whiskeysierra:
@lukasniemeier Should we have both versions available, fully qualified vs. path?

whiskeysierra:
> @lukasniemeier Should we have both versions available, fully qualified vs. path?

I'll merge this now. If we decide to do that I'm going to open another issue/pull-request.

PR #36, opened 2015-12-04
Removed ERROR dispatch type from readme and auto configuration

### LINKED ISSUE #32: Logging unauthorized request (error chain)
Imagine the following filter chain configuration:

	Mapping filter: 'tracerFilter' to: [/*]
	Mapping filter: 'unauthorizedLogbookFilter' to: [/*]
	Mapping filter: 'springSecurityFilterChain' to: [/*]
	Mapping filter: 'authorizedLogbookFilter' to: [/*]

Let's say during the _springSecurityFilterChain_ execution an exception occures (internal server error). Sending an unauthorized request to a non-existing resource will result in a log output similar to the following:

	[org.zalando.security.filter] Getting token info from: https://secure.example.org/oauth2/tokeninfo
	[org.zalando.logbook.Logbook] {"correlation":"26004c4a-972a-42d9-8963-b61e237d7b67","remote":"127.0.1.1","method":"GET","uri":"/oauth2/tokeninfo?access_token=XXX","headers":{"Accept":["application/json, application/*+json"],"X-Flow-ID":["RET07vhrT32gzFPnAr3GaQ"],"Host":["secure.example.org"]}}
	[org.zalando.security.filter] Exception getting token info --> status 500
	[org.zalando.logbook.Logbook] {"correlation":"4cd11e6b-0ceb-4392-b607-c61114001e2e","remote":"0:0:0:0:0:0:0:1","method":"POST","uri":"/error","headers":{"Authorization":["XXX"],"Accept":["*/*"],"Host":["localhost:8080"],"Content-Length":["4"],"Content-Type":["text/plain;charset=UTF-8"]},"body":"abcd"}
	[org.zalando.logbook.Logbook] {"correlation":"4cd11e6b-0ceb-4392-b607-c61114001e2e","status":500,"headers":{"Content-Type":["application/json;charset=UTF-8"]},"body":{"status":500,"message":"Error in security filter.","path":"/lala"}}

As you can see

1. The real request is never logged
2. Since logbook was registered in error chain it has consumed the body of the request that was never authorized

Is there a way to detect during error chain execution that the request was never authenticated? Is there a way to still log the initial request (using the security strategy)?

whiskeysierra:
Fixes #32


lukasniemeier:
:+1:

whiskeysierra:
I'll close this for now and try again later.

PR #39, opened 2015-12-08
Fallback to simple String for invalid JSON body

### LINKED ISSUE #38: JsonHttpLogFormatter should be resilient against malformed JSON
Right now `JsonHttpLogFormatter#compactJson` tries to parse the request body based on the content-type. If you pass a malformed body (e.g. syntax error) an exception is thrown and the request will not be handled (and logged).

lukasniemeier-zalando:
Fixes #38

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
@@ -104,14 +104,24 @@ public final class JsonHttpLogFormatter implements HttpLogFormatter {
         final String body = request.getBodyAsString();
 
         if (isJson(request.getContentType())) {
-            try {
-                builder.put("body", body.isEmpty() ? JsonBody.EMPTY : new JsonBody(compactJson(body)));
-                return;
-            } catch (final Exception e) {
-                LOG.trace("Unable to parse body as JSON", e);
-            }
+            builder.put("body", tryParseBodyAsJson(body));
+        } else {
+            addUnless(builder, "body", body, String::isEmpty);
         }
-        addUnless(builder, "body", body, String::isEmpty);
+    }
+
+    private Object tryParseBodyAsJson(final String body) {
+        if (body.isEmpty()) {
+            return JsonBody.EMPTY;
+        }
+
+        try {
+            return new JsonBody(compactJson(body));
+        } catch (final IOException e) {
+            LOG.trace("Unable to parse body as JSON", e);
+        }
+
+        return body;
     }
 
     private boolean isJson(final String contentType) {
--- a/logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
@@ -29,6 +29,8 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.CharMatcher;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.net.MediaType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.StringWriter;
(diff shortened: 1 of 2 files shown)


logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
```diff
@@ -101,10 +104,14 @@ private void addBody(final HttpMessage request, final ImmutableMap.Builder<Strin
         final String body = request.getBodyAsString();
 
         if (isJson(request.getContentType())) {
-            builder.put("body", body.isEmpty() ? JsonBody.EMPTY : new JsonBody(compactJson(body)));
-        } else {
-            addUnless(builder, "body", request.getBodyAsString(), String::isEmpty);
+            try {
+                builder.put("body", body.isEmpty() ? JsonBody.EMPTY : new JsonBody(compactJson(body)));
+                return;
```
jhorstmann: Control flow is a bit confusing. Can we maybe extract a method returning an optional or nullable `JsonBody`?

lukasniemeier-zalando: Yes - actually, why do we differ in behavior for empty JSON-body and empty non-JSON-body @whiskeysierra ?

whiskeysierra: > Yes - actually, why do we differ in behavior for empty JSON-body and empty non-JSON-body @whiskeysierra ?

I think we can switch those conditions and 

1. check for empty body
2. check for content type json

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
```diff
@@ -101,10 +104,14 @@ private void addBody(final HttpMessage request, final ImmutableMap.Builder<Strin
...
+            } catch (final Exception e) {
```
whiskeysierra: Can't we catch something more specific here? `JsonProcessingException` maybe?

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
lukasniemeier-zalando: I tried this but then some tests were failing (those that are checking that for JSON an empty body is serialized as an empty string and for all other content-type the body is omitted). I'd go with omitting the body for an empty body, regardless of the content-type.

whiskeysierra: > I tried this but then some tests were failing (those that are checking that for JSON an empty body is serialized as an empty string and for all other content-type the body is omitted). I'd go with omitting the body for an empty body, regardless of the content-type.

I think I remember now. Depending on which RFC one tries to honor, a json body can be an empty *json* string, i.e. `""`, which would be different from an empty body.

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
lukasniemeier-zalando: We would still have the unlikely case of an `IOException`...

whiskeysierra: > We would still have the unlikely case of an IOException...

Ok, so this then?

```java
} catch (JsonProcessingException | IOException e) {
```

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
whiskeysierra: > Control flow is a bit confusing. Can we maybe extract a method returning an optional or nullable JsonBody?

Would it make sense to use some kind of SPI here to allow *embedding* bodies as json for different content types? Because with the current approach one always needs to change the `JsonHttpLogFormatter`.

lukasniemeier-zalando:
As this is a kind-of breaking bug in this library, should we merge this? We could open an issue to implement the SPI later.

whiskeysierra:
> As this is a kind-of breaking bug in this library, should we merge this? We could open an issue to implement the SPI later.

Agreed. I created #41

PR #51, opened 2016-02-25
Undertow module

twz123:
Logbook for Undertow

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,10 +364,13 @@ served by the [Undertow HTTP server](http://undertow.io/):
 
 ```java
 Undertow.builder()
-        .addHandler(new LogbookHandler(logbook).setNext(next))
         .addHttpListener(8080, "localhost")
+        .setHandler(new LogbookHandler(logbook).setNext(next))
         .build();
 ```
+
+As Undertow is an asynchronous web server, request and response payloads cannot be accessed and hence won't be logged 
+by `LogbookHandler`.
 
 ## Attributions
 
--- a/README.md
+++ b/README.md
@@ -347,6 +347,28 @@ logbook:
         level: INFO
 ```
 
+## Undertow
+
+### Dependency
+
+```xml
+<dependency>
+    <groupId>org.zalando</groupId>
+    <artifactId>logbook-undertow</artifactId>
+    <version>${logbook.version}</version>
+</dependency>
+```
+
+The `logbook-undertow` module contains a `LogbookHandler` that can be used to log HTTP requests and responses 
+served by the [Undertow HTTP server](http://undertow.io/):
+
+```java
+Undertow.builder()
+        .addHandler(new LogbookHandler(logbook).setNext(next))
+        .addHttpListener(8080, "localhost")
+        .build();
+```
+
 ## Attributions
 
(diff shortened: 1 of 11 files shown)


whiskeysierra:
The standard servlet filter doesn't work with undertow?

twz123:
Servlet support is in the separate module `undertow-servlet`. For this scenario, the servlet filter may be a more appropriate approach, since then you also get the payloads.

This module is for plain undertow without using servlets.

twz123:
I'll clarify this in the README.

whiskeysierra:
I'm hesitant to approve or even merge this because I have the feeling that it's a bit too specific and that the use case for this (users of undertow that don't use the servlet API? correct me if I'm wrong) is minimal.

Right now the idea was to offer a core part of the library that anyone can implement *easily* (side note: not sure that we did a good job on that, but that's a different topic), if needed (as you did). For the most common technology stacks we ship default, ready-to-use implementations. On the server side I consider the Servlet API as *the* only relevant specification. Most frameworks (that I know) do in fact operate on top of it (again, correct me if I'm wrong). So the general idea was to implement it once and run it everywhere (tm).

I appreciate the contribution and I'd be willing to include a link/section in the readme for it, but I'd prefer to keep it out of the *official* logbook repository.

Feel free to object, I'm open for compelling arguments.

twz123:
Yeah, Undertow without Servlet API may not be the most typical use case, but I think it would be an interesting addition. The Logbook API is quite easy to implement, true, and additional Logbook bindings could be managed in their own projects, with their own release cycles and their own maintainers. It would just feel more natural to have them as a part of the main Logbook distribution. Providing out-of-the-box support for other HTTP Server and Client solutions would make this project more attractive and versatile in general. Currently, there are quite a few alternatives evolving out there, like Vert.x, Akka HTTP and Play, just to name the ones that come directly to my mind. There are also alternatives on the client side, like [AsyncHttpClient][ahc].

I was also planning to create a module for AsyncHttpClient, and I also thought that it would be best to provide it as an additional module here.

I don't have any figures to argue about the minimality of the use cases for those things, but I know that there is always more than one valuable approach to solve a given probem.

[ahc]: https://github.com/AsyncHttpClient/async-http-client

whiskeysierra:
I re-read the pull-request and reconsidered based on your arguments, but this part made me reject it in its current form:

> As Undertow is an asynchronous web server, request and response payloads cannot be accessed

This is one of the corner stones of logbook. Without seeing concrete payloads, I fear we're missing too much here.

I agree that integrations with other async technologies would be nice, but this is a deal breaker for me.

> I was also planning to create a module for AsyncHttpClient, and I also thought that it would be best to provide it as an additional module here.

Totally agree.

twz123:
> > As Undertow is an asynchronous web server, request and response payloads cannot be accessed

> This is one of the corner stones of logbook. Without seeing concrete payloads, I fear we're missing too much here.

I see… I already thought about if its actually possible to make the payloads available also for vanilla Undertow. Would this PR have a chance to get accepted if there's a feasible solution for that?

whiskeysierra:
> Would this PR have a chance to get accepted if there's a feasible solution for that?

Yes.

PR #53, opened 2016-02-28
Added origin property to requests and responses

### LINKED ISSUE #47: Provide a way to distinguish between incoming and outgoing request/responses

whiskeysierra:
This fixes #47

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/Origin.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/Origin.java
@@ -22,6 +22,6 @@ package org.zalando.logbook;
 
 public enum Origin {
 
-    LOCALHOST, REMOTE
+    LOCAL, REMOTE
 
 }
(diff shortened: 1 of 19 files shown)


whiskeysierra:
@AlexanderYastrebov Would this work for you?

logbook-core/src/main/java/org/zalando/logbook/Origin.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+
+public enum Origin {
+
+    LOCALHOST, REMOTE
```
jhorstmann: How about renaming `LOCALHOST` to `LOCAL`, to avoid confusing it with a host name?

whiskeysierra: > How about renaming LOCALHOST to LOCAL, to avoid confusing it with a host name?

Good idea. I'll change it.

whiskeysierra: > > How about renaming LOCALHOST to LOCAL, to avoid confusing it with a host name?

> Good idea. I'll change it.

Done

lukasniemeier-zalando:
if you fix the example in the README :+1:

PR #56, opened 2016-03-16
Conditional logging

### LINKED ISSUE #45: Provide a way to disable logging of requests/responses on certain urls
Two approaches suggested:
1. Provide internal url mapping inside LogbookFilter to activate/deactivate logging only on certain URLs
2. Reuse underlying logging library functionality by appending request URL path to logger name (so then it is possible to disable logging like this: 
`<Logger name="org.zalando.logbook.Logbook=/some/resource/path" level="off"/>`)

whiskeysierra:
This fixes #45


whiskeysierra:
@AlexanderYastrebov Would this work for you?

AlexanderYastrebov:
Looks ok for me

Issue #50, opened 2016-02-23
Implement query parameter obfuscation in a less invasive way


whiskeysierra:
Right now we try to parse, modify and recreate the request URI for parameter obfuscation. Maybe there is a way to implement it e.g. using regular expressions?

twz123:
One way to solve the issue is to replace the "parameter obfuscation" approach by a complete "request URI obfuscation". So, instead of accepting a key/value obfuscator just for query strings, a Logbook instance could accept a simple requestUriObfuscator that simply obfuscates the whole string (basically a `Function<String, String>`). This way we would have pluggable implementations that may be based on Regexes, or whatever suits best. Moreover, we're able to obfuscate path parameters as well.

I worked on this in my [request-uri-obfuscation](https://github.com/zalando/logbook/compare/master...twz123:request-uri-obfucsation) branch. A nice side-effect is that the whole problematic query-string analysis is now ripped out of the `ObfuscatedHttpRequest` and confined into its own class, that may or may not be used. The users may decide on their own.

whiskeysierra:
> a Logbook instance could accept a simple requestUriObfuscator that simply obfuscates the whole string (basically a Function<String, String>). This way we would have pluggable implementations that may be based on Regexes, or whatever suits best. Moreover, we're able to obfuscate path parameters as well.

Cool idea. I created a pull request that changes the way logbook handles the request uri internally. Implementations of Raw/HttpRequest/Response now need to supply the request uri (without query string) and a Multimap of query parameters. I realized that the servlet API already parses and provides query parameters as a `Map<String, String[]>` so that came for free. I basically removed the QueryParameter class that could parse the query string. For the Http Client module I switched to a preexisting method in URLEncodedUtils or something that provides a way to parse the query string into a list of name/value pairs.

We could enhance it now to allow request url obfuscation in addition

PR #60, opened 2016-04-03
Test to remove guava

ouaibsky:
* All guava code has been move into Util (Probably have to be re-arranged)
* Multimap interface have been created in order to decrease changes
* in starter fix yml file (I guess test is broken on master ?)


logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
```diff
@@ -136,17 +132,15 @@ private boolean isJson(final String contentType) {
...
         }
 
-        final MediaType mediaType = MediaType.parse(contentType);
-
-        final boolean isJson = mediaType.is(APPLICATION_JSON);
+        final boolean isJson = contentType.equalsIgnoreCase(APPLICATION_JSON);
 
-        final boolean isApplication = mediaType.is(MediaType.ANY_APPLICATION_TYPE);
-        final boolean isCustomJson = mediaType.subtype().endsWith("+json");
+        final boolean isApplication = contentType.startsWith("application");
```
AlexanderYastrebov: case sensitive?

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
```diff
@@ -171,8 +165,8 @@ private String compactJson(final String json) throws IOException {
     }
 
     // this wouldn't catch spaces in json, but that's ok for our use case here
-    private boolean isAlreadyCompacted(final String json) {
-        return PRETTY_PRINT.matchesNoneOf(json);
+    private static boolean isAlreadyCompacted(final String json) {
+        return json.matches("[\n\r]");
```
AlexanderYastrebov: PRETTY_PRINT was "\n\t"

logbook-core/src/main/java/org/zalando/logbook/QueryParameters.java
```diff
@@ -20,33 +20,29 @@
...
-final class QueryParameters extends ForwardingMultimap<String, String> {
+final class QueryParameters implements Multimap<String, String>  {
 
     private static final QueryParameters EMPTY = new QueryParameters();
 
     private final Multimap<String, String> parameters;
 
     private QueryParameters() {
-        this(ImmutableMultimap.of());
+        this(Util.of());
```
AlexanderYastrebov: poor naming

logbook-core/src/main/java/org/zalando/logbook/Util.java
```diff
@@ -0,0 +1,118 @@
...
+import java.nio.channels.ReadableByteChannel;
+import java.nio.channels.WritableByteChannel;
+import java.util.*;
+import java.util.stream.Collectors;
+
+/**
+ * Created by clalleme on 03/04/2016.
+ */
+public final class Util {
+    public static <K, V> Multimap<K, V> of() {
```
AlexanderYastrebov: should be immutableOf()

logbook-core/src/main/java/org/zalando/logbook/Util.java
```diff
@@ -0,0 +1,118 @@
...
+        Multimap<String, String> result = of();
+        map.forEach((k, v) -> {
+            result.put(k, v.stream().map(i -> obfuscator.obfuscate(k, i)).collect(Collectors.toCollection(ArrayList::new)));
+        });
+        return result;
+    }
+
+    public static byte[] toByteArray(InputStream is) throws IOException {
+        try (ByteArrayOutputStream os = new ByteArrayOutputStream()) {
+            byte[] buffer = new byte[0xFFFF];
```
AlexanderYastrebov: Seems too big. I would suggest copying this impl from guava directly.

logbook-core/src/main/java/org/zalando/logbook/Util.java
```diff
@@ -0,0 +1,118 @@
...
+            for (int len; (len = is.read(buffer)) != -1; )
+                os.write(buffer, 0, len);
```
AlexanderYastrebov: I would always suggest to use curly braces

logbook-core/src/main/java/org/zalando/logbook/Util.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+    
+    public static <K, V> Multimap<K, V> immutableOf(K x1, V y1, K x2, V y2) {
+        final Multimap<K, V> map = of();
+        map.computeIfAbsent(x1, k -> new ArrayList<>()).add(y1);
+        map.computeIfAbsent(x2, k -> new ArrayList<>()).add(y2);
+        return map;
+    }
+    
+    public static Multimap<String, String> transformEntries(Multimap<String, String> map, final Obfuscator obfuscator) {
```
AlexanderYastrebov: Inconsistent final usage through the class

logbook-spring-boot-starter/src/test/java/org/zalando/logbook/spring/HeaderObfuscatorTest.java
```diff
@@ -37,6 +37,7 @@
     @Test
     public void shouldCreateCompoundObfuscatorFromProperties() {
         assertThat(headerObfuscator.obfuscate("Authorization", "s3cr3t"), is("XXX"));
+        // This test failed because in headers.yam it was K-Access-Token
```
AlexanderYastrebov: Drop comment if it is fixed

logbook-spring-boot-starter/src/test/java/org/zalando/logbook/spring/MockHttpRequest.java
```diff
@@ -20,30 +20,30 @@
...
 import static org.zalando.logbook.Origin.REMOTE;
+import static org.zalando.logbook.Util.firstNonNull;
 
 public final class MockHttpRequest implements HttpRequest {
 
-    private final Origin origin;
-    private final String remote;
-    private final String method;
-    private final String requestUri;
+    private final Origin                   origin;
```
AlexanderYastrebov: I would suggest not to use such indentation

pom.xml
```diff
@@ -311,10 +311,10 @@
             <groupId>com.google.gag</groupId>
             <artifactId>gag</artifactId>
         </dependency>
-        <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-        </dependency>
+        <!--<dependency>-->
```
AlexanderYastrebov: I would suggest to remove instead of commenting out

logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogWriter.java
```diff
@@ -81,7 +81,7 @@ public DefaultHttpLogWriter(final Logger logger, final Level level) {
         }
     }
 
-    @VisibleForTesting
+//    @VisibleForTesting
```
AlexanderYastrebov: I would suggest introduction of own annotation with same name/semantics or just removal instead of commenting out

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
```diff
@@ -59,7 +59,7 @@ public String format(final Precorrelation<HttpRequest> precorrelation) throws IO
         final String correlationId = precorrelation.getId();
         final HttpRequest request = precorrelation.getRequest();
 
-        final ImmutableMap.Builder<String, Object> builder = ImmutableMap.<String, Object>builder();
+        final Map<String, Object> builder = new HashMap<>();
```
AlexanderYastrebov: It is not builder anymore

logbook-core/src/main/java/org/zalando/logbook/QueryParameters.java
```diff
@@ -111,16 +246,15 @@ public static QueryParameters parse(@Nullable final String queryString) {
             return EMPTY;
         }
 
-        final List<String> entries = Splitter.on("&").splitToList(queryString);
-        final Multimap<String, String> parameters = LinkedHashMultimap.create(entries.size(), 1);
-        final Splitter splitter = Splitter.on('=');
+        final List<String> entries = Arrays.asList(queryString.split("&"));//Splitter.on("&").splitToList(queryString);
```
AlexanderYastrebov: BTW semicolon is also legal separator

logbook-core/src/main/java/org/zalando/logbook/Util.java
```diff
@@ -0,0 +1,118 @@
...
+        try (ByteArrayOutputStream os = new ByteArrayOutputStream()) {
+            byte[] buffer = new byte[0xFFFF];
+            for (int len; (len = is.read(buffer)) != -1; )
+                os.write(buffer, 0, len);
+            os.flush();
+            return os.toByteArray();
+        }
+    }
+    
+    public static class BasicMultimap<K, V> extends LinkedHashMap<K, Collection<V>> implements Multimap<K, V> {
```
AlexanderYastrebov: Could be top level class with bunch of factory methods immutableOf* moved out from here

logbook-core/src/main/java/org/zalando/logbook/Util.java
```diff
@@ -0,0 +1,118 @@
...
+    public static <N> N firstNonNull(final N... values) {
+        return Arrays.stream(values).filter(Objects::nonNull).findFirst().get();
+    }
+
+    public static void checkArgument(boolean condition, String errorMsg, Object... errorArgs) {
+        if (!condition)
+            throw new IllegalArgumentException(String.format(errorMsg, errorArgs));
+    }
+
+    public static void copy(final InputStream src, final OutputStream dest) throws IOException {
```
AlexanderYastrebov: All family of byte stream like methods could be moved to i.e. ByteStreamUtils

logbook-core/src/main/java/org/zalando/logbook/Util.java
ouaibsky: what do you mean ?

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
ouaibsky: Yes, I though it doesn't provide any added value there.

logbook-core/src/main/java/org/zalando/logbook/QueryParameters.java
ouaibsky: I didn't notice semicolon was checked in the previous version ?

logbook-core/src/main/java/org/zalando/logbook/ByteStreamUtils.java
```diff
@@ -0,0 +1,53 @@
...
+    public static byte[] toByteArray(InputStream in) throws IOException {
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+        copy(in, out);
+        return out.toByteArray();
+    }
+
+    public static void copy(final InputStream src, final OutputStream dest) throws IOException {
+        final Reader reader = new InputStreamReader(src);
+        final Writer writer = new OutputStreamWriter(dest);
+        copy(reader, writer);
```
jhorstmann: This method should copy byte-wise instead of converting to the default charset

logbook-core/src/main/java/org/zalando/logbook/Multimap.java
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+import java.util.*;
+
+import static java.util.stream.Collectors.toSet;
+
+/**
+ * Created by clalleme on 02/04/2016.
+ */
+public interface Multimap<K, V> extends Map<K, Collection<V>> {
```
jhorstmann: Please note that the semantics of some of the guava MulitMap methods are different, for example getting the value for a non-exitant key would return an empty collection in guava.

logbook-core/src/main/java/org/zalando/logbook/BaseHttpMessage.java
```diff
@@ -20,8 +20,6 @@
  * #L%
  */
 
-import com.google.common.collect.Multimap;
-
 import java.nio.charset.Charset;
 
 public interface BaseHttpMessage {
```
jhorstmann: Maybe the interface can be changed so that `getHeader` returns a `Map<String, Collection<String>>`, making the `MultiMap` completely an implementation detail?

whiskeysierra: > Maybe the interface can be changed so that getHeader returns a Map<String, Collection<String>>, making the MultiMap completely an implementation detail?

The part of the contract where an non-existing key yields an empty collection doesn't work well when this is reduced to a traditional map, where it should be null.

logbook-core/src/main/java/org/zalando/logbook/Util.java
AlexanderYastrebov: @ouaibsky somewhere you use final method parameters and somewhere not

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
```diff
@@ -171,8 +167,8 @@ private String compactJson(final String json) throws IOException {
     }
 
     // this wouldn't catch spaces in json, but that's ok for our use case here
-    private boolean isAlreadyCompacted(final String json) {
-        return PRETTY_PRINT.matchesNoneOf(json);
+    final static boolean isAlreadyCompacted(final String json) {
+        return !json.matches("(?s).*[\n\r\t].*");
```
jhorstmann: Pattern could be static final field, and I think the pattern can be simplified to just `[\n\r\t]`.

logbook-core/src/main/java/org/zalando/logbook/QueryParameters.java
jhorstmann: There is a separate [issue for improving query parameter][1] handling, let's not change this in this pr.

 [1]: https://github.com/zalando/logbook/issues/50

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
ouaibsky: I tested with basic regex and it doesn't work.

logbook-core/src/main/java/org/zalando/logbook/Multimap.java
ouaibsky: Note that my intend is not to rewrite guava here. All tested have been passed while returning null.
if we need this type of feature we can user Map.getOrDefault
Anyway in BasicMultimap implementation I overrode 'get' in order to mimic the guava feature.

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
jhorstmann: > I tested with basic regex and it doesn't work.

The pattern should work, the current pattern searches for `[\n\r\t]` with any characters before or after it, to get the same effect with the simplified pattern you'd have to use `find` instead of `matches`. I don't think there would be a big performance difference, but it would be clearer what's being tested here.

logbook-core/src/test/java/org/zalando/logbook/ByteStreamUtilsTest.java
```diff
@@ -0,0 +1,35 @@
...
+import org.junit.Test;
+
+import static org.junit.Assert.assertTrue;
+
+public class ByteStreamUtilsTest {
+
+    @Test
+    public void shouldConstructorBePackageProtected() {
+        // just for coverage
+        assertTrue(new ByteStreamUtils() != null);
```
jhorstmann: Could be moved to `EnforceCoverageTest` to clarify that this line is not really testing anything.

Ideally I'd like to have the constructor of an utility class to be private, but to reach the coverage goal that would then require reflection in the tests.

ouaibsky: I agree, it's why on other projects I'm working on we didn't fix coverage threshold at 100%.
Other option could be to use PowerMock to avoid repeating reflection stuff.

ouaibsky:
I don't understand why, but each type I rebase from origin master I need to do a push force on my fork.

whiskeysierra:
> I don't understand why, but each type I rebase from origin master I need to do a push force on my fork.

Rebase is rewriting all your commits and you can make a fast-forward push, because git by default forbids it.

logbook-core/src/main/java/org/zalando/logbook/BaseHttpMessage.java
jhorstmann: Another idea, maybe it would be better to introduce a `Headers` class to encapsulate this behaviour. This might also make it easier to make sure all implementations handle headers case-insensitive as in PR #73. It might even share some code with `QueryParameters` regarding obfuscation.

ouaibsky:
Hi 
before re-basing over and over
is there a chance this PR will be accepted ?
If yes, I may have missed some comments, please can you sum-up what is still missing ?
Thx
Christophe

whiskeysierra:
> is there a chance this PR will be accepted ?

Personally I still feel like we benefit from having of Guava, more than we suffer. Initially you said

> I think it's a shame for logbook to get dep on guava which is very huge library

How big would you consider to be acceptable? I did some testing with the Maven Shade Plugin and

> In terms of size I was able to reduce the size of Guava from 2269kb to 878kb.

I still believe that going that way this preferable over copying/rewriting those parts that we use from Guava, since all of the hard work is done automatically. I'd consider having a `logbook-uberjar` (or `logbook-deps`?) that includes all dependencies. To be discussed: repackaged or as-is?

@AlexanderYastrebov @jhorstmann @lukasniemeier-zalando What do you think?

AlexanderYastrebov:
As far as I see the issue is with having dependency on guava itself, not with size. 
For this kind of small utility library it makes sense to have no dependencies at all - this could spread usage of library.
I personally do not like any voodoo with repackaging, so I would prefer refactoring guava out over jar magic.
If it is impossible to complete now, I could imagine general plan to remove it from *-api module and then from others one by one.

jhorstmann:
> What do you think?

I would be in favour of removing the dependency to guava if possible. With java8 the need for guava is already drastically reduced. I think the only open point about this PR is to find an alternative to using `Multimap` for headers and parameters.

whiskeysierra:
Ok, let's do that then. I'd say we can go for a `Map<String, List<String>>` as headers. Since we never mutate it, we should be fine in regards to get and null lists.

whiskeysierra:
@ouaibsky I'd suggest to delay the rebasing/rewrite until #77 is merged. It's going to be a bigger change.

whiskeysierra:
See #88

PR #62, opened 2016-04-15
Update README.md

LappleApple:
Line 17: "customizable conditions which requests to log" -- the language was a bit unclear. Is my edit correct? Should it be "conditions that make requests to log" or "conditions that make logging requests"?
Line 60:  "All integrations require an instance of Logbook which holds all configuration and wires all necessary parts together" does this mean that you need "an instance of Logbook that holds all configuration...," or does this mean that "Logbook [always] holds all configuration...?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Logbook
+# Logbook: an extensible Java library for HTTP request and response logging
 
 [![Logbook](docs/logbook.jpg)](#attributions)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,36 +7,28 @@
...
+### Dependencies
 
 - Java 8
 - Any build tool using Maven Central, or direct download
 - Servlet Container (optional)
 - Apache HTTP Client (optional)
 - Spring Boot (optional)
 
-## Installation
+### Installation
```
whiskeysierra: Threre is no second level headline now, is there?

README.md
```diff
@@ -7,36 +7,28 @@
...
-
-## Dependencies
+**Logbook** is an extensible Java library to enable complete request and response logging for different client- and server-side technologies. It satisfies a special need by a) allowing web application developers to log any HTTP traffic that an application receives or sends b) in a way that makes it easy to persist and analyze it later. This can be useful for traditional log analysis, meeting audit requirements or investigating individual historic traffic issues. 
+
+Logbook is ready to use out of the box for most common setups. Even for uncommon applications and technologies, it should be simple to implement the necessary interfaces to connect a library/framework/etc. to it.
+
+### Features
+
+-  *Logging*: of HTTP requests and responses, including the body; partial logging (no body) for unauthorized requests
+-  *Customization*: of logging format, logging destination, and conditions that request to log
```
whiskeysierra: > conditions that request to log

What I meant to say was that a user can provide custom conditions to disable logging for certain requests (and responses)

README.md
```diff
@@ -7,36 +7,28 @@
...
+-  *Support*: for Servlet containers, Apache’s HTTP client, and (via its elegant API) other frameworks
```
whiskeysierra: > via its elegant API

Patting ourselves on the back, aren't we?

README.md
```diff
@@ -1,4 +1,4 @@
-# Logbook
+# Logbook: an extensible Java library for HTTP request and response logging
```
whiskeysierra: I wouldn't put this in the headline. Now it breaks into multiple lines. Either a normal-text *sub-headline* or we put it as the first sentence to the description.

LappleApple: +1. let me play with it -- I'll get it to one line.

README.md
LappleApple: Got it. So you could say: 

*Customization*: of logging format, logging destination, and any provided conditions that disable logging for certain requests (and responses)

README.md
LappleApple: why not?

README.md
whiskeysierra: > Customization: of logging format, logging destination, and any provided conditions that disable logging for certain requests (and responses)

:+1:

README.md
whiskeysierra: > why not?

I usually try to be more humble and yes, I realize the irony in a statement like that ;)

LappleApple: OK, we can take it out :)

README.md
LappleApple: I just compared the original with the edit and the levels are the same; wondering what you're seeing.

README.md
```diff
@@ -1,4 +1,4 @@
-# Logbook
+## Logbook: HTTP request and response logging
```
whiskeysierra: The former web developer in me cringes, because there should be a single h1 headline on a page. This feels wrong. Have you tried:

```html
<h1>Logbook <small>HTTP request and response logging</small></h1>
```

whiskeysierra: > Have you tried

Doesn't work 👎

LappleApple: Ah, I had shortened it but because of the old change request couldn't tell if I'd gotten it down to a single line. We can just keep Logbook for now and think about it a bit later, wdyt?

whiskeysierra: > We can just keep Logbook for now and think about it a bit later, wdyt?

:+1:

PR #73, opened 2016-04-18
Made headers case insensitive

### LINKED ISSUE #67: Header map in requests/responses should be case-insensitive

whiskeysierra:
Fixes #67


logbook-servlet/src/main/java/org/zalando/logbook/servlet/TeeResponse.java
```diff
@@ -71,14 +70,20 @@ public Origin getOrigin() {
     }
 
     @Override
-    public Multimap<String, String> getHeaders() {
-        final Multimap<String, String> headers = ArrayListMultimap.create();
+    public ListMultimap<String, String> getHeaders() {
+        this.headers = Optional.ofNullable(headers)
```
jhorstmann: I'd prefer an if statement here instead of constantly re-assigning the field. That would also tell you if the case is actually covered by tests :)

whiskeysierra: > I'd prefer an if statement here instead of constantly re-assigning the field. That would also tell you if the case is actually covered by tests :)

You have eyes like a hawk. To be honest that was the exact reason to do it like that, to trick code coverage m)

whiskeysierra: > That would also tell you if the case is actually covered by tests :)

I re-checked and caching of headers is not required for any implementation of Raw/HttpRequest/Response, since the default implementation of Logbook and obfuscation can do that once and for all.

PR #74, opened 2016-04-18
Extracted query parameters from request uri

### LINKED ISSUE #50: Implement query parameter obfuscation in a less invasive way

whiskeysierra:
This fixes #50


logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogFormatter.java
```diff
@@ -42,7 +43,12 @@ public String format(final Precorrelation<HttpRequest> precorrelation) throws IO
     }
 
     private String formatRequestLine(final HttpRequest request) {
-        return String.format("%s %s HTTP/1.1", request.getMethod(), request.getRequestUri());
+        return String.format("%s %s HTTP/1.1", request.getMethod(), renderRequestUri(request));
```
jhorstmann: Unrelated to this PR, but we don't actually know the request was done with HTTP/1.1, for a servlet implementation this string could be retrieved from `ServletRequest.getProtocol`.

logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogFormatter.java
```diff
@@ -42,7 +43,12 @@ public String format(final Precorrelation<HttpRequest> precorrelation) throws IO
...
+    }
+
+    private String renderRequestUri(HttpRequest request) {
+        final String query = Joiner.on("&").join(request.getQueryParameters().entries());
```
jhorstmann: This looks like it is relying on the `toString` implementation of a `Map.Entry`, I don't think this is guaranteed to be in the wanted format.

logbook-servlet/src/main/java/org/zalando/logbook/servlet/RemoteRequest.java
```diff
@@ -73,14 +75,22 @@ public String getRemote() {
...
-        @Nullable final String queryString = getQueryString();
-        return queryString == null ? uri : uri + "?" + queryString;
+        return getRequestURL().toString();
+    }
+
+    @Override
+    public ListMultimap<String, String> getQueryParameters() {
+        final ListMultimap<String, String> parameters = ArrayListMultimap.create();
+        
+        getParameterMap().forEach((name, values) ->
```
jhorstmann: Unfortunately this would include POST parameters and the servlet api is not low-level enough to distinguish between query string and body parameters.

logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogFormatter.java
whiskeysierra: > Unrelated to this PR, but we don't actually know the request was done with HTTP/1.1, for a servlet implementation this string could be retrieved from ServletRequest.getProtocol.

I created #75 to keep track on that.

logbook-servlet/src/main/java/org/zalando/logbook/servlet/RemoteRequest.java
whiskeysierra: > Unfortunately this would include POST parameters and the servlet api is not low-level enough to distinguish between query string and body parameters.

You're right. We could check for the method and assume that nobody is using query and form parameters at the same time. Or we openly discourage the use of `application/x-www-form-urlencoded` e.g. in the readme.

logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogFormatter.java
whiskeysierra: > This looks like it is relying on the toString implementation of a Map.Entry, I don't think this is guaranteed to be in the wanted format.

Correct. I fixed it my partially reverting the deletion of our custom QueryParameters class. I only left the rendering in there.

whiskeysierra: > This looks like it is relying on the toString implementation of a Map.Entry, I don't think this is guaranteed to be in the wanted format.

Correct. I fixed it my partially reverting the deletion of our custom QueryParameters class. I only left the rendering in there.

PR #77, opened 2016-04-21
Added support for excludes in Spring Boot

### LINKED ISSUE #57: Add Spring Boot support for predicate

### LINKED ISSUE #79: URI obfuscation

whiskeysierra:
This fixes #57
This fixes #79


whiskeysierra:
@matosf @waleed-madanat What do you think?

logbook-spring-boot-starter/src/test/java/org/zalando/logbook/spring/ExcludeTest.java
```diff
@@ -0,0 +1,97 @@
...
+        public Logger httpLogger() {
+            final Logger logger = mock(Logger.class);
+            when(logger.isTraceEnabled()).thenReturn(true);
+            return logger;
+        }
+        
+        @Bean
+        public Predicate<RawHttpRequest> requestPredicate() {
+            return request -> 
+                    !"/health".equals(request.getRequestUri());
```
matosf: in a "real" application the request.getRequestUri() will contain the full URL, not only the path (i.e. http://localhost/health).
should the RawHttpRequest contain a getPath method to facility this predicate mapping?

whiskeysierra: > in a "real" application the request.getRequestUri() will contain the full URL, not only the path (i.e. http://localhost/health).

You're absolutely right.

> should the RawHttpRequest contain a getPath method to facility this predicate mapping?

Most implementations of that interface will only have the full URI at hand. In order to access just the path one could use `URI.create(..).getPath()`, but that will lead us back to issue #49:

> **Problem**
When broken or malicious clients issue HTTP requests with invalid request targets (e.g. wrongly escaped query parameters) in the HTTP Request Line, logbook fails by throwing a IllegalArgumentException/URISyntaxException. This results in unlogged HTTP requests and most probably in 500 Internal Server Error responses from HTTP backends.
>
> **Proposed fix**
Don't use java.net.URI as return type of BaseHttpRequest::getRequestUri, but a plain String. Implementation was straight forward, with one exception: Query Parameter Obfuscation. If the request URI is malformed, query parameters cannot be extracted in the usual way in order to apply the Obfuscator to them. Currently, malformed URIs are then forwarded as-is. This may not be desirable. What do you think?

I'm not sure what to do here.

Another, rather unrelated question: I introduced `RequestPredicates` in another pull request to the core. It offers some predicates to match/exclude requests. It has support to match URIs 100% (`String.equals`) and based on a pattern (`Pattern.matches`). Since the `AntPathMatcher` is a part of Spring I wouldn't want to include it in the core. What's the best course of action here? Introduce another module `logbook-matcher` that has a spring dependency just for that single class so users of the library can choose whether they want it or not?

whiskeysierra: @twz123 Since you found the original issue, do you have an opinion about this?

twz123: First of all, to summarize the problem (to double check if I understood it correctly): In order to implement the request predicate more easily, we want to extract the path from the URI?

In the example in the test, I'd just use a regex or the like (e.g. `"^/health($|\\?)"`) to match all URIs with that path.

More generally, I'd not introduce URL parsing directly into `RawHttpRequest` because this would mean that we need to cope with invalid URLs inside Logbook itself, which in turn could result in additional configuration options that influence the handling of those URLs. So, in my opinion, the implementor should cope with the parsing. Maybe no parsing is necessary at all, maybe a regex is enough, maybe some more complex logic needs to be in place. What could be offered by Logbook itself is some utility functionality for URL parsing that may be used by implementors, if appropriate. I experimented with a [fallback URL parsing mechanism](https://github.com/zalando/logbook/compare/master...twz123:leninet-obfuscate-urls#diff-4651f587bef3d5da34fd5047d667a09aR76) back in the days of #49 to address this. It just tries to parse an URL on a best-effort basis and never fails. Maybe this could be reworked to a real utility method/class that parses a String into an URI in a fault-tolerant way.

whiskeysierra: > In order to implement the request predicate more easily, we want to extract the path from the URI?

Exactly.

> I'd just use a regex or the like (e.g. "^/health($|\\?)") to match all URIs with that path.

As @matosf correctly pointed out, real requests may be absolute, e.g. `http://localhost/health`.

> More generally, I'd not introduce URL parsing directly into RawHttpRequest because this would mean that we need to cope with invalid URLs inside Logbook itself

Agreed.

> So, in my opinion, the implementor should cope with the parsing.

Implementor of what? The predicate or the RawHttpRequest interface?

> Maybe no parsing is necessary at all, maybe a regex is enough, maybe some more complex logic needs to be in place.

I liked the idea to use Ant-style path matchers, e.g. `/admin/**`. We could of course transform those into a regex, that might be less effort.

> What could be offered by Logbook itself is some utility functionality for URL parsing

That's what I would like to avoid. At least inside of Logbook. Could see this as its own project though.

matosf:
Then we should use regex instead of Ant-style path matchers, because its a more flexible solution and does not depend on Spring, even if its not so user friendly.

whiskeysierra:
> Then we should use regex instead of Ant-style path matchers, because its a more flexible solution and does not depend on Spring, even if its not so user friendly.

Maybe there is an *easy* way to translate from glob/ant-style patterns to regex?

logbook-spring-boot-starter/src/test/java/org/zalando/logbook/spring/ExcludeTest.java
twz123: > As @matosf correctly pointed out, real requests may be absolute, e.g. http://localhost/health.

That was my misunderstanding ;) Could still be done using a regex, though. But yes, it gets uglier.

> Implementor of what? The predicate or the RawHttpRequest interface?

The predicate.

> I liked the idea to use Ant-style path matchers, e.g. /admin/**. We could of course transform those into a regex, that might be less effort.

I'm afraid that I'm still missing something here. Are we just talking about the Spring Starter or Logbook in general? For the Spring Starter, using Ant matchers for the excludes list is probably the expected choice. So no need to add an additional custom request predicate that again just uses Ant matchers. Since it's already a Spring app, an additional Spring dependency should be no big deal.

Non-Spring apps have to create Logbook instances in a different way. It's just up to their decision how they implement the request predicate. Logbook should not bother, and currently, it indeed doesn't.

> That's what I would like to avoid. At least inside of Logbook. Could see this as its own project though.

Not sure why you want to avoid it, but this was just a thought. Perfectly fine with not having this inside Logbook as a util. If an app needs a special request predicate, it can come up best with it's own strategies how to match requests. If that involves more complex URL analysis, the app can do what fits best.

whiskeysierra: > Are we just talking about the Spring Starter or Logbook in general?

Right now I'm using the ant-style path matcher only inside the starter module which is fine, in terms of dependency management. But there are a couple of generic predicates already in the core (not part of this PR, since already merged), see `RequestPredicates`. Those match on string equals and Pattern (regex). I'd like to only support one of those, and I prefer the ant-style.

> If an app needs a special request predicate, it can come up best with it's own strategies how to match requests. If that involves more complex URL analysis, the app can do what fits best.

I'd consider URL path matching predicates universally useful to all our clients, so I see the use case for having something like it in the logbook *ecosystem*.

whiskeysierra:
> Maybe there is an easy way to translate from glob/ant-style patterns to regex?

Well, there is [now](https://github.com/whiskeysierra/logbook/blob/feature/spring-boot-predicate/logbook-core/src/main/java/org/zalando/logbook/Glob.java#L27).

whiskeysierra:
@lukasniemeier-zalando Do you want to take a look?

logbook-api/src/main/java/org/zalando/logbook/Logbook.java
```diff
@@ -33,16 +35,35 @@ static Logbook create() {
...
+        return Creator.create(predicate, formatter, writer, headerObfuscator, queryObfuscator, bodyObfuscator);
+    }
 
-        final LogbookFactory factory = LogbookFactory.INSTANCE;
-        return factory.create(formatter, writer, predicate, headerObfuscator, parameterObfuscator, bodyObfuscator);
+    static Creator.Builder builder() {
+        return Creator.builder();
     }
 
+    @Hack("The Lombok IDEA plugin doesn't like @Builder on static interface methods")
```
whiskeysierra: Does anybody else have that issue?

I have the latest version of IDEA and the Lombok plugin running. It works for normal classes, but shows generated methods in interfaces as red/non-existent. Compiling/testing/running works.

logbook-api/src/main/java/org/zalando/logbook/QueryObfuscator.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    static QueryObfuscator compound(final Iterable<QueryObfuscator> obfuscators) {
+        return StreamSupport.stream(obfuscators.spliterator(), false)
+                .reduce(none(), (left, right) ->
+                        query ->
+                                left.obfuscate(right.obfuscate(query)));
+    }
+
+    static QueryObfuscator accessToken() {
```
whiskeysierra: Should we enable this by default (even outside of the spring boot starter)?

logbook-api/src/main/java/org/zalando/logbook/QueryObfuscator.java
```diff
@@ -0,0 +1,62 @@
...
+
+public interface QueryObfuscator {
+
+    String obfuscate(final String query);
+
+    static QueryObfuscator none() {
+        return query -> query;
+    }
+
+    static QueryObfuscator obfuscate(final String name, final String replacement) {
```
whiskeysierra: Should we move this into `logbook-core`? Feels a lot more like an implementation that an API. Should we have static factory classes for the obfuscators? `QueryObfuscators`?

logbook-core/src/main/java/org/zalando/logbook/Glob.java
```diff
@@ -0,0 +1,73 @@
...
+    private static final Pattern GLOB_PATTERN = Pattern.compile("\\?|(/\\*{2}$)|\\*{2}|(/\\*$)|\\*");
+
+    public static Predicate<String> compile(final String glob) {
+        final StringBuilder result = new StringBuilder();
+        final Matcher matcher = GLOB_PATTERN.matcher(glob);
+
+        int end = 0;
+
+        if (matcher.find()) {
+            do {
```
whiskeysierra: I'm pretty confident that this right here is my very first `do-while` loop, like ever.

whiskeysierra:
I merged it for now, because it got too big already. Feel free to review master and open issues.

I didn't want to create feature branches on top of feature branches.

Issue #59, opened 2016-04-01
Remove Guava to make logbook lightweight


ouaibsky:
Hi

I think it's a shame for logbook to get dep on guava which is very huge library
For what I seen, logbook is using java8 and usage of guava in logbook could be replace by basic lambda

If I propose a PR for guava removing, in is there chance to be accepted ?
Thx
Christophe

whiskeysierra:
We are using Guava for much, much more than just functions/predicates. In fact we don't use the functional stuff from guava at all, in favor of the Java 8 versions. Same goes for Optional. We use it for the following:

- Splitter
- Multimap
- ByteArrayDataOutput
- ByteStreams
- CharStreams
- ImmutableCollections
- Preconditions (partially obsolete due to Objects)

None of those has an equivalent in the standard libraries.

You can check for yourself: https://github.com/zalando/logbook/search?utf8=%E2%9C%93&q=com.google

Also I'd like to challenge that
> guava [..] is very huge library

Guava itself has no dependencies, i.e. it's self-contained. Google is actually very reluctant in blindly adding new features to it. We consider it one of the best libraries out there in terms of signal-to-noise ratio and quite frankly, it should be part of the JDK itself.

whiskeysierra:
In addition, if you feel like it, please fork logbook and try to replace all usages of Guava. I'm sure midway through you'll realize that Guava is worth it.

ouaibsky:
Hi

So I was lucky to ask you  this question before starting a PR ;)
My point was not to depreciate guava which is very useful in some case.
The pb (from my point of view, and I accept that it's not shared by everybody) is the size (2.2 Mo) and the fact that you have to pick everything.
Your project weights 18k and aims to be a third library used by more final projects, in some case (no so rare) you have to use many third libraries with conflict in transitive dependencies.

According to me putting this type of dependency is really an architecture choice with a set of trade-off.

I did a quick analysis on logbook-core and removing  guava generates less than 50 compilation errors mainly focused on 
* Multimap (which is mainly a Map<K, Collections<V>)
* ForwardingMultimap
* MediaType and CharMatcher (1 usage)

At the end it's not so much and that's why I though it could be discuss.
But you're the boss on your project :)

Cheers
Christophe.

whiskeysierra:
I wouldn't mind seeing a version of logbook without Guava. If you're
willing to do the work and open a pull request I think we would have more
concrete discussions. I'm not totally opposed to the idea.

[...]

PR #88, opened 2016-05-13
Remove guava dependency

AlexanderYastrebov:
Looks like now everybody will do this)


AlexanderYastrebov:
Fixes #59

whiskeysierra:
Ah! Ok...

logbook-core/src/main/java/org/zalando/logbook/DefaultLogbook.java
```diff
@@ -70,7 +68,7 @@
         }
     }
 
-    @VisibleForTesting
+    //@VisibleForTesting
```
whiskeysierra: I'd just drop it

logbook-core/src/main/java/org/zalando/logbook/DefaultLogbookFactory.java
```diff
@@ -42,15 +42,15 @@ public Logbook create(
             @Nullable final HttpLogWriter writer) {
 
 
-        final HeaderObfuscator header = firstNonNull(headerObfuscator, authorization());
-        final BodyObfuscator body = firstNonNull(bodyObfuscator, BodyObfuscator.none());
+        final HeaderObfuscator header = Optional.ofNullable(headerObfuscator).orElse(authorization());
```
whiskeysierra: Could use `orElseGet(Obfuscators::authorization)`

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
```diff
@@ -59,20 +55,18 @@ public String format(final Precorrelation<HttpRequest> precorrelation) throws IO
         final String correlationId = precorrelation.getId();
         final HttpRequest request = precorrelation.getRequest();
 
-        final ImmutableMap.Builder<String, Object> builder = ImmutableMap.<String, Object>builder();
+        final Map<String, Object> content = new HashMap<>();
```
whiskeysierra: I'd use a LinkedHashMap

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
```diff
@@ -138,12 +130,12 @@ private boolean isJson(final String contentType) {
...
-        final MediaType mediaType = MediaType.parse(contentType);
+        final String t = contentType.toLowerCase();
 
-        final boolean isJson = mediaType.is(APPLICATION_JSON);
+        final boolean isJson = t.startsWith("application/json");
 
-        final boolean isApplication = mediaType.is(MediaType.ANY_APPLICATION_TYPE);
-        final boolean isCustomJson = mediaType.subtype().endsWith("+json");
+        final boolean isApplication = t.startsWith("application/");
+        final boolean isCustomJson = t.endsWith("+json");
```
whiskeysierra: It's missing support for parameters, e.g. `application/custom+json;version=2`

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
```diff
@@ -174,7 +166,7 @@ private String compactJson(final String json) throws IOException {
 
     // this wouldn't catch spaces in json, but that's ok for our use case here
     private boolean isAlreadyCompacted(final String json) {
-        return PRETTY_PRINT.matchesNoneOf(json);
+        return (json.indexOf('\n') == -1);
```
whiskeysierra: Parenthesis are not needed.

logbook-core/src/test/java/org/zalando/logbook/ConditionsTest.java
```diff
@@ -130,32 +131,36 @@ public void headerShouldNotMatchNameAndValue() {
...
     }
 
-    @Test
-    public void headerShouldNotMatchPredicate() {
-        final Predicate<RawHttpRequest> unit = header((name, value) -> 
-                name.equalsIgnoreCase("X-Secret") && value.equalsIgnoreCase("false"));
-
-        assertThat(unit.test(request), is(false));
-    }
+//    @Test
```
whiskeysierra: What happened here?

logbook-core/src/test/java/org/zalando/logbook/ObfuscatorsTest.java
```diff
@@ -59,4 +63,15 @@ public void authorizationShouldNotObfuscateNonAuthorization() {
         assertThat(unit.obfuscate("Accept", "text/plain"), is(equalTo("text/plain")));
     }
 
-}
\ No newline at end of file
+    @Test
+    public void shouldObfuscateMap() {
```
whiskeysierra: The test is not verifying that it actually obfuscates.

logbook-core/src/test/java/org/zalando/logbook/io/EnforceCoverageTest.java
```diff
@@ -0,0 +1,63 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+
+ /*
+ * ⁣​
+ * Jackson-datatype-Problem
```
whiskeysierra: Copy and paste?

logbook-servlet/src/test/java/org/zalando/logbook/servlet/ForwardingHttpLogFormatter.java
```diff
@@ -53,4 +51,8 @@ public String format(final Correlation<HttpRequest, HttpResponse> correlation) t
         return delegate().format(correlation);
     }
 
+    @Override
+    public String toString() {
+        return delegate().toString();
```
whiskeysierra: Good catch! I totally forgot that.

logbook-test/src/test/java/org/zalando/logbook/MockHeadersTest.java
```diff
@@ -0,0 +1,78 @@
...
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+
+ /*
```
whiskeysierra: Double headers?

whiskeysierra:
Almost scary how similar our code is :ghost:

whiskeysierra:
:+1:

logbook-api/src/main/java/org/zalando/logbook/BaseHttpMessage.java
```diff
@@ -35,22 +35,42 @@
...
 
     Charset getCharset();
 
-    class Headers {
+    class HeadersBuilder {
+
+        private final Map<String, List<String>> headers;
 
-        Headers() {
+        public HeadersBuilder() {
```
whiskeysierra: Could be its own class now.

logbook-core/src/main/java/org/zalando/logbook/Conditions.java
```diff
@@ -70,18 +71,18 @@
...
         return request ->
                 request.getHeaders().get(key).stream().anyMatch(predicate);
     }
 
-    public static Predicate<RawHttpRequest> header(final BiPredicate<String, String> predicate) {
-        return request ->
-                request.getHeaders().entries().stream()
-                        .anyMatch(e -> predicate.test(e.getKey(), e.getValue()));
-    }
+//    public static Predicate<RawHttpRequest> header(final BiPredicate<String, String> predicate) {
```
whiskeysierra: Why the comments?

logbook-core/src/main/java/org/zalando/logbook/DefaultLogbookFactory.java
whiskeysierra: Same goes for the other places where the alternative is a method call.

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
whiskeysierra: Not 100% correct, but I'd go with it.

```java
final boolean isCustomJson = t.endsWith("+json") || t.contains("+json;");
```

logbook-core/src/main/java/org/zalando/logbook/Obfuscators.java
```diff
@@ -56,4 +58,12 @@ public static HeaderObfuscator authorization() {
         return obfuscate("Authorization"::equalsIgnoreCase, "XXX");
     }
 
+    static Map<String, List<String>> obfuscateHeaders(final Map<String, List<String>> map, final BiFunction<String, String, String> f) {
+        final BaseHttpMessage.HeadersBuilder builder = new BaseHttpMessage.HeadersBuilder();
+        for (Map.Entry<String, List<String>> e : map.entrySet()) {
+            final String k = e.getKey();
+            builder.put(k, e.getValue().stream().map(x -> f.apply(k, x)).collect(Collectors.toList()));
```
whiskeysierra: Static import for `toList`?

logbook-core/src/test/java/org/zalando/logbook/ConditionsTest.java
```diff
@@ -130,32 +131,36 @@ public void headerShouldNotMatchNameAndValue() {
...
         assertThat(unit.test(request), is(false));
     }
 
-    @Test
-    public void headerShouldMatchPredicate() {
-        final Predicate<RawHttpRequest> unit = header((name, value) -> 
-                name.equalsIgnoreCase("X-Secret") && value.equalsIgnoreCase("true"));
-
-        assertThat(unit.test(request), is(true));
+    private static Set<String> setOf(String... values) {
```
whiskeysierra: :+1:

logbook-core/src/test/java/org/zalando/logbook/EnforceCoverageTest.java
```diff
@@ -20,7 +20,7 @@
  * #L%
  */
 
-/*
+ /*
  * ⁣​
  * Jackson-datatype-Problem
```
whiskeysierra: Copy and paste?

logbook-core/src/main/java/org/zalando/logbook/io/ByteStreams.java
```diff
@@ -0,0 +1,75 @@
+package org.zalando.logbook.io;
+
+/*
```
whiskeysierra: Double header?

logbook-test/src/main/java/org/zalando/logbook/MockHeaders.java
```diff
@@ -20,18 +20,47 @@
  * #L%
  */
 
-import com.google.common.collect.ListMultimap;
+/*
+ * Copyright 2016 Zalando SE.
```
whiskeysierra: Double header?

logbook-test/src/main/java/org/zalando/logbook/MockHeaders.java
```diff
@@ -2,7 +2,7 @@
 
 /*
  * #%L
- * Logbook: Test
```
whiskeysierra: Should be Test, based on the file location

logbook-core/src/main/java/org/zalando/logbook/Conditions.java
AlexanderYastrebov: Did not find a straightforward way to implement this so just dropped for now)

logbook-api/src/main/java/org/zalando/logbook/BaseHttpMessage.java
AlexanderYastrebov: I am thinking of implementing Header and Headers implements Iterable<Header> classes instead of map of lists as this will allow to extend it in future.

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
AlexanderYastrebov: But what for? I am not sure it will serialize in insertion order.

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
AlexanderYastrebov: Sure. It is just for visual grouping.

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
whiskeysierra: Yes, it will

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
whiskeysierra: But there is only one group... :point_up:

logbook-api/src/main/java/org/zalando/logbook/BaseHttpMessage.java
whiskeysierra: But we need a map for easy lookups, e.g. in conditions.

logbook-core/src/main/java/org/zalando/logbook/Conditions.java
whiskeysierra: How about this?

```java
return request ->
        request.getHeaders().entrySet().stream()
                .anyMatch(e -> e.getValue().stream().anyMatch(v -> predicate.test(e.getKey(), v)));
```

logbook-core/src/main/java/org/zalando/logbook/Conditions.java
```diff
@@ -70,18 +71,18 @@
 
     public static Predicate<RawHttpRequest> header(final String key, final String value) {
         return request ->
-                request.getHeaders().containsEntry(key, value);
+                request.getHeaders().getOrDefault(key, Collections.emptyList()).contains(value);
     }
 
     public static Predicate<RawHttpRequest> header(final String key, final Predicate<String> predicate) {
         return request ->
                 request.getHeaders().get(key).stream().anyMatch(predicate);
```
whiskeysierra: This will throw an exception now if a key doesn't exist. Can you verify with a test?

logbook-core/src/main/java/org/zalando/logbook/io/ByteStreams.java
```diff
@@ -0,0 +1,59 @@
...
+        long total = 0;
+        while (true) {
+            int r = from.read(buf);
+            if (r == -1) {
+                break;
+            }
+            to.write(buf, 0, r);
+            total += r;
+        }
+        return total;
```
whiskeysierra: We don't need the total anywhere

logbook-core/src/test/java/org/zalando/logbook/ConditionsTest.java
whiskeysierra: Although asList woulf also work. Since we are only using it in tests.

logbook-core/src/test/java/org/zalando/logbook/ObfuscatorsTest.java
```diff
@@ -59,4 +64,17 @@ public void authorizationShouldNotObfuscateNonAuthorization() {
...
-}
\ No newline at end of file
+    @Test
+    public void shouldObfuscateHeaders() {
+        final Map<String, List<String>> m = Obfuscators.obfuscateHeaders(
+            MockHeaders.of("hello", "world", "chao", "bambina"),
+            (k, v) -> v.toUpperCase()
+        );
+
+        assertEquals(2, m.size());
```
whiskeysierra:
```java
assertThat(m, hasSize(2));
assertThat(m, hasKey("hello"));
assertThat(m, hasKey("chao"));
assertThat(m, hasEntry(equalTo("hello"), contains("WORLD")));
assertThat(m, hasEntry(equalTo("chao"), contains("BAMBINA")));
```

logbook-httpclient/src/test/java/org/zalando/logbook/httpclient/EnforceCoverageTest.java
```diff
@@ -20,26 +20,6 @@
  * #L%
  */
 
-/*
- * ⁣​
- * Jackson-datatype-Problem
```
whiskeysierra: I guess I copied them from somewhere else before...

logbook-servlet/src/test/java/org/zalando/logbook/servlet/FormattingTest.java
```diff
@@ -87,13 +88,13 @@ public void shouldFormatRequest() throws Exception {
...
         final HttpRequest request = interceptRequest();
-        
+
         assertThat(request, hasFeature("remote address", HttpRequest::getRemote, is("127.0.0.1")));
         assertThat(request, hasFeature("method", HttpRequest::getMethod, is("GET")));
         assertThat(request, hasFeature("url", HttpRequest::getRequestUri,
                 hasToString("http://localhost/api/sync?limit=1")));
         assertThat(request, hasFeature("parameters", HttpRequest::getQuery, is("limit=1")));
-        assertThat(request, hasFeature("headers", HttpRequest::getHeaders, is(of("Accept", "text/plain"))));
+        assertThat(request, hasFeature("headers", HttpRequest::getHeaders, is(Collections.singletonMap("Accept", Collections.singletonList("text/plain")))));
```
whiskeysierra: Static imports?

logbook-core/src/main/java/org/zalando/logbook/io/ByteStreams.java
AlexanderYastrebov: I would leave it as is so somebody can use. Overhead is negligible

logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogFormatter.java
```diff
@@ -59,19 +58,19 @@ private String formatStatusLine(final HttpResponse response) {
     private <H extends HttpMessage> String format(final H message, final String type, final String correlationId,
             final Function<H, String> lineCreator)
             throws IOException {
-        final List<String> lines = Lists.newArrayListWithExpectedSize(4);
+        final List<String> lines = new ArrayList<>();
```
jhorstmann: Could still give the expected capacity (which should be 5 instead of 4 to avoid resizing). Indeed for performance reasons this should use a StringBuilder which could be passed to the helper methods. Could be a separate PR though.

logbook-core/src/main/java/org/zalando/logbook/DefaultLogbookFactory.java
```diff
@@ -42,15 +40,15 @@ public Logbook create(
...
+        final BodyObfuscator body = Optional.ofNullable(bodyObfuscator).orElseGet(BodyObfuscator::none);
 
         return new DefaultLogbook(
-                firstNonNull(condition, $ -> true),
+                Optional.ofNullable(condition).orElse($ -> true),
                 combine(queryObfuscator, header, body, requestObfuscator),
                 combine(header, body, responseObfuscator),
-                firstNonNull(formatter, new DefaultHttpLogFormatter()),
-                firstNonNull(writer, new DefaultHttpLogWriter())
+                Optional.ofNullable(formatter).orElseGet(DefaultHttpLogFormatter::new),
```
jhorstmann: Nice, looks like previously we always unnecessarily created the default instances.

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
```diff
@@ -82,17 +76,15 @@ public String format(final Correlation<HttpRequest, HttpResponse> correlation) t
...
-        final ImmutableMap.Builder<String, Object> builder = ImmutableMap.<String, Object>builder();
-
-        builder.put("origin", translate(response.getOrigin()));
-        builder.put("type", "response");
-        builder.put("correlation", correlationId);
-        builder.put("protocol", response.getProtocolVersion());
-        builder.put("status", response.getStatus());
-        addUnless(builder, "headers", response.getHeaders().asMap(), Map::isEmpty);
-        addBody(response, builder);
+        final Map<String, Object> content = new HashMap<>();
```
jhorstmann: Should also be `LinkedHashMap`

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
```diff
@@ -133,19 +125,20 @@ private Object tryParseBodyAsJson(final String body) {
...
 
-        final MediaType mediaType = MediaType.parse(contentType);
+        final int pi = type.indexOf(';');
+        final int limit = (pi != -1 ? pi : type.length());
 
-        final boolean isJson = mediaType.is(APPLICATION_JSON);
-
-        final boolean isApplication = mediaType.is(MediaType.ANY_APPLICATION_TYPE);
-        final boolean isCustomJson = mediaType.subtype().endsWith("+json");
+        return type.regionMatches(true, 0, "application/json", 0, limit)
```
jhorstmann: Might be more readable as a regular expression like

    final Pattern MEDIA_TYPE_JSON_PATTERN = Pattern.compile("application/(?:json|.+\\+json)(?:$|\\s|;)");

    return MEDIA_TYPE_JSON_PATTERN.matcher(type).lookingAt();

logbook-core/src/main/java/org/zalando/logbook/io/ByteStreams.java
```diff
@@ -0,0 +1,59 @@
...
+import java.io.OutputStream;
+import java.util.Objects;
+
+public final class ByteStreams {
+
+    ByteStreams() {
+        // package private so we can trick code coverage
+    }
+
+    private static final int BUF_SIZE = 0x1000; // 4K
```
jhorstmann: Just write 4096 and remove the comment

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
whiskeysierra: I haven't tried it, but I think the custom sub type should be non-greedy

logbook-core/src/test/java/org/zalando/logbook/io/EnforceCoverageTest.java
```diff
@@ -22,15 +22,23 @@
...
 
     @Test
-    public void shouldUseMockHeadersConstructor() {
-        new MockHeaders();
+    public void shouldUseByteStreamsConstructor() {
+        new ByteStreams();
     }
 
+    @Test
+    public void shouldCoverByteStreams() throws IOException {
```
jhorstmann: This should be covered by normal tests, it's not to enforce coverage of untestable methods.

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
whiskeysierra: According to regex101.com it would incorrectly match `application/xml; incompatible=+json`

logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogFormatter.java
whiskeysierra: Default is 10, I guess it wouldn't make much of a difference.

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
jhorstmann: > According to regex101.com it would incorrectly match application/xml; incompatible=+json

Good catch. The definition of tokens and delimiters is here https://tools.ietf.org/html/rfc7230#section-3.2.6 but I think matching everything except space and semicolon should work.

    private static final Pattern MEDIA_TYPE_JSON_PATTERN = Pattern.compile("application/(?:json|[^\t ;]+\\+json)(?:$|\t| |;)");

AlexanderYastrebov: But this one does not look more readable))

logbook-core/src/test/java/org/zalando/logbook/io/EnforceCoverageTest.java
AlexanderYastrebov: I copied ByteStreams methods literaly from guava, but this jacoco enforces 100 coverage that is why it is as is.

logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java
whiskeysierra: Compared to this?

```java
    private boolean isJson(final String type) {
        if (type.isEmpty()) {
            return false;
        }

        final int pi = type.indexOf(';');
        final int limit = (pi != -1 ? pi : type.length());

        return type.regionMatches(true, 0, "application/json", 0, limit)
            || (type.startsWith("application/") && endsWith(type, "+json", limit));
    }

    private boolean endsWith(final String target, final String what, int limit) {
        return target.regionMatches(true, limit - what.length(), what, 0, what.length());
    }
```

I'd say it's still very competitive.

logbook-core/src/test/java/org/zalando/logbook/io/EnforceCoverageTest.java
whiskeysierra: Granted, but it's still a valid test beyond enforcing code coverage. Of course, one could always argue that the only reason we have tests in the first place is to satisfy the coverage requirement.

whiskeysierra:
:+1:

AlexanderYastrebov:
:+1:  nice collective effort

AlexanderYastrebov:
+1

AlexanderYastrebov:
:+1:

PR #93, opened 2016-05-16
Avoid array copy on LocalResponse.getBodyAsString

AlexanderYastrebov:
Make withBody noop and getBody create bytearray lazily
Use output's buffer directly in getBodyAsString


logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
```diff
@@ -111,9 +110,17 @@ public PrintWriter getWriter() throws IOException {
...
     public byte[] getBody() {
+        if (body == null) {
+            body = output.toByteArray();
+        }
         return body;
     }
 
+    @Override
+    public String getBodyAsString() throws IOException {
+        return output.toString(getCharset().name());
```
whiskeysierra: Isn't the default implementation doing pretty much the same?

logbook-servlet/src/test/java/org/zalando/logbook/servlet/LocalResponseTest.java
```diff
@@ -0,0 +1,51 @@
...
+        when(mock.getOutputStream()).thenReturn(new ServletOutputStream() {
+            @Override
+            public void write(int b) throws IOException {
+            }
+        });
+
+        final LocalResponse response = new LocalResponse(mock, "1");
+        response.getOutputStream().write("test".getBytes());
+
+        final byte[] body1 = response.getBody();
```
whiskeysierra: That test is not verying the change, is it? I mean the old implementation would pass this too (assuming you call withBody before).

logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
AlexanderYastrebov: Default one uses getBody() which uses output.getByteArray() which makes copy.
So if codepath uses getBodyAsString, this setup avoids array copying by constructing string directly from output.

whiskeysierra: > Default one uses getBody() which uses output.getByteArray() which makes copy.
> So if codepath uses getBodyAsString, this setup avoids array copying by constructing string directly from output.

But `getBody` doesn't do the copying, `withBody` does and that should only happen once.

AlexanderYastrebov: With this change you do not have copy in withBody now.

whiskeysierra:
:+1:

whiskeysierra:
:+1:

AlexanderYastrebov:
+1

AlexanderYastrebov:
+1

whiskeysierra:
@AlexanderYastrebov Can you approve + merge?

PR #98, opened 2016-05-20
Make headers immutable

### LINKED ISSUE #97: HeadersBuilder builds a mutable data structure
Even though the map is immutable, its values aren't
https://github.com/zalando/logbook/blob/d95a0ca62bd75d6c9055b734c1ebba90761870ba/logbook-api/src/main/java/org/zalando/logbook/BaseHttpMessage.java#L73

AlexanderYastrebov:
Resolve #97

### DIFF
--- a/logbook-api/src/main/java/org/zalando/logbook/BaseHttpMessage.java
+++ b/logbook-api/src/main/java/org/zalando/logbook/BaseHttpMessage.java
@@ -43,7 +43,7 @@ public interface BaseHttpMessage {
 
     class HeadersBuilder {
 
-        private final Map<String, List<String>> headers;
+        private Map<String, List<String>> headers;
 
         public HeadersBuilder() {
             // package private so we can trick code coverage
@@ -73,7 +73,8 @@ public interface BaseHttpMessage {
             for (Map.Entry<String, List<String>> e : headers.entrySet()) {
                 e.setValue(Collections.unmodifiableList(e.getValue()));
             }
-            return Collections.unmodifiableMap(headers);
+            headers = Collections.unmodifiableMap(headers);
+            return headers;
         }
     }
 }
--- a/logbook-api/src/main/java/org/zalando/logbook/BaseHttpMessage.java
+++ b/logbook-api/src/main/java/org/zalando/logbook/BaseHttpMessage.java
@@ -70,6 +70,9 @@ public interface BaseHttpMessage {
         }
 
         public Map<String, List<String>> build() {
+            for (Map.Entry<String, List<String>> e : headers.entrySet()) {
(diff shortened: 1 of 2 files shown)


logbook-api/src/main/java/org/zalando/logbook/BaseHttpMessage.java
```diff
@@ -43,7 +43,7 @@
 
     class HeadersBuilder {
 
-        private final Map<String, List<String>> headers;
+        private Map<String, List<String>> headers;
```
whiskeysierra: I think builders should be reusable, otherwise we are violating the Principle of least astonishment.

logbook-api/src/main/java/org/zalando/logbook/BaseHttpMessage.java
```diff
@@ -70,7 +70,11 @@ public HeadersBuilder put(final String key, final Iterable<String> values) {
         }
 
         public Map<String, List<String>> build() {
-            return Collections.unmodifiableMap(headers);
```
whiskeysierra: What about this?

```java
return headers.entrySet().stream()
    collect(collectingAndThen(
        toMap(Map.Entry::getKey, e -> unmodifiableList(e.getValue()),
        Collections::unmodifiableMap));
```

AlexanderYastrebov: This one loses case insensitive comparator and order of headers

logbook-api/src/main/java/org/zalando/logbook/BaseHttpMessage.java
AlexanderYastrebov: This one is helper for private usage. I do not see any reason to make double copy of Map only to make it reusable, knowing that we always use it only once. On the other side it is defended against reuse so if you ever will need to reuse it you can just refactor it.

logbook-api/src/main/java/org/zalando/logbook/BaseHttpMessage.java
whiskeysierra: > This one loses case insensitive comparator and order of headers

Good point! :+1:

logbook-api/src/main/java/org/zalando/logbook/BaseHttpMessage.java
whiskeysierra: Ok, agreed.

whiskeysierra:
:+1:

whiskeysierra:
:+1:

AlexanderYastrebov:
+1

PR #115, opened 2016-06-27
Allow passing of URIs without defined port

lukasniemeier-zalando:
Changes in `RequestURI` and `LocalRequest` to handle an URI with an undefined port more explicit. The reason for this change is:

```
URI.create("https://example.com/no-port").getPort() == -1
```

### DIFF
--- a/logbook-api/src/main/java/org/zalando/logbook/RequestURI.java
+++ b/logbook-api/src/main/java/org/zalando/logbook/RequestURI.java
@@ -21,6 +21,7 @@ package org.zalando.logbook;
  */
 
 import java.util.EnumSet;
+import java.util.Optional;
 import java.util.Set;
 
 import static java.util.Arrays.asList;
@@ -50,7 +51,7 @@ final class RequestURI {
     private static String reconstruct(final BaseHttpRequest request, final Set<Component> components) {
         final String scheme = request.getScheme();
         final String host = request.getHost();
-        final int port = request.getPort();
+        final Optional<Integer> port = request.getPort();
         final String path = request.getPath();
         final String query = request.getQuery();
 
@@ -63,9 +64,12 @@ final class RequestURI {
         if (components.contains(AUTHORITY)) {
             url.append("//").append(host);
 
-            if (isNotStandardPort(scheme, port)) {
-                url.append(':').append(port);
-            }
+            port.ifPresent(p -> {
+                if (isNotStandardPort(scheme, p)) {
+                    url.append(':').append(p);
+                }
+            });
+
         } else if (components.contains(SCHEME)) {
             url.append("//");
         }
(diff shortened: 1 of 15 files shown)


logbook-api/src/test/java/org/zalando/logbook/ForwardingHttpRequestTest.java
```diff
@@ -51,7 +52,7 @@ public void shouldDelegate() throws IOException {
         assertThat(unit.getRequestUri(), is("http://localhost/"));
         assertThat(unit.getScheme(), is("http"));
         assertThat(unit.getHost(), is("localhost"));
-        assertThat(unit.getPort(), is(80));
+        assertThat(unit.getPort(), is(Optional.of(8080)));
```
AlexanderYastrebov: why 80 became 8080?

logbook-api/src/main/java/org/zalando/logbook/RequestURI.java
```diff
@@ -63,9 +64,12 @@ private static String reconstruct(final BaseHttpRequest request, final Set<Compo
         if (components.contains(AUTHORITY)) {
             url.append("//").append(host);
 
-            if (isNotStandardPort(scheme, port)) {
-                url.append(':').append(port);
-            }
+            port.ifPresent(p -> {
+                if (isNotStandardPort(scheme, p)) {
```
AlexanderYastrebov: What if I run http on 443 or https on 80? Should we care here at all? Maybe just append if present

logbook-api/src/test/java/org/zalando/logbook/RequestURITest.java
```diff
@@ -57,22 +59,29 @@ public void shouldReconstructFully() {
     }
 
     @Test
+    public void shouldNotIncludeStandardHttpPort() {
+        when(request.getScheme()).thenReturn("http");
+        when(request.getPort()).thenReturn(Optional.of(80));
```
AlexanderYastrebov: I would say if it was providednthen it should be included

logbook-api/src/main/java/org/zalando/logbook/RequestURI.java
whiskeysierra: > What if I run http on 443 or https on 80?

This should be handled correctly.

> Should we care here at all? Maybe just append if present

I agree, now that we know whether a port was specified, we don't need the reconstruction logic.

lukasniemeier-zalando: I wanted to keep the logic as-is in this PR, but I agree.

logbook-api/src/test/java/org/zalando/logbook/ForwardingHttpRequestTest.java
lukasniemeier-zalando: Promoted after sooo many successful test runs in the last quarter I assume.

whiskeysierra:
+1

PR #125, opened 2016-07-25
GH-124 Fixed async support for http client

### LINKED ISSUE #124: Incompatibility with Apache's Async HTTP Client
When using the async http client, as zalando/riptide is now doing as of version 2.x, we're failing with an exception when logging responses:

    java.lang.IllegalStateException: Content has not been provided​

I found [this post](http://mail-archives.apache.org/mod_mbox/hc-httpclient-users/201304.mbox/%3C1366275818.17900.23.camel@ubuntu%3E) from an old mailing list (2013, not sure if it still applies) and it states:

> HTTP messages passed to async message consumers contain full message
head (status line or request line plus all headers) and an HttpEntity
which is not attached to any data stream and is merely a descriptor.

It doesn't directly refer to interceptors, but it makes sense that it also applies.

After reading the docs of HttpResponseInterceptor it should be possible to delay the logging until someone provides the content:

> Interceptors can also manipulate content entities enclosed with messages.
Usually this is accomplished by using the 'Decorator' pattern where a wrapper
entity class is used to decorate the original entity.

Unfortunately the method interface is not making it obvious how to do that:

```java
 void process(HttpResponse response, HttpContext context)
    throws HttpException, IOException;
```

I guess this one could do the trick, although we need to be careful **not** to consume the underlying stream:

```java
/**
 * Associates a response entity with this response.
 * <p>
 * Please note that if an entity has already been set for this response and it depends on
 * an input stream ({@link HttpEntity#isStreaming()} returns {@code true}),
 * it must be fully consumed in order to ensure release of resources.
 *
 * @param entity    the entity to associate with this response, or
 *                  {@code null} to unset
 *
 * @see HttpEntity#isStreaming()
 * @see org.apache.http.util.EntityUtils#updateEntity(HttpResponse, HttpEntity)
 */
void setEntity(HttpEntity entity);
```

whiskeysierra:
Fixes #124

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,7 +321,7 @@ CloseableHttpAsyncClient client = HttpAsyncClientBuilder.create()
         .build();
         
 // and then wrap your response consumer
-client.client.execute(producer, new LogbookHttpAsyncResponseConsumer<>(consumer), callback)
+client.execute(producer, new LogbookHttpAsyncResponseConsumer<>(consumer), callback)
 ```
 
 ### Spring Boot Starter
--- a/README.md
+++ b/README.md
@@ -313,6 +313,17 @@ CloseableHttpClient client = HttpClientBuilder.create()
         .build();
 ```
 
+Since the `LogbookHttpResponseInterceptor` is incompatible with the `HttpAsyncClient` there is another way to log responses:
+
+```java
+CloseableHttpAsyncClient client = HttpAsyncClientBuilder.create()
+        .addInterceptorFirst(new LogbookHttpRequestInterceptor(logbook))
+        .build();
+        
+// and then wrap your response consumer
+client.client.execute(producer, new LogbookHttpAsyncResponseConsumer<>(consumer), callback)
+```
+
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -313,6 +313,17 @@ CloseableHttpClient client = HttpClientBuilder.create()
...
 
+Since the `LogbookHttpResponseInterceptor` is incompatible with the `HttpAsyncClient` there is another way to log responses:
+
+```java
+CloseableHttpAsyncClient client = HttpAsyncClientBuilder.create()
+        .addInterceptorFirst(new LogbookHttpRequestInterceptor(logbook))
+        .build();
+        
+// and then wrap your response consumer
+client.client.execute(producer, new LogbookHttpAsyncResponseConsumer<>(consumer), callback)
```
lukasniemeier-zalando: `client.client` ?

lukasniemeier-zalando:
:+1:

PR #126, opened 2016-08-04
unknown host exception in local request

tkrop:
Currently logbook will produce an unknown host exception whenever it cannot resolve the IP address of an host name.

```
Caused by: java.net.UnknownHostException: zalando: zalando: unknown error
	at java.net.InetAddress.getLocalHost(InetAddress.java:1505) ~[?:1.8.0_77]
	at org.zalando.logbook.httpclient.Localhost.getAddress(Localhost.java:29) ~[logbook-httpclient-1.0.0-RC3.jar:?]
	at org.zalando.logbook.httpclient.LocalRequest.getRemote(LocalRequest.java:86) ~[logbook-httpclient-1.0.0-RC3.jar:?]
	... 95 more
Caused by: java.net.UnknownHostException: zalando: unknown error
	at java.net.Inet4AddressImpl.lookupAllHostAddr(Native Method) ~[?:1.8.0_77]
	at java.net.InetAddress$2.lookupAllHostAddr(InetAddress.java:928) ~[?:1.8.0_77]
```

In this cases it may be helpful to provide a fallback to a the link local address to prevent local test servers from failing because of logbook.

### DIFF
--- a/logbook-httpclient/src/main/java/org/zalando/logbook/httpclient/LocalRequest.java
+++ b/logbook-httpclient/src/main/java/org/zalando/logbook/httpclient/LocalRequest.java
@@ -1,5 +1,17 @@
 package org.zalando.logbook.httpclient;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.apache.http.util.EntityUtils.toByteArray;
+
+import java.io.IOException;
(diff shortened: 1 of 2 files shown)


whiskeysierra:
:+1:

PR #131, opened 2016-09-09
Customization of HTTP/JSON output

### LINKED ISSUE #114: Un-finalize formatting class
I'm trying to use *logbook*, especially `httpclient` integration. However I want to customize output, here I want to add authenticated user id to track call origin (at least from which user `httpclient` has been triggered).

I was thinking about extending `JsonHttpLogFormatter` to add context user. But `JsonHttpLogFormatter` is `final`... 

Moreover if you plan to remove `final` from `JsonHttpLogFormatter`, it can be cool that `JsonHttpLogFormatter` has a method that returning `Map<String, Object>` or `JsonNode` to easily extend content.

---

PS it may be have a better way to achieve what I want, I'm open to any other solution. **How to achieve it is not really important, only thing I want is to add some additional informations**

whiskeysierra:
Fixes #114 
@kakawait

Adds public `prepare` and `format` methods to the `Default`- and `JsonHttpLogFormatter`. Prepare returns a data structure that can be modified (`List<String>` and `Map<String, Object>` respectively). The format method renders the corresponding data structure into a string.

The intended use case is to write a custom formatter that has an instance of one of those two and uses it to generate and customize the output.

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogFormatter.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogFormatter.java
@@ -2,10 +2,10 @@ package org.zalando.logbook;
 
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.function.Function;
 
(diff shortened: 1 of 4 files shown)


kakawait:
@whiskeysierra I will check that. But from what I just see I think it will match 100% of my use case!

Thanks for hard working on it :)

Issue #135, opened 2016-11-09
[httpclient] RemoteResponse.withBody() breaks gzip decompression


BGehrels:
`RemoteResponse.withBody()` replaces `response.getEntity()` with a newly constructed `ByteArrayEntity`. It does so by extracting the original entities content (`InputStream`), transforming it to a `byte[]` and then constructing the new `ByteArrayEntity` from it.

Sadly, the `contentStream` is not the only piece of data present in an `AbstractHttpEntity`: It also contains a `contentEncoding` field, which is lost when during copying.

The `contentEncoding` field is used by httpclient to determine if it must decompress the content. As a result, if you put the logbook interceptor as first interceptor in the chain, gziped requests will not be unziped any more.

whiskeysierra:
Do you have control over the ordering of the interceptors? I'd put the logbook interceptor at the last place (or at least after gzip), assuming we're talking about response interceptors here.

BGehrels:
Is probably a good idea anyway, because otherwise we will log compressed bodies. Yes, just tried it and it works as a workaround.

whiskeysierra:
> because otherwise we will log compressed bodies

Exactly!

I'll adjust the readme and clarify it in the *Known issues* secion.

whiskeysierra:
@BGehrels Would be willing to share a simple code snippet before and after your change here?

whiskeysierra:
Thanks for the snippets.

> Is there a good reason not to include the following in the RemoteResponse?

No, there isn't. If you want you can create a PR, otherwise I'll do it.

whiskeysierra:
@BGehrels Fixed in [1.0.1](https://github.com/zalando/logbook/releases/tag/1.0.1), thanks for the pull request :+1:

PR #136, opened 2016-11-11
Issue #135: Copy the chunked flag and the ContentEncoding/ContentType…

BGehrels:
… headers to the new ByteArrayEntity when replacing the response body

### DIFF
--- a/logbook-httpclient/src/test/java/org/zalando/logbook/httpclient/RemoteResponseTest.java
+++ b/logbook-httpclient/src/test/java/org/zalando/logbook/httpclient/RemoteResponseTest.java
@@ -11,8 +11,6 @@ import org.junit.rules.ExpectedException;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


logbook-httpclient/src/test/java/org/zalando/logbook/httpclient/RemoteResponseTest.java
```diff
@@ -2,12 +2,17 @@
...
+import org.apache.http.entity.BasicHttpEntity;
 import org.apache.http.message.BasicHttpResponse;
+import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.io.StringBufferInputStream;
```
whiskeysierra: I think this one is unused. Travis complains with a compiler error.

whiskeysierra:
:+1:

PR #137, opened 2016-12-02
Feature - Filter body

rubenjgarcia:


whiskeysierra:
Thanks for the pull request! I'll leave some inline comments.

logbook-core/src/main/java/org/zalando/logbook/BodyFilters.java
```diff
@@ -10,15 +10,15 @@
     }
 
     public static BodyFilter defaultValue() {
-        return accessToken();
+        return replaceProperty("access_token|open_id|id_token", "XXX");
```
whiskeysierra: A couple of things I noticed:
- having multiple properties separated by a pipe leaks the implementation detail that we're using a regular expression
- I'd propose to a parameter of type `Set<String>`
- property names are not escaped/quoted, i.e. it allows for "regex injection", see [`Pattern.quote(String)`](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#quote-java.lang.String-)
- so basically: `properties.stream().map(Pattern::quote).collect(joining("|"))`
- the whole idea only works for string properties, which makes it immediately less useful
- I'd rename it `replaceStringProperties`

logbook-core/src/main/java/org/zalando/logbook/BodyFilters.java
```diff
@@ -10,15 +10,15 @@
...
     }
 
-    public static BodyFilter accessToken() {
```
whiskeysierra: Removing this method is a breaking change. You can only add the new one next to it.

logbook-core/src/main/java/org/zalando/logbook/BodyFilters.java
```diff
@@ -10,15 +10,15 @@
...
+    public static BodyFilter replaceProperty(String property, String replacement) {
         final Predicate<String> json = MediaTypeQuery.compile("application/json", "application/*+json");
-        final Pattern pattern = Pattern.compile("(\"(?:access_token|open_id|id_token)\"\\s*\\:\\s*)\".+?\"");
+        final Pattern pattern = Pattern.compile("(\"(?:" + property + ")\"\\s*\\:\\s*)\".+?\"");
 
         return (contentType, body) -> json.test(contentType) ?
-                pattern.matcher(body).replaceAll("$1\"XXX\"") : body;
+                pattern.matcher(body).replaceAll("$1\"" + replacement + "\"") : body;
```
whiskeysierra: If somebody is using a valid group expression, e.g. `$1` it would also not behave as expected.

whiskeysierra:
> Since this implementation deals only with json, I am wondering should it be reflected in method name somehow

Yes, absolutely!

> would not it be better to use JSONPointer as an addressing approach?

That would require to parse the whole body, but yes, it would be the more universal solution.

whiskeysierra:
> JSONPointer

JSON Path would also be nice, since it would allow to obfuscate multiple things at once, e.g. `$.items[*].secret`

AlexanderYastrebov:
> JSON Path would also be nice, since it would allow to obfuscate multiple things at once, e.g. $.items[*].secret

Evaluating of JSONPath/Pointer approach should consider support for existing usecase "replace specific property regardless of node level"

jhorstmann:
In the json formatter we are already parsing the json body, maybe this could be generalized so we could do preprocessing on the parsed representation?

rubenjgarcia:
Hey guys

Thanks for the feedback. Do you want me to do these changes?

whiskeysierra:
> In the json formatter we are already parsing the json body

Not necessarily. We use use `@RawJsonValue` (see [here](https://github.com/zalando/logbook/blob/master/logbook-core/src/main/java/org/zalando/logbook/JsonHttpLogFormatter.java#L181)). We may parse it in order to compact the json, but that is not always done.

whiskeysierra:
> Do you want me to do these changes?

I think the most flexible way would be to have a method like this:

```java
static BodyFilter replaceJson(JsonPath path, String replacement);
```

That would allow to implement `accessToken()` on top of it:

```java
static BodyFilter accessToken() {
    return Arrays.asList(
        replaceJson("$..access_token", "XXX"),
        replaceJson("$..open_id", "XXX"),
        replaceJson("$..id_token", "XXX"))
        .stream().reduce(BodyFilter::merge).orElseThrow(..);
}
```

Unfortunately I'd require to add a new (optional?) dependency to get JsonPath. I'm not 100% sure whether this is worth it.

whiskeysierra:
> Evaluating of JSONPath/Pointer approach should consider support for pre-existing usecase "replace specific property regardless of node level"

JSONPath supports that: `$..access_token`. JSON Pointer doesn't.

AlexanderYastrebov:
JSONPointer is already available as a dependency which is good.
If pre-existing usecase do not require substitution on any level but only on known one then JSONPointer could be still an option.
Parsing json also is not for free, so the real question is it worth the effort to mess with json for filtering.

logbook-core/src/main/java/org/zalando/logbook/BodyFilters.java
```diff
@@ -14,11 +18,23 @@ public static BodyFilter defaultValue() {
...
 
     public static BodyFilter accessToken() {
+        Set<String> properties = new HashSet<>();
+        properties.add("access_token");
+        properties.add("open_id");
+        properties.add("id_token");
+        return replaceJSONProperty(properties, "XXX");
+    }
+
+    public static BodyFilter replaceJSONProperty(Set<String> properties, String replacement) {
```
whiskeysierra: `replaceJsonStringProperty`?

whiskeysierra:
:+1:

whiskeysierra:
@rubenjgarcia Thanks for the contribution! I released [1.1.0](https://github.com/zalando/logbook/releases/tag/1.1.0) and it should be available in central soon.

rubenjgarcia:
Thanks @whiskeysierra 

I need this feature for other project. Glad to contribute with this project

PR #147, opened 2017-03-20
Added chunking support

### LINKED ISSUE #142: Support chunking of large log messages
For rather big messages it could make sense to split up a message into multiple smaller chunks that can be logged individually. It would break single messages apart which decreases readability, but at least the whole content could be logged then.

whiskeysierra:
Fixes #142

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/ChunkingHttpLogWriter.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/ChunkingHttpLogWriter.java
@@ -4,15 +4,18 @@ import org.zalando.logbook.DefaultLogbook.SimpleCorrelation;
 import org.zalando.logbook.DefaultLogbook.SimplePrecorrelation;
 
 import java.io.IOException;
-import java.util.regex.Pattern;
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.lang.Math.min;
 
 public final class ChunkingHttpLogWriter implements HttpLogWriter {
 
-    private final Pattern pattern;
+    private final int size;
     private final HttpLogWriter writer;
 
     public ChunkingHttpLogWriter(final int size, final HttpLogWriter writer) {
-        this.pattern = Pattern.compile("(?<=\\G.{" + size + "})");
+        this.size = size;
         this.writer = writer;
     }
 
@@ -35,8 +38,14 @@ public final class ChunkingHttpLogWriter implements HttpLogWriter {
         }
     }
 
-    private String[] split(final String s) {
-        return pattern.split(s);
+    private Iterable<String> split(final String s) {
+        final List<String> parts = new ArrayList<>(s.length() / size + 1);
+
+        for (int i = 0; i < s.length(); i += size) {
+            parts.add(s.substring(i, min(s.length(), i + size)));
+        }
+        return parts;
     }
+
 
 }
--- a/logbook-core/src/main/java/org/zalando/logbook/ChunkingHttpLogWriter.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/ChunkingHttpLogWriter.java
@@ -10,6 +10,7 @@ import java.util.Spliterator;
 import java.util.function.Consumer;
 import java.util.stream.Stream;
 
+import static java.lang.Integer.signum;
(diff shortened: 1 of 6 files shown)


logbook-core/src/main/java/org/zalando/logbook/ChunkingHttpLogWriter.java
```diff
@@ -0,0 +1,42 @@
...
+import java.io.IOException;
+import java.util.regex.Pattern;
+
+public final class ChunkingHttpLogWriter implements HttpLogWriter {
+
+    private final Pattern pattern;
+    private final HttpLogWriter writer;
+
+    public ChunkingHttpLogWriter(final int size, final HttpLogWriter writer) {
+        this.pattern = Pattern.compile("(?<=\\G.{" + size + "})");
```
jhorstmann: This looks like a very inefficient way to split a string into fixed length subparts. A better alternative might be to use a `StringReader`.

logbook-core/src/main/java/org/zalando/logbook/ChunkingHttpLogWriter.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public void writeResponse(final Correlation<String, String> correlation) throws IOException {
+        for (final String part : split(correlation.getResponse())) {
+            writer.writeResponse(new SimpleCorrelation<>(correlation.getId(), correlation.getRequest(), part));
+        }
+    }
+
+    private Iterable<String> split(final String s) {
```
jhorstmann: Still creating the whole list up front, I was thinking about something like

    private void splitted(final String string, Consumer<String> consumer) throws IOException {
        final char[] buf = new char[size];
        int len;
        try (final StringReader r = new StringReader(string)) {
            while (-1 != (len = r.read(buf))) {
                consumer.accept(new String(buf, 0, len));
            }
        }
    }

Although you would need a `Consumer` interface that is allowed to throw `IOException` to make it this generic. Or just inline the code twice.

logbook-core/src/test/java/org/zalando/logbook/ChunkingHttpLogWriterTest.java
```diff
@@ -0,0 +1,103 @@
...
+        final List<String> precorrelation = captureResponse("Hello");
+        assertThat(precorrelation, contains("Hello"));
+
+    }
+
+    @Test
+    public void shouldWriteResponseInChunksIfLengthExceeded() throws IOException {
+        final List<String> precorrelation = captureResponse("Lorem ipsum dolor sit amet, consectetur adipiscing elit");
+        assertThat(precorrelation,
+                contains("Lorem ipsu", "m dolor si", "t amet, co", "nsectetur ", "adipiscing", " elit"));
```
AlexanderYastrebov: Given response body is usually a plain text 
And log is intended to be read by human
When logbook splits long line into chunks
Then split boundary should be (possibly) located at white-space, within say `size ± delta` interval

whiskeysierra: Good idea. Any idea how to do that efficiently?

AlexanderYastrebov: > Any idea how to do that efficiently?
Leap by size and look around within `± delta` for whitespace

AlexanderYastrebov: After thinking for a while, I have some more points:
1. Split boundary should be within `[size-delta, size]` so chunk is not longer then `size`
2. `delta` should be fixed size of e.g. to average "token" length. I think 10 to 16 characters should be fine
3. Split should happen on whitespaces as well as after common json punctuation characters `,:`

Having said that split algorithm could leap for `size` number of characters and look back `delta` characters for whitespaces and punctuation.

AlexanderYastrebov: I will raise a separate ticket

AlexanderYastrebov:
:+1:

PR #149, opened 2017-03-22
GH-148 Splits chunks on whitespace or punctuation

### LINKED ISSUE #148: Split long lines on whitespace
Given response body is usually a plain text
And log is intended to be read by human
When logbook splits long line into chunks
Then split boundary should be (possibly) located at white-space, within say size ± delta interval

1. Split boundary should be within `[size-delta, size]` so chunk is not longer then `size`
2. `delta` should be fixed size of e.g. to average "token" length. I think 10 to 16 characters should be fine
3. Split should happen on whitespaces as well as after common json punctuation characters `,:`

Having said that split algorithm could leap for `size` number of characters and look back `delta` characters for whitespaces and punctuation.

Imporovement on top of #142

AlexanderYastrebov:
Fixes #148

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/StringSpliterator.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/StringSpliterator.java
@@ -6,12 +6,28 @@ import java.util.function.Consumer;
 class StringSpliterator implements Spliterator<String> {
 
     private final String string;
-    private final int size;
+    private final int minChunkSize;
+    private final int maxChunkSize;
     private int position;
 
-    StringSpliterator(final String string, final int size) {
+    StringSpliterator(final String string, final int maxChunkSize) {
+        this(string, maxChunkSize, maxChunkSize);
+    }
+
+    StringSpliterator(final String string, final int minChunkSize, final int maxChunkSize) {
+        if (maxChunkSize <= 0) {
+            throw new IllegalArgumentException("maxChunkSize is expected to be greater than zero");
+        }
+        if (minChunkSize <= 0) {
+            throw new IllegalArgumentException("minChunkSize is expected to be greater than zero");
+        }
+        if (minChunkSize > maxChunkSize) {
+            throw new IllegalArgumentException("minChunkSize is expected to be less or equal to " + maxChunkSize);
+        }
+
         this.string = string;
-        this.size = size;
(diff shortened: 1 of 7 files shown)


logbook-core/src/main/java/org/zalando/logbook/StringSpliterator.java
```diff
@@ -0,0 +1,82 @@
+package org.zalando.logbook;
+
+import java.util.Spliterator;
+import java.util.function.Consumer;
+
+class StringSpliterator implements Spliterator<String> {
```
AlexanderYastrebov: Should it be renamed to `Chunk(ing)Spliterator`?

whiskeysierra: I like `ChunkingSpliterator`.

logbook-core/src/main/java/org/zalando/logbook/StringSpliterator.java
```diff
@@ -0,0 +1,82 @@
...
+            final char ch = string.charAt(i - 1);
+            if (isSplitCharacter(ch)) {
+                return string.substring(position, i);
+            }
+        }
+        return string.substring(position, chunkMax);
+    }
+
+    private static boolean isSplitCharacter(final char ch) {
+        return ch == ' ' || ch == ',' || ch == ':';
```
whiskeysierra: Could be a switch.

logbook-core/src/test/java/org/zalando/logbook/StringSpliteratorTest.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Test
+    public void shouldNotSupportPartitions() {
+        assertThat(new StringSpliterator("", 1).trySplit(), is(nullValue()));
+    }
+
+    @Test
+    public void shouldBeSizedWhenMinEqualToMax() {
+        assertTrue((new StringSpliterator("Hello", 5).characteristics() | SIZED) != 0);
```
whiskeysierra: There should probably be the inverse test for this as well.

whiskeysierra:
:+1:

whiskeysierra:
:+1:

PR #152, opened 2017-04-13
GH-151 Allows removal of headers

### LINKED ISSUE #151: Exclude some headers from output
Hello! Is there a way to log only some of the http headers(and not just obfuscating, literally exclude headers)?
Thanks

AlexanderYastrebov:
Fixes #151 

I have added a couple of factory methods into `HeaderFilters` to allow removal of headers, though I am not sure that it is a good idea to remove headers in general.

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/HeaderFilters.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/HeaderFilters.java
@@ -2,8 +2,9 @@ package org.zalando.logbook;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.function.BiFunction;
 import java.util.function.BiPredicate;
-import java.util.function.BinaryOperator;
 import java.util.function.Predicate;
 
 import static java.util.stream.Collectors.toList;
@@ -31,30 +32,27 @@ public final class HeaderFilters {
     }
 
     public static HeaderFilter removeHeaders(final Predicate<String> keyPredicate) {
-        return eachHeader((key, value) -> keyPredicate.test(key), (key, value) -> value);
+        return eachHeader((key, value) -> keyPredicate.test(key) ? null : value);
     }
 
     public static HeaderFilter removeHeaders(final BiPredicate<String, String> predicate) {
-        return eachHeader(predicate, (key, value) -> value);
+        return eachHeader((key, value) -> predicate.test(key, value) ? null : value);
     }
 
-    public static HeaderFilter eachHeader(final BinaryOperator<String> operator) {
-        return eachHeader((key, value) -> false, operator);
-    }
-
-    public static HeaderFilter eachHeader(final BiPredicate<String, String> remove, final BinaryOperator<String> change) {
+    public static HeaderFilter eachHeader(final BiFunction<String, String, String> transform) {
         return headers -> {
             final BaseHttpMessage.HeadersBuilder result = new BaseHttpMessage.HeadersBuilder();
 
             for (final Map.Entry<String, List<String>> e : headers.entrySet()) {
                 final String k = e.getKey();
                 result.put(k, e.getValue().stream()
-                        .filter(v -> !remove.test(k, v))
-                        .map(v -> change.apply(k, v))
+                        .map(v -> transform.apply(k, v))
+                        .filter(Objects::nonNull)
                         .collect(toList()));
             }
 
             return result.build();
         };
     }
+
 }
--- a/logbook-core/src/main/java/org/zalando/logbook/HeaderFilters.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/HeaderFilters.java
@@ -42,7 +42,7 @@ public final class HeaderFilters {
         return eachHeader((key, value) -> false, operator);
     }
 
-    private static HeaderFilter eachHeader(final BiPredicate<String, String> remove, final BinaryOperator<String> change) {
+    public static HeaderFilter eachHeader(final BiPredicate<String, String> remove, final BinaryOperator<String> change) {
         return headers -> {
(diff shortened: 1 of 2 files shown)


logbook-core/src/main/java/org/zalando/logbook/HeaderFilters.java
```diff
@@ -30,13 +31,24 @@ public static HeaderFilter replaceHeaders(final BiPredicate<String, String> pred
...
-    public static HeaderFilter eachHeader(final BinaryOperator<String> operator) {
+    public static HeaderFilter removeHeaders(final Predicate<String> keyPredicate) {
+        return eachHeader((key, value) -> keyPredicate.test(key) ? null : value);
+    }
+
+    public static HeaderFilter removeHeaders(final BiPredicate<String, String> predicate) {
+        return eachHeader((key, value) -> predicate.test(key, value) ? null : value);
+    }
+
+    public static HeaderFilter eachHeader(final BiFunction<String, String, String> transform) {
```
AlexanderYastrebov: This should be a safe change since `BinaryOperator extends BiFunction`, but now it will remove value if `transform` returns `null`

logbook-core/src/main/java/org/zalando/logbook/HeaderFilters.java
```diff
@@ -30,13 +29,24 @@ public static HeaderFilter replaceHeaders(final BiPredicate<String, String> pred
...
+        return eachHeader((key, value) -> predicate.test(key, value) ? null : value);
+    }
+
     public static HeaderFilter eachHeader(final BinaryOperator<String> operator) {
         return headers -> {
             final BaseHttpMessage.HeadersBuilder result = new BaseHttpMessage.HeadersBuilder();
 
             for (final Map.Entry<String, List<String>> e : headers.entrySet()) {
                 final String k = e.getKey();
-                result.put(k, e.getValue().stream().map(x -> operator.apply(k, x)).collect(toList()));
```
whiskeysierra: I'd prefer not to change the semantics of `eachHeader`.

AlexanderYastrebov: I understand, but is it really a change? And what should happen with null values then?

whiskeysierra: They should be rendered as defined by the corresponding formatter. e.g. `null` JSON.

AlexanderYastrebov: But header can not have null value. E.g. in servlet spec `getHeader()` returns null if there is no such header. My idea was that it you do nto want to have nulls there anyway so you may use it as a way to remove particular header.

whiskeysierra: The servlet-based implementation is only one of many. We also use Apache's HttpClient, e.g.

AlexanderYastrebov: I initially implemented it as a a pair of modify-operator and remove-predicate, but then I realized that it is always either modification or removal so I went with considering nulls as removal sign. I would say that null contract was no well documented, that is why it is not strictly a breaking change.

AlexanderYastrebov: But if we want nulls to be possible values, then we have to implement sister method to remove keys

whiskeysierra: > But if we want nulls to be possible values, then we have to implement sister method to remove keys

Exactly.

logbook-core/src/main/java/org/zalando/logbook/HeaderFilters.java
```diff
@@ -30,6 +30,27 @@ public static HeaderFilter replaceHeaders(final BiPredicate<String, String> pred
...
+
+    public static HeaderFilter removeHeaders(final BiPredicate<String, String> predicate) {
+        return headers -> {
+            final BaseHttpMessage.HeadersBuilder result = new BaseHttpMessage.HeadersBuilder();
+
+            for (final Map.Entry<String, List<String>> e : headers.entrySet()) {
+                for (final String value : e.getValue()) {
+                    final String key = e.getKey();
+                    if (!predicate.test(key, value)) {
+                        result.put(key, value);
```
AlexanderYastrebov: I would also change `eachHeader` to use `.forEach(v -> result.put(k, v))` and maybe name local variables here and there in the same style

logbook-core/src/main/java/org/zalando/logbook/HeaderFilters.java
```diff
@@ -30,6 +30,27 @@ public static HeaderFilter replaceHeaders(final BiPredicate<String, String> pred
...
+        return removeHeaders((key, value) -> keyPredicate.test(key));
+    }
+
+    public static HeaderFilter removeHeaders(final BiPredicate<String, String> predicate) {
+        return headers -> {
+            final BaseHttpMessage.HeadersBuilder result = new BaseHttpMessage.HeadersBuilder();
+
+            for (final Map.Entry<String, List<String>> e : headers.entrySet()) {
+                for (final String value : e.getValue()) {
+                    final String key = e.getKey();
```
AlexanderYastrebov: This should be moved out of cycle

AlexanderYastrebov: Values could be streamed like
```java
e.getValue().stream()
    .filter(value -> !predicate.test(key, value))
    .forEach(value -> result.put(key, value));
```

AlexanderYastrebov:
:+1:

PR #153, opened 2017-04-25
Delay calls to getOutputStream and getWriter

### LINKED ISSUE #146: getOutputStream() has already been called for this response
When I use Logbook auto configuration with Spring Security with Spring Boot app packaged as war and deployed to Tomcat and try to access non existent url, I get

```
Whitelabel Error Page

This application has no explicit mapping for /error, so you are seeing this as a fallback.

Tue Mar 07 18:14:22 CET 2017
There was an unexpected error (type=Internal Server Error, status=500).
getOutputStream() has already been called for this response
```

I have prepared minimal example to reproduce it [here](https://github.com/davidjirovec/Logbook-Spring-Security-Problem)

without Logbook it works as expected - basic auth is required
without Spring Security it works as expected - not found error message is shown
when packaged as jar with embedded Tomcat it works perfectly

The problem is not present only when accessing non existent url - if there would be any controllers in my example, you would not be able to access them neither.

whiskeysierra:
Fixes #146

### DIFF
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
@@ -28,8 +28,8 @@ final class LocalResponse extends HttpServletResponseWrapper implements RawHttpR
     private ByteArrayOutputStream bytes = new ByteArrayOutputStream();
     private DataOutput output = new DataOutputStream(bytes);
 
-    private ServletOutputStream stream = new TeeServletOutputStream();
-    private PrintWriter writer = new TeePrintWriter();
+    private ServletOutputStream stream;
+    private PrintWriter writer;
     private final String protocolVersion;
 
     /**
@@ -40,6 +40,8 @@ final class LocalResponse extends HttpServletResponseWrapper implements RawHttpR
 
     LocalResponse(final HttpServletResponse response, final String protocolVersion) throws IOException {
         super(response);
+        this.stream = new TeeServletOutputStream(output, super::getOutputStream);
+        this.writer = new TeePrintWriter(stream, getCharset());
         this.protocolVersion = protocolVersion;
     }
 
@@ -84,17 +86,23 @@ final class LocalResponse extends HttpServletResponseWrapper implements RawHttpR
         this.body = null;
         this.bytes = new ByteArrayOutputStream(0);
         this.output = new DataOutputStream(NULL);
-        this.stream = super.getOutputStream();
-        this.writer = super.getWriter();
+        this.stream = null;
+        this.writer = null;
     }
 
     @Override
     public ServletOutputStream getOutputStream() throws IOException {
+        if (stream == null) {
+            return super.getOutputStream();
+        }
         return stream;
     }
 
     @Override
     public PrintWriter getWriter() throws IOException {
+        if (writer == null) {
+            return super.getWriter();
+        }
         return writer;
     }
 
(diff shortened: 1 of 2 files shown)


logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
```diff
@@ -106,42 +114,54 @@ public PrintWriter getWriter() throws IOException {
...
+    private static final class TeeServletOutputStream extends ServletOutputStream {
 
-        private final OutputStream original;
+        private final DataOutput output;
+        private final LazyStream original;
 
-        private TeeServletOutputStream() throws IOException {
-            this.original = LocalResponse.super.getOutputStream();
+        private TeeServletOutputStream(final DataOutput output,
+                final LazyStream original) {
```
AlexanderYastrebov: `Supplier<OutputStream>` ?

whiskeysierra: The accessor throws an `IOException`.

AlexanderYastrebov: Then maybe rename it to 
```java
   @FunctionalInterface
    interface StreamSupplier {
        OutputStream get() throws IOException;
    }
```
since there is nothing lazy happening

whiskeysierra: The whole point of this supplier is to access the stream lazily. That's the cause of this bug actually, since we access it too early we currently fail.

AlexanderYastrebov: Then, to make it safe (i.e. client does not cache result of `retrieveLazily`) you may implement `class LazyStream extends OutputStream` and hide original there, so from POV of `TeeServletOutputStream` it is just another output stream.

AlexanderYastrebov: Also I think current implementation does not solve the core issue, i.e. getOutputStream could be called multiple times.

I would propose:
```diff
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
@@ -40,7 +40,7 @@ final class LocalResponse extends HttpServletResponseWrapper implements RawHttpR
 
     LocalResponse(final HttpServletResponse response, final String protocolVersion) throws IOException {
         super(response);
-        this.stream = new TeeServletOutputStream(output, super::getOutputStream);
+        this.stream = new TeeServletOutputStream(output, new LazyResponseOutputStream(response));
         this.writer = new TeePrintWriter(stream, getCharset());
         this.protocolVersion = protocolVersion;
     }
@@ -116,46 +116,76 @@ final class LocalResponse extends HttpServletResponseWrapper implements RawHttpR
 
     private static final class TeeServletOutputStream extends ServletOutputStream {
 
-        private final DataOutput output;
-        private final LazyStream original;
+        private final OutputStream first;
+        private final OutputStream second;
 
-        private TeeServletOutputStream(final DataOutput output,
-                final LazyStream original) {
-            this.output = output;
-            this.original = original;
+        private TeeServletOutputStream(final OutputStream first, final OutputStream second) {
+            this.first = first;
+            this.second = second;
         }
 
         @Override
         public void write(final int b) throws IOException {
-            output.write(b);
-            original().write(b);
+            first.write(b);
+            second.write(b);
         }
 
         @Override
         public void write(final byte[] b, final int off, final int len) throws IOException {
-            output.write(b, off, len);
-            original().write(b, off, len);
+            first.write(b, off, len);
+            second.write(b, off, len);
         }
 
         @Override
         public void flush() throws IOException {
-            original().flush();
+            first.flush();
+            second.flush();
         }
 
         @Override
         public void close() throws IOException {
-            original().close();
+            first.close();
+            second.close();
         }
 
-        private OutputStream original() throws IOException {
-            return original.retrieveLazily();
+    }
+
+    private static final class LazyResponseOutputStream extends OutputStream {
+
+        private final HttpServletResponse response;
+        private OutputStream stream;
+
+        LazyResponseOutputStream(final HttpServletResponse response) {
+            this.response = response;
         }
 
-    }
+        @Override
+        public void write(final int b) throws IOException {
+            getStream().write(b);
+        }
+
+        @Override
+        public void write(final byte[] b, final int off, final int len) throws IOException {
+            getStream().write(b, off, len);
+        }
+
+        @Override
+        public void flush() throws IOException {
+            getStream().flush();
+        }
+
+        @Override
+        public void close() throws IOException {
+            getStream().close();
+        }
+
+        private OutputStream getStream() throws IOException {
+            if (stream == null) {
+                stream = response.getOutputStream();
+            }
+            return stream;
+        }
 
-    @FunctionalInterface
-    public interface LazyStream {
-        ServletOutputStream retrieveLazily() throws IOException;
     }
 
     private static final class TeePrintWriter extends PrintWriter {

```

whiskeysierra: Calling it multiple times is ok, but calling both, `getOutputStream` and `getWriter` is not allowed.

AlexanderYastrebov: Or maybe just introducing flag `withoutBody` and postponing creation of `TeeServletOutputStream` until `LocalResponse.getOutputStream` is called would be even better way and will eliminate all this unobvious lazy stuff

AlexanderYastrebov: Or if `withBody` is required to be called anyway, `TeeServletOutputStream` creation could be moved there

whiskeysierra: Can you propose changes on top of this branch?

AlexanderYastrebov: I made another attempt, lets review https://github.com/zalando/logbook/pull/154

whiskeysierra:
Closed in favor of #154

PR #154, opened 2017-05-02
GH-146 Delegates getOutputStream and getWriter after withoutBody was called

### LINKED ISSUE #146: getOutputStream() has already been called for this response

AlexanderYastrebov:
Fixes #146

### DIFF
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
@@ -1,16 +1,6 @@
 package org.zalando.logbook.servlet;
 
-import org.zalando.logbook.HttpResponse;
-import org.zalando.logbook.Origin;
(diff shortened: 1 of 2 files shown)


whiskeysierra:
:+1:

PR #159, opened 2017-05-24
Cleaned up dependencies

### LINKED ISSUE #156: Consider removing 'com.google.gag' from dependencies
Hi! Probably not the most important stuff to do... but still... This library has nothing to do with the functionality itself. Was surprised to see "GAGs" on my classpath.

**Update:** **jcip-annotations** library seems to be yet another syntactic sugar.

whiskeysierra:
Fixes #156

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,9 +69,9 @@
 
         <slf4j.version>1.7.21</slf4j.version>
         <jackson.version>2.8.2</jackson.version>
-        <spring.version>4.3.2.RELEASE</spring.version>
-        <spring-boot.version>1.4.0.RELEASE</spring-boot.version>
-        <spring-security.version>4.1.3.RELEASE</spring-security.version>
+        <spring.version>4.3.8.RELEASE</spring.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -114,11 +114,13 @@
                 <groupId>com.google.code.findbugs</groupId>
                 <artifactId>jsr305</artifactId>
                 <version>3.0.1</version>
+                <scope>provided</scope>
             </dependency>
             <dependency>
                 <groupId>com.google.gag</groupId>
                 <artifactId>gag</artifactId>
                 <version>1.0.1</version>
+                <scope>provided</scope>
```
AlexanderYastrebov: Will not it fail if it is not provided on class path at runtime?

AlexanderYastrebov:
:+1:

pom.xml
whiskeysierra: There are only annotations in there and the java runtime allows annotations to be absent, i.e. not on the classpath.

PR #164, opened 2017-06-13
Fixed missing content-type

### LINKED ISSUE #143: There is no Content-Type header in response with async requests
I have a spring controller that is defined like this:

@GetMapping(path = "somePath" produces = MediaType.APPLICATION_JSON_VALUE)
public ListenableFuture<SomeClass> retrieveSomething() {
    ...
}

Without logbook it returns the content type header, with logbook autoconfiguration the header is missing in response.

I haven't done any debuging on this issue. So i can not provide more details. Sorry.

### LINKED ISSUE #163: logbook messy code
I have upload the contain-bug project to my repository
https://github.com/wyzssw/logbookbug
please just execute LogbookbugApplication in IDE, and open browse http://localhost:8080/test

whiskeysierra:
Fixes #163
Fixes #143 

- [x] Same for request?
- [x] Make sure we're null safe there

### DIFF
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
@@ -54,11 +54,6 @@ final class LocalResponse extends HttpServletResponseWrapper implements RawHttpR
     }
 
     @Override
-    public String getContentType() {
-        return Optional.ofNullable(super.getContentType()).orElse("");
-    }
-
-    @Override
     public Charset getCharset() {
         return Optional.ofNullable(getCharacterEncoding()).map(Charset::forName).orElse(UTF_8);
     }
--- a/logbook-api/src/main/java/org/zalando/logbook/BaseHttpMessage.java
+++ b/logbook-api/src/main/java/org/zalando/logbook/BaseHttpMessage.java
@@ -1,5 +1,6 @@
 package org.zalando.logbook;
 
+import javax.annotation.Nullable;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -17,6 +18,7 @@ public interface BaseHttpMessage {
 
     Map<String, List<String>> getHeaders();
 
+    @Nullable
     String getContentType();
 
     Charset getCharset();
--- a/logbook-api/src/main/java/org/zalando/logbook/BodyFilter.java
+++ b/logbook-api/src/main/java/org/zalando/logbook/BodyFilter.java
@@ -1,9 +1,11 @@
 package org.zalando.logbook;
 
+import javax.annotation.Nullable;
+
 @FunctionalInterface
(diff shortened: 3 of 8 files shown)


logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
```diff
@@ -54,11 +54,6 @@ public String getProtocolVersion() {
     }
 
     @Override
-    public String getContentType() {
-        return Optional.ofNullable(super.getContentType()).orElse("");
```
AlexanderYastrebov: The fix is not obvious from the first sight, I suppose it changes `""` to `null`. Could you add a test maybe?

whiskeysierra: What it did before was to change `null` to `""`. The not obvious issue is/was that `LocalResponse` implements two different interfaces, `org.zalando.logbook.BaseHttpMessage` and `javax.servlet.http.HttpServletResponse`, and both share a common method `String getContentType()`. By converting `null` to `""` we accidentially changed the behaviour of the later one because from what I can tell the container (or the framework?) checks whether the content type was set by checking for `null`. 

I have no clue how to test this other than running the example provided by @wyzssw.

AlexanderYastrebov: At least we may assert that:
```diff
diff --git a/logbook-servlet/src/test/java/org/zalando/logbook/servlet/LocalResponseTest.java b/logbook-servlet/src/test/java/org/zalando/logbook/servlet/LocalResponseTest.java
index e0e7b39..43d684c 100644
--- a/logbook-servlet/src/test/java/org/zalando/logbook/servlet/LocalResponseTest.java
+++ b/logbook-servlet/src/test/java/org/zalando/logbook/servlet/LocalResponseTest.java
@@ -7,8 +7,11 @@ import javax.servlet.http.HttpServletResponse;
 import org.junit.Before;
 import org.junit.Test;
 
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.nullValue;
 import static org.junit.Assert.assertNotSame;
 import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertThat;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
@@ -107,4 +110,10 @@ public class LocalResponseTest {
         unit.withBody();
     }
 
+    @Test
+    public void shouldReturnNullContentType() throws IOException {
+        when(mock.getContentType()).thenReturn(null);
+
+        assertThat(unit.getContentType(), is(nullValue()));
+    }
 }
```

AlexanderYastrebov: Ok, I have added an assertion test

AlexanderYastrebov:
:+1:

PR #166, opened 2017-06-20
Added curl request formatter

### LINKED ISSUE #160: logbook support log http request in CURL format
As a user, i want to print HttpClient request in CURL format which can easy to test by just execute at command line terminal, is this feature in your future plan？

whiskeysierra:
Fixes #160

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/CurlHttpLogFormatter.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/CurlHttpLogFormatter.java
@@ -32,6 +32,8 @@ public final class CurlHttpLogFormatter implements HttpLogFormatter {
         command.add("-X");
         command.add(request.getMethod());
 
+        command.add(quote(request.getRequestUri()));
+
         request.getHeaders().forEach((header, values) -> {
             values.forEach(value -> {
                 command.add("-H");
@@ -42,11 +44,9 @@ public final class CurlHttpLogFormatter implements HttpLogFormatter {
         final String body = request.getBodyAsString();
 
         if (!body.isEmpty()) {
-            command.add("-d");
+            command.add("--data-binary");
             command.add(quote(body));
         }
-
-        command.add(quote(request.getRequestUri()));
 
         return command.stream().collect(joining(" "));
     }
--- a/logbook-core/src/main/java/org/zalando/logbook/CurlHttpLogFormatter.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/CurlHttpLogFormatter.java
@@ -1,0 +1,67 @@
+package org.zalando.logbook;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+import static java.util.stream.Collectors.joining;
+
+/**
+ * Formats requests as cURL commands.
+ */
+public final class CurlHttpLogFormatter implements HttpLogFormatter {
+
+    private final HttpLogFormatter fallback;
+
+    public CurlHttpLogFormatter() {
+        this(new DefaultHttpLogFormatter());
+    }
+
+    public CurlHttpLogFormatter(final HttpLogFormatter fallback) {
+        this.fallback = fallback;
+    }
+
+    @Override
+    public String format(final Precorrelation<HttpRequest> precorrelation) throws IOException {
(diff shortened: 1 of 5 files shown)


wyzssw:
It's very good ! Expecting new release .

logbook-core/src/main/java/org/zalando/logbook/CurlHttpLogFormatter.java
```diff
@@ -0,0 +1,67 @@
...
+            command.add(quote(body));
+        }
+
+        command.add(quote(request.getRequestUri()));
+
+        return command.stream().collect(joining(" "));
+    }
+
+    private static String quote(final String s) {
+        return "'" + escape(s) + "'";
```
AlexanderYastrebov: In some cases quoting could be optional - e..g if you omit space after colon in headers and if there is no query string in url.

logbook-core/src/test/java/org/zalando/logbook/CurlHttpLogFormatterTest.java
```diff
@@ -0,0 +1,61 @@
...
+        final String correlationId = "c9408eaa-677d-11e5-9457-10ddb1ee7671";
+        final HttpRequest request = MockHttpRequest.create()
+                .withProtocolVersion("HTTP/1.0")
+                .withOrigin(Origin.REMOTE)
+                .withPath("/test")
+                .withQuery("limit=1")
+                .withHeaders(MockHeaders.of(
+                        "Accept", "application/json",
+                        "Content-Type", "text/plain"))
+                .withBodyAsString("Hello, world!");
```
AlexanderYastrebov: Could you add test with json payload which itself contains strings with single and double quotes?

logbook-core/src/main/java/org/zalando/logbook/CurlHttpLogFormatter.java
```diff
@@ -0,0 +1,67 @@
...
+            values.forEach(value -> {
+                command.add("-H");
+                command.add(quote(header + ": " + value));
+            });
+        });
+
+        final String body = request.getBodyAsString();
+
+        if (!body.isEmpty()) {
+            command.add("-d");
```
AlexanderYastrebov: I have checked 'Copy as cURL' in chrome dev tools:
1. They put URL in front of headers and body which makes sense
2. They use ` --data-binary` which also looks more proper.

logbook-core/src/main/java/org/zalando/logbook/CurlHttpLogFormatter.java
```diff
@@ -0,0 +1,67 @@
...
+    @Override
+    public String format(final Precorrelation<HttpRequest> precorrelation) throws IOException {
+        final HttpRequest request = precorrelation.getRequest();
+        final List<String> command = new ArrayList<>();
+
+        command.add("curl");
+        command.add("-v"); // TODO optional?
+
+        command.add("-X");
+        command.add(request.getMethod());
```
AlexanderYastrebov: I believe most requests are GETs so I woudl add shortcut here to omit `-X GET`

AlexanderYastrebov: Or better leave it for consistency

logbook-core/src/main/java/org/zalando/logbook/CurlHttpLogFormatter.java
whiskeysierra: I'd just always quote to be consistent. Also keeps the code simpler.

logbook-core/src/main/java/org/zalando/logbook/CurlHttpLogFormatter.java
whiskeysierra: Done.

logbook-core/src/test/java/org/zalando/logbook/CurlHttpLogFormatterTest.java
whiskeysierra: Done.

AlexanderYastrebov:
:+1:

whiskeysierra:
I'm still contemplating whether to put this into it's own module/artifact/repo outside of logbook, e.g. as a personal repo.

PR #167, opened 2017-06-27
Added possibility to delay logging of requests until response can be inspected

### LINKED ISSUE #165: How to log request/response based on response status
With a high load it could be wasteful to log all request/responses.
It should be possible to log only failed request responses, e.g. where response status is not 20x.

whiskeysierra:
As a side effect a writer now has access to the request/response object in case any decisions need to be taken based on them.

Fixes #165

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/ChunkingHttpLogWriter.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/ChunkingHttpLogWriter.java
@@ -23,7 +23,7 @@ public final class ChunkingHttpLogWriter implements HttpLogWriter {
         if (size <= 0) {
             throw new IllegalArgumentException("size is expected to be greater than zero");
         }
-        this.minChunkSize = (size > MIN_MAX_DELTA ? size - MIN_MAX_DELTA : size);
+        this.minChunkSize = size > MIN_MAX_DELTA ? size - MIN_MAX_DELTA : size;
         this.maxChunkSize = size;
         this.writer = writer;
     }
@@ -42,8 +42,13 @@ public final class ChunkingHttpLogWriter implements HttpLogWriter {
     @Override
     public void writeResponse(final Correlation<String, String> correlation) throws IOException {
         split(correlation.getResponse()).forEach(throwing(part ->
-                writer.writeResponse(new SimpleCorrelation<>(correlation.getId(), correlation.getDuration(),
-                        correlation.getRequest(), part))));
+                writer.writeResponse(new SimpleCorrelation<>(
+                        correlation.getId(),
+                        correlation.getDuration(),
+                        correlation.getRequest(),
+                        part,
+                        correlation.getOriginalRequest(),
+                        correlation.getOriginalResponse()))));
     }
 
     private static <T> Consumer<T> throwing(final ThrowingConsumer<T> consumer) {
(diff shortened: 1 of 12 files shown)


whiskeysierra:
@AlexanderYastrebov @lukasniemeier-zalando Thoughts?

logbook-core/src/main/java/org/zalando/logbook/DefaultLogbook.java
```diff
@@ -56,7 +56,7 @@
                 final Correlation<HttpRequest, HttpResponse> correlation =
                         new SimpleCorrelation<>(correlationId, duration, request, response);
                 final String message = formatter.format(correlation);
-                writer.writeResponse(new SimpleCorrelation<>(correlationId, duration, format, message));
```
AlexanderYastrebov: Would just different correlation type serve the need?
Lets say you introduce DelayedCorrelation (or extend simple correlation) which holds both request and response and DelayedLogWriter which does nothing on writeRequest but writes both on writeResponse thus you do not need to change `HttpLogWriter` interface

lukasniemeier-zalando: Or extend `Correlation` interface to expose `getOriginalRequest` and `getOriginalResponse`

whiskeysierra:
I updated it.

whiskeysierra:
@AlexanderYastrebov @lukasniemeier-zalando Can you take another look?

logbook-core/src/main/java/org/zalando/logbook/ChunkingHttpLogWriter.java
```diff
@@ -42,8 +42,13 @@ public void writeRequest(final Precorrelation<String> precorrelation) throws IOE
     @Override
     public void writeResponse(final Correlation<String, String> correlation) throws IOException {
         split(correlation.getResponse()).forEach(throwing(part ->
-                writer.writeResponse(new SimpleCorrelation<>(correlation.getId(), correlation.getDuration(),
-                        correlation.getRequest(), part))));
+                writer.writeResponse(new SimpleCorrelation<>(
```
AlexanderYastrebov: Why not to introduce SimpleCorrelation constructor which takes correlation as an argument?

whiskeysierra: Because I want to modify the body. `Correlation.withRequest(..)`?

logbook-core/src/test/java/org/zalando/logbook/DelayedResponseLogWriter.java
```diff
@@ -0,0 +1,32 @@
...
+        // do nothing
+    }
+
+    @Override
+    public void writeResponse(final Correlation<String, String> correlation) throws IOException {
+        final HttpResponse response = correlation.getOriginalResponse();
+
+        if (response.getStatus() >= 400) {
+            // delayed request logging until we have the response at hand
+            delegate.writeRequest(new SimplePrecorrelation<>(correlation.getId(), correlation.getRequest()));
```
AlexanderYastrebov: What is the difference between originalRequest and request and originalResponse and response?
Why do we need originals?

whiskeysierra: We use `Correlation<Request, Response>` as a type for both formatting and writing. In the first case it will be `Correlation<HttpRequest, HttpResponse>`, in the second `Correlation<String, String>`. The original request/response only makes sense for the later case.

whiskeysierra:
@lukasniemeier-zalando @AlexanderYastrebov Can you take another look?

lukasniemeier-zalando:
:+1:

lukasniemeier-zalando:
IMO this solution is more straight-forwarded compared to the previous proposal :+1:

PR #168, opened 2017-06-27
Added logging of durations

### LINKED ISSUE #116: Log duration of response

whiskeysierra:
Fixes #116

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,7 @@ Hello world!
 
 ```http
 Outgoing Response: 2d66e4bc-9a0d-11e5-a84c-1f39510f0d6b
+Duration: 25 ms
 HTTP/1.1 200
 Content-Type: application/json
 
@@ -219,6 +220,7 @@ Content-Type: application/json
   "origin": "local",
   "type": "response",
   "correlation": "2d66e4bc-9a0d-11e5-a84c-1f39510f0d6b",
+  "duration": 25,
   "protocol": "HTTP/1.1",
   "status": 200,
   "headers": {
@@ -236,6 +238,7 @@ a JSON response body will **not** be escaped and represented as a string:
   "origin": "local",
   "type": "response",
   "correlation": "2d66e4bc-9a0d-11e5-a84c-1f39510f0d6b",
+  "duration": 25,
   "protocol": "HTTP/1.1",
   "status": 200,
   "headers": {
--- a/logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogFormatter.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogFormatter.java
@@ -133,15 +133,16 @@ public final class DefaultHttpLogFormatter implements HttpLogFormatter {
         final int status = response.getStatus();
         final String reasonPhrase = REASON_PHRASES.getOrDefault(Integer.toString(status), "");
         final String statusLine = String.format("%s %d %s", response.getProtocolVersion(), status, reasonPhrase).trim();
-        return prepare(response, "Response", correlation.getId(), statusLine);
+        return prepare(response, "Response", correlation.getId(),
+                "Duration: " + correlation.getDuration().toMillis() + " ms", statusLine);
     }
 
     private <H extends HttpMessage> List<String> prepare(final H message, final String type,
-            final String correlationId, final String line) throws IOException {
+            final String correlationId, final String... prefixes) throws IOException {
         final List<String> lines = new ArrayList<>();
 
         lines.add(direction(message) + " " + type + ": " + correlationId);
-        lines.add(line);
+        Collections.addAll(lines, prefixes);
         lines.addAll(formatHeaders(message));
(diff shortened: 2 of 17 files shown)


logbook-core/src/main/java/org/zalando/logbook/DefaultLogbook.java
```diff
@@ -35,6 +39,7 @@
     @Override
     public Optional<Correlator> write(final RawHttpRequest rawHttpRequest) throws IOException {
         if (writer.isActive(rawHttpRequest) && predicate.test(rawHttpRequest)) {
+            final Instant start = Instant.now(clock);
```
AlexanderYastrebov: I was thinking about duration as about correlation duration which in fact it is.
Thus I would make correlation start time a property of correlation itself, i.e. refactor `String correlationId` into value object with uuid and start time.
I think it also makes sense to name it correlationTime or correlationDuration to explicitly note that it is not roundtrip time.

logbook-api/pom.xml
```diff
@@ -24,4 +24,11 @@
         <developerConnection>scm:git:git@github.com:zalando//logbook.git</developerConnection>
     </scm>
 
+    <dependencies>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
```
jhorstmann: Only needed in test scope, right?

whiskeysierra: It has the correct scope in the dependency management section in the parent.

logbook-core/src/main/java/org/zalando/logbook/DefaultLogbook.java
whiskeysierra: I rewrote it.

logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogFormatter.java
```diff
@@ -133,15 +133,16 @@ public String format(final Correlation<HttpRequest, HttpResponse> correlation) t
...
     }
 
     private <H extends HttpMessage> List<String> prepare(final H message, final String type,
-            final String correlationId, final String line) throws IOException {
+            final String correlationId, final String... prefixes) throws IOException {
         final List<String> lines = new ArrayList<>();
 
         lines.add(direction(message) + " " + type + ": " + correlationId);
-        lines.add(line);
+        Collections.addAll(lines, prefixes);
```
AlexanderYastrebov: Nice. Haven't seen that method before

logbook-core/src/main/java/org/zalando/logbook/DefaultLogbook.java
```diff
@@ -44,12 +49,14 @@
             writer.writeRequest(new SimplePrecorrelation<>(correlationId, format));
 
             return Optional.of(rawHttpResponse -> {
+                final Instant end = Instant.now(clock);
+                final Duration duration = Duration.between(start, end);
                 final RawHttpResponse filteredRawHttpResponse = rawResponseFilter.filter(rawHttpResponse);
                 final HttpResponse response = responseFilter.filter(filteredRawHttpResponse.withBody());
                 final Correlation<HttpRequest, HttpResponse> correlation =
-                        new SimpleCorrelation<>(correlationId, request, response);
+                        new SimpleCorrelation<>(correlationId, duration, request, response);
```
AlexanderYastrebov: Looks much cleaner now :+1:

whiskeysierra: Absolutely!

README.md
```diff
@@ -219,6 +220,7 @@ Content-Type: application/json
   "origin": "local",
   "type": "response",
   "correlation": "2d66e4bc-9a0d-11e5-a84c-1f39510f0d6b",
+  "duration": 25,
```
AlexanderYastrebov: ms implied.
I still feel taht we have to find better name for log message since `duration` could mislead users to think about total request/response duration. What about `correlation_duration`?

logbook-httpclient/src/main/java/org/zalando/logbook/httpclient/package-info.java
```diff
@@ -0,0 +1,4 @@
+@ParametersAreNonnullByDefault
+package org.zalando.logbook.httpclient;
+
+import javax.annotation.ParametersAreNonnullByDefault;
```
AlexanderYastrebov: newlines? :)

AlexanderYastrebov:
:+1:

README.md
whiskeysierra: I like shorter names. We may need to document all fields anyway.

logbook-httpclient/src/main/java/org/zalando/logbook/httpclient/package-info.java
whiskeysierra: Meh

PR #173, opened 2017-08-22
Reconstruct the body and ensure getParameterMap works when an x-www-f…

hanfak:
…orm-urlencoded form is submitted

### DIFF
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/NormalStrategy.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/NormalStrategy.java
@@ -29,6 +29,8 @@ final class NormalStrategy implements Strategy {
             final HttpServletResponse httpResponse, final FilterChain chain) throws ServletException, IOException {
 
         final RemoteRequest request = new RemoteRequest(httpRequest);
+        request.getParameterMap(); // force parameters to be parsed
+
         final Optional<Correlator> correlator = logRequestIfNecessary(logbook, request);
 
         if (correlator.isPresent()) {
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/RemoteRequest.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/RemoteRequest.java
@@ -18,8 +18,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import static java.lang.String.join;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Collections.list;
+import static java.util.stream.Collectors.joining;
 
 
 final class RemoteRequest extends HttpServletRequestWrapper implements RawHttpRequest, HttpRequest {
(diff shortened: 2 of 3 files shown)


hanfak:
We are not able to spend any more time on this, please feel free to modify as required.

logbook-servlet/src/main/java/org/zalando/logbook/servlet/RemoteRequest.java
```diff
@@ -82,22 +84,39 @@ public String getQuery() {
...
 
+    @SuppressWarnings("unchecked") // warnings appear when using Servlet API 2.5
+    @Override
+    public Map<String, String[]> getParameterMap() {
+        Map<String, String[]> parameterMap = super.getParameterMap();
+        if (getContentType() == null || !getContentType().toLowerCase().startsWith("application/x-www-form-urlencoded")) {
+            return parameterMap;
+        }
+        String formEncoded = parameterMap.entrySet().stream()
+                .map(entry -> entry.getKey() + "=" + join(",", entry.getValue()))
```
jhorstmann: Is separation by comma correct here? I think it should be multiple key=value pairs instead.

hanfak: Good point, thanks.

logbook-servlet/src/main/java/org/zalando/logbook/servlet/NormalStrategy.java
```diff
@@ -29,6 +29,8 @@ public void doFilter(final Logbook logbook, final HttpServletRequest httpRequest
             final HttpServletResponse httpResponse, final FilterChain chain) throws ServletException, IOException {
 
         final RemoteRequest request = new RemoteRequest(httpRequest);
+        request.getParameterMap(); // force parameters to be parsed
```
whiskeysierra: This won't allow any subsequent code to read the parameters from the body directly. I'm not sure whether we want to make that decision here.

logbook-servlet/src/main/java/org/zalando/logbook/servlet/RemoteRequest.java
```diff
@@ -82,22 +84,39 @@ public String getQuery() {
         return builder.build();
     }
 
+    @SuppressWarnings("unchecked") // warnings appear when using Servlet API 2.5
+    @Override
+    public Map<String, String[]> getParameterMap() {
+        Map<String, String[]> parameterMap = super.getParameterMap();
+        if (getContentType() == null || !getContentType().toLowerCase().startsWith("application/x-www-form-urlencoded")) {
```
whiskeysierra: I'd be more strict here and explicitly check for `application/x-www-form-urlencoded`:

```java
if ("application/x-www-form-urlencoded".equals(getContentType())) {
```

logbook-servlet/src/main/java/org/zalando/logbook/servlet/RemoteRequest.java
```diff
@@ -82,22 +84,39 @@ public String getQuery() {
...
+            return parameterMap;
+        }
+        String formEncoded = parameterMap.entrySet().stream()
+                .flatMap(entry -> stream(entry.getValue()).map(value -> entry.getKey() + "=" + value))
+                .collect(joining("&"));
+        body = formEncoded.getBytes(getCharset());
```
whiskeysierra: This is a rather nasty side-effect. Can't we move this decision somewhere else? Technically this would need to happen for all `getParameter*` methods, not just `getParameterMap`.

whiskeysierra:
Replaced by #184

PR #175, opened 2017-08-22
Make LogbookFilter non final in order to allow extension (this is use…

hanfak:
…ful to have a default constructor that can be used in a web.xml)


hanfak:
We are not able to spend any more time on this, please feel free to modify as required.

whiskeysierra:
Opening up a class to be subclassed introduces a whole new set of issues in itself. I'd rather go with composition here - something that is already doable right now, without any modification:

```java
public class CustomLogbookFilter {
    private final Filter delegate = new LogbookFilter(Logbook.builder().build());

    // TODO delegate methods
}
```

PR #178, opened 2017-08-29
Added Spring Boot 2.0 build profile

### LINKED ISSUE #176: Spring Boot 2.0.0 (milestone) incompatibility
Hello,

It appears that the latest release of logbook -- `1.4.0` -- is *incompatible* with the latest milestone release of spring boot 2.0 (`2.0.0.M3`), and this problem will persist into upcoming 2.0.0 release.

Specifically, `LogbookAutoConfiguration` has the following code:
```java
@AutoConfigureAfter({
        JacksonAutoConfiguration.class,
        WebMvcAutoConfiguration.class,
})
```
`WebMvcAutoConfiguration` is imported as
```java
import org.springframework.boot.autoconfigure.web.WebMvcAutoConfiguration;
```
however in spring boot 2, it has been moved to `org.springframework.boot.autoconfigure.web.servlet` package. The end result is that an app trying to use both spring boot 2.0.0.M3 and logbook 1.4.0 fails to start with
```
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'authorizedLogbookFilter' defined in class path resource [org/zalando/logbook/spring/LogbookAutoConfiguration$ServletFilterConfiguration.class]: Unsatisfied dependency expressed through method 'authorizedLogbookFilter' parameter 0; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'org.zalando.logbook.spring.LogbookAutoConfiguration': Bean instantiation via constructor failed; nested exception is java.lang.ArrayStoreException: sun.reflect.annotation.TypeNotPresentExceptionProxy
```

Obviously this is a backwards-incompatible change. How do you guys suggest to handle it? Is there a separate branch of logbook aligned with spring boot 2.0.0?

whiskeysierra:
Fixes #176

### DIFF
--- a/logbook-spring-boot-starter/pom.xml
+++ b/logbook-spring-boot-starter/pom.xml
@@ -51,6 +51,10 @@
             <optional>true</optional>
             <exclusions>
                 <exclusion>
+                    <groupId>org.springframework</groupId>
+                    <artifactId>spring-jcl</artifactId>
+                </exclusion>
+                <exclusion>
                     <groupId>aopalliance</groupId>
                     <artifactId>aopalliance</artifactId>
                 </exclusion>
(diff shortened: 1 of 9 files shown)


logbook-spring-boot-starter/pom.xml
```diff
@@ -121,6 +106,36 @@
         </dependency>
     </dependencies>
 
+    <profiles>
+        <profile>
+            <id>junit5</id>
```
whiskeysierra: This profile can be removed, once we fully switched to Spring Boot 2.x.

logbook-spring-boot-starter/src/main/java/org/zalando/logbook/spring/LogbookAutoConfiguration.java
```diff
@@ -242,6 +241,7 @@ public LogbookHttpResponseInterceptor logbookHttpResponseInterceptor() {
         @ConditionalOnMissingBean(name = FILTER_NAME)
         public FilterRegistrationBean authorizedLogbookFilter(final Logbook logbook) {
             final Filter filter = new LogbookFilter(logbook);
+            @SuppressWarnings("unchecked") // as of Spring Boot 2.x
```
whiskeysierra: `FilterRegistrationBean` is generic, as of Spring Boot 2.x.

whiskeysierra:
:+1:

whiskeysierra:
@eric239 Do you need this to be released to central? Otherwise I'd just leave this in here till the next release or an official release of Spring Boot 2.0. Whatever happens first.

PR #179, opened 2017-09-29
Fix NPE in header condition, when given header not found

innokenty:
The issue I ran into when using logbook, is that it was easy to check that a given header contains some value, but it was not so easy to check that a header was absent. Which I assume is quite often the case.

I tried to do this:
```java
header(MY_HEADER_NAME, Objects::isNull)
```
But got an NPE. This syntax seems totally correct, because this works, for example, when the header is there:
```java
header(MY_HEADER_NAME, StringUtils::isNotBlank)
```

I had to use this syntax to fulfill my goal, which looks a little bit too ugly for such a basic use-case:
```java
header((key, value) -> key.equals(MY_HEADER_NAME) && value == null)
```

Therefore I decided to fix this. Did I misunderstand something, or is it actually a good use case?

Anyway, here's a PR in a TDD style: first commit adds a breaking test, while second commit fixes the NPE in production code. Feel free to squash via GitHub when merging.

_P.S. Hi, Willi!_

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/Conditions.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/Conditions.java
@@ -7,6 +7,7 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 
 import static java.util.Collections.emptyList;
+import static java.util.Optional.ofNullable;
 import static org.zalando.logbook.RequestURI.Component.AUTHORITY;
 import static org.zalando.logbook.RequestURI.Component.PATH;
 import static org.zalando.logbook.RequestURI.Component.SCHEME;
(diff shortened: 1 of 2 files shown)


AlexanderYastrebov:
@innokenty Hi. If I understand correctly, you want to check presence of a header.
I would suggest adding a new method 
```java
    public static <T extends BaseHttpMessage> Predicate<T> header(final String key) {
        return message -> message.getHeaders().containsKey(key);
    }
```

IMO `Objects::isNull` or other kind of ad-hoc check looks clumsy.

whiskeysierra:
I agree with @AlexanderYastrebov. Checking for the presence of the key should be enough here.

whiskeysierra:
@innokenty You should always be able to tell absence apart from `null`. Treating them equal opens up a lot of possibilities for trouble. I'd suggest to implement the predicate you need as @AlexanderYastrebov proposed. I wouldn't consider this predicate to be useful enough to be part of the logbook core. Therefore I'm closing this PR now without merging it. 

Feel free to re-open if you have additional arguments/changes.

innokenty:
Well, I agree, this is all true what you said. But this PR is not about providing new functionality, but avoiding NPE in some cases. Which I think makes sense. It's not necessarily the presence of the header, that users want to check, it can be any other check. But they still will get an NPE instead of `false` if the header doesn't exist.

AlexanderYastrebov:
@innokenty Hi. You are tight. I have created https://github.com/zalando/logbook/pull/180 to deal with the issue you have discovered. Thanks!

PR #174, opened 2017-08-22
Initial sketch of a solution for https://github.com/zalando/logbook/i…

hanfak:
…ssues/172. We do not have a good way of testing against 2.5 in the build yet, suggestions welcome :)

PR #182, opened 2017-11-09
First draft of Servlet 2.5 support

### LINKED ISSUE #172: Add support for Servlet API 2.5
All versions of logbook depend on javax.servlet version 3. 

Can you please add support for version 2.5 of javax.servlet?

We cannot easily upgrade the servlet version.

whiskeysierra:
Fixes #172 
Replaces #174 

@hanfak I made your tests pass and changed the implementation from being reflection-based to two distinct implementations. Code is 100% covered, but I can't verify it reliably against servlet api 2.5 since that would require to rewrite almost all tests from the logbook-servlet module.

### DIFF
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
@@ -13,10 +13,10 @@ import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.nio.charset.Charset;
+import java.time.Instant;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
-import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -175,7 +175,7 @@ final class LocalResponse extends HttpServletResponseWrapper implements RawHttpR
     }
 
     private String format(final long date) {
-        return new Date(date).toInstant().atZone(ZoneId.of("GMT"))
+        return Instant.ofEpochMilli(date).atZone(ZoneId.of("GMT"))
                 .format(DateTimeFormatter.RFC_1123_DATE_TIME);
     }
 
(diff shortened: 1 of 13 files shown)


logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
```diff
@@ -1,22 +1,26 @@
...
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
 import java.nio.charset.Charset;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
```
AlexanderYastrebov: Is it used?

logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
```diff
@@ -108,6 +120,82 @@ private Tee tee() throws IOException {
...
+    @Override
+    public void sendError(final int status, final String msg) throws IOException {
+        this.status = status;
+        super.sendError(status, msg);
+    }
+
+    @Override
+    public void addDateHeader(final String name, final long date) {
+        super.addDateHeader(name, date);
+        addMultiMapHeader(name, format(date));
```
AlexanderYastrebov: Format could be simplified with `Instant.ofEpochMilli()` instead of `new Date()`

logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
whiskeysierra: SimpleDateFormat?

whiskeysierra: I organized imports again. Should be gone now.

AlexanderYastrebov: Yes, for me it is a red flag because it is not thread safe and java 8 has better alternative

whiskeysierra:
@hanfak I consulted with some colleagues about this. We took a look at the code and discussed what still would need to be done. The conclusion that we came to is that we don't see the value. It would increase the complexity of logbook disproportionately to the gained benefit.

PR #188, opened 2017-11-21
Added support to query for absent content type

### LINKED ISSUE #185: Request fails when using a streaming response in spring boot apps
Hello,
I am experimenting with logbook and wanted to try to see how it behaves with streams and asynchronous responses.
I have created a small example to show the issue [here](https://github.com/sgandon/test-zalando-stream)
When I set the `logbook.write.level=INFO` and call the endpoint this just breaks the request completely

The stack trace is quite big but we can see the following messages.

> Caused by: java.lang.IllegalStateException: getOutputStream() has already been called for this response
> java.nio.BufferOverflowException: null
> and some NPEs.

whiskeysierra:
Fixes #185

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/Conditions.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/Conditions.java
@@ -50,6 +50,10 @@ public final class Conditions {
                 query.test(message.getContentType());
     }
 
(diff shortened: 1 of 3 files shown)


AlexanderYastrebov:
:+1:

PR #191, opened 2017-11-24
Added support for more streaming media types

### LINKED ISSUE #190: Support other streaming media types
Should be treated equal to `application/x-json-stream`:

- [`application/stream+json`](https://tools.ietf.org/id/draft-snell-activity-streams-type-01.html)
-  [`text/event-stream`](https://tools.ietf.org/html/rfc6202)

whiskeysierra:
Fixes #190


AlexanderYastrebov:
:+1:

PR #201, opened 2018-02-06
Removed Localhost interface for resolving host address

### LINKED ISSUE #200: Logbook resolves localhost address via InetAddress.getLocalHost()
Due to the fact that logbook tries to resolve the localhost address via the java api (https://github.com/zalando/logbook/blob/master/logbook-httpclient/src/main/java/org/zalando/logbook/httpclient/Localhost.java#L9), all requests processed by logbook got slowdown by 5 seconds in my application (https://thoeni.io/post/macos-sierra-java/). I know this not a fail of logbook but maybe it's enough to just use static value to distinguish between local and remote requests.

whiskeysierra:
Fixes #200


AlexanderYastrebov:
:+1:

tkrop:
While this has now definitely not the same behavior as before it might even solve some other strange issues. As Java does lookup `localhost.<resolv.conf/serach>` it may even result in an address that is not the expected `localhost`.

tkrop:
👍

PR #203, opened 2018-02-09
Support for Feign logger

skjolber:
Feign logger support.

This PR does meet the code coverage requirements, but I'd like to get some comments on the general approach first. I was not able to work with the correlator-thing without a ThreadLocal. Perhaps a few changes could be made in the Feign project for a better overall result.

### DIFF
--- a/logbook-feign-logger/src/main/java/org/zalando/logbook/feignlogger/FeignLogger.java
+++ b/logbook-feign-logger/src/main/java/org/zalando/logbook/feignlogger/FeignLogger.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2013 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.zalando.logbook.feignlogger;
+
+import java.io.IOException;
+import java.util.Optional;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.zalando.logbook.Correlator;
+import org.zalando.logbook.Logbook;
+
+import feign.Request;
+import feign.Response;
+
+/**
+ * Logs to SLF4J. 
+ */
+public class FeignLogger extends feign.Logger {
+
+    /*
+     * 
+     * Implementation note: There is no way to set an attribute on a feign request, so using a thread local instead.
+     * 
+     */
+    
+    protected ThreadLocal<Optional<Correlator>> threadLocal = new InheritableThreadLocal<Optional<Correlator>>() {
+        
+        @Override
+        protected Optional<Correlator> childValue(Optional<Correlator> parentValue) {
+            return parentValue;
+        }
+    };
+    
+    private final Logger logger;
+    private final Logbook logbook;
+
+    
+    public FeignLogger(Logbook logbook) {
+        this(feign.Logger.class, logbook);
+    }
+
+    public FeignLogger(Class<?> clazz, Logbook logbook) {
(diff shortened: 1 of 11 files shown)


logbook-feign-logger/src/main/java/org/zalando/logbook/feignlogger/FeignLogger.java
```diff
@@ -0,0 +1,96 @@
...
+        if(optional != null) {
+            optional.ifPresent(r -> new RemoteResponse(feignResponse));
+        }
+
+        return feignResponse;
+    }
+
+    @Override
+    protected void log(String configKey, String format, Object... args) {
+        throw new RuntimeException(); // should never happen
```
AlexanderYastrebov: Why should it never happen?

logbook-feign-logger/src/main/java/org/zalando/logbook/feignlogger/Localhost.java
```diff
@@ -0,0 +1,18 @@
+package org.zalando.logbook.feignlogger;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+
+interface Localhost {
```
AlexanderYastrebov: This interface was removed recently. See https://github.com/zalando/logbook/pull/201/files

logbook-feign-logger/src/main/java/org/zalando/logbook/feignlogger/FeignLogger.java
AlexanderYastrebov: E.g `logRetry` calls it as far as I can see https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/Logger.java#L74

whiskeysierra:
Thanks for the proposal!

Have you considered using [Feign's Apache HttpClient support](https://github.com/OpenFeign/feign/tree/master/httpclient)? Because logbook already has support for that one.

We generally prefer support for the underlying HTTP libraries, instead of the abstraction, as it would benefit more users.

As far as I can tell Feign supports:
- [ ] [JDK's HttpUrlConnection](https://github.com/OpenFeign/feign/blob/e1ad7fc75d5abb8227e34db85d2f50034eb34c88/core/src/main/java/feign/Client.java#L58)
- [x] [Apache HttpClient](https://github.com/OpenFeign/feign/tree/master/httpclient)
- [ ] [OkHttp](https://github.com/OpenFeign/feign/tree/master/okhttp)
- [ ] [Ribbon](https://github.com/OpenFeign/feign/tree/master/ribbon)

Maybe we could target those, instead of Feign directly.

logbook-feign-logger/src/main/java/org/zalando/logbook/feignlogger/FeignLogger.java
skjolber: Right, thats a bug then.

logbook-feign-logger/src/main/java/org/zalando/logbook/feignlogger/FeignLogger.java
```diff
@@ -0,0 +1,96 @@
...
+            logger.error("Problem writing HTTP request", e);
+        }
+    }
+
+    @Override
+    protected Response logAndRebufferResponse(String configKey, Level logLevel, Response feignResponse, long elapsedTime) throws IOException {
+        Optional<Correlator> optional = threadLocal.get();
+        
+        if(optional != null) {
+            optional.ifPresent(r -> new RemoteResponse(feignResponse));
```
AlexanderYastrebov: Looks like correlator is not used properly here, it should be something like this: `r -> r.write(new RemoteResponse(feignResponse))`

I see couple of problems with thread local approach as well
- can we be sure that response logged in the same thread as request?
- there is no clean up code for thread local

skjolber: The logging is performed in [SynchronousMethodHandler](https://github.com/OpenFeign/feign/blob/57c609009c2e2e852ef5695b26cdd2b2238b0537/core/src/main/java/feign/SynchronousMethodHandler.java) so it seems fair enough, but perhaps creating a wrapper for the Optional<Correlator> and original request would be better. 

The Feign SLF4J logger does not allow for saving any fields on the request, at the same time there is overlap in the correlator inner class functionality (duration, linking request and response together). Perhaps saving the state via Feign Request would be best.

skjolber:
We're going change from HttpUrlConnection to Apache HttpClient, so that will let us use Feign+Logbook with the existing code.

PR #206, opened 2018-02-14
Added support for OkHttp

### LINKED ISSUE #204: Add support for OkHttp
See https://github.com/square/okhttp/wiki/Interceptors

Inspired by #203

whiskeysierra:
Fixes #204

### DIFF
--- a/logbook-okhttp/pom.xml
+++ b/logbook-okhttp/pom.xml
@@ -1,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <prerequisites>
+        <maven>3.0.4</maven>
+    </prerequisites>
+
+    <parent>
+        <groupId>org.zalando</groupId>
+        <artifactId>logbook-parent</artifactId>
+        <version>1.6.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>logbook-okhttp</artifactId>
+
+    <name>Logbook: OkHttp</name>
+    <description>HTTP Client interceptor for request and response logging</description>
+
+    <scm>
+        <url>https://github.com/zalando/logbook</url>
+        <connection>scm:git:git@github.com:zalando//logbook.git</connection>
+        <developerConnection>scm:git:git@github.com:zalando//logbook.git</developerConnection>
+    </scm>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.zalando</groupId>
+            <artifactId>logbook-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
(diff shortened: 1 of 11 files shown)


logbook-okhttp/src/main/java/org/zalando/logbook/okhttp/RemoteResponse.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Override
+    public int getStatus() {
+        return response.code();
+    }
+
+    @Override
+    public String getProtocolVersion() {
+        // TODO find the real thing
+        return "HTTP/1.1";
```
whiskeysierra: `return response.protocol().toString();` (+ to upper case?)

https://github.com/square/okhttp/blob/f23067f6ccba94db598485530d722fcb60f9e3e0/okhttp/src/main/java/okhttp3/Protocol.java#L41

logbook-okhttp/src/main/java/org/zalando/logbook/okhttp/LocalRequest.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    @Override
+    public String getHost() {
+        return request.url().host();
+    }
+
+    @Override
+    public Optional<Integer> getPort() {
+        // TODO empty for 80/443?
```
whiskeysierra: How are we dealing with this in `logbook-httpclient`?

whiskeysierra:
@lukasniemeier-zalando @AlexanderYastrebov Can you take a look?

logbook-okhttp/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+            <artifactId>logbook-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-nop</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.github.rest-driver</groupId>
+            <artifactId>rest-client-driver</artifactId>
```
AlexanderYastrebov: `<scope>test</scope>`

logbook-okhttp/src/main/java/org/zalando/logbook/okhttp/LogbookInterceptor.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    @Override
+    public Response intercept(final Chain chain) throws IOException {
+        final LocalRequest request = new LocalRequest(chain.request());
+        final Optional<Correlator> correlator = logbook.write(request);
+
+        final RemoteResponse response = new RemoteResponse(chain.proceed(request.toRequest()));
+
+        if (correlator.isPresent()) {
```
AlexanderYastrebov: `ifPresent`?

logbook-okhttp/pom.xml
whiskeysierra: Scope is specified in the parent pom, so I don't have to repeat myself all the time.

logbook-okhttp/src/main/java/org/zalando/logbook/okhttp/LogbookInterceptor.java
whiskeysierra: `write` throws an `IOException` and the functional interfaces don't allow checked exceptions.

AlexanderYastrebov:
:+1:

PR #209, opened 2018-02-23
 Update isFormRequest condition

whiskeysierra:
Replaces #208

### DIFF
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/MimeTypes.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/MimeTypes.java
@@ -4,6 +4,7 @@ import lombok.SneakyThrows;
 
 import javax.activation.MimeType;
 import javax.activation.MimeTypeParseException;
+import java.util.Optional;
 
 final class MimeTypes {
 
@@ -11,9 +12,12 @@ final class MimeTypes {
         // package private so we can trick code coverage
     }
 
-    @SneakyThrows(MimeTypeParseException.class)
-    static MimeType parse(final String mimeType) {
-        return new MimeType(mimeType);
+    static Optional<MimeType> parse(final String mimeType) {
+        try {
+            return Optional.of(new MimeType(mimeType));
+        } catch (final MimeTypeParseException e) {
+            return Optional.empty();
+        }
     }
 
 }
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/MimeTypes.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/MimeTypes.java
@@ -1,0 +1,19 @@
+package org.zalando.logbook.servlet;
+
+import lombok.SneakyThrows;
+
+import javax.activation.MimeType;
+import javax.activation.MimeTypeParseException;
+
+final class MimeTypes {
+
+    MimeTypes() {
+        // package private so we can trick code coverage
+    }
+
+    @SneakyThrows(MimeTypeParseException.class)
+    static MimeType parse(final String mimeType) {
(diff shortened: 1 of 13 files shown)


logbook-servlet/src/main/java/org/zalando/logbook/servlet/MimeTypes.java
```diff
@@ -0,0 +1,19 @@
...
+import javax.activation.MimeType;
+import javax.activation.MimeTypeParseException;
+
+final class MimeTypes {
+
+    MimeTypes() {
+        // package private so we can trick code coverage
+    }
+
+    @SneakyThrows(MimeTypeParseException.class)
```
AlexanderYastrebov: I am not sure it is a good idea to fail in case of invalid content type. Should it return `Optional<MimeType>` instead?

whiskeysierra: Done

logbook-servlet/src/main/java/org/zalando/logbook/servlet/RemoteRequest.java
```diff
@@ -117,13 +115,10 @@ public HttpRequest withBody() throws IOException {
...
         @Nullable final String contentType = getContentType();
 
-        if (contentType != null) {
-            final MimeType mimeType = MimeTypes.parse(contentType);
-            return "application".equals(mimeType.getPrimaryType()) &&
-                    "x-www-form-urlencoded".equals(mimeType.getSubType());
-        }
-
-        return false;
+        return contentType != null && MimeTypes.parse(contentType)
```
AlexanderYastrebov: You could do `Optional.ofNullable(contentType).map(MimeTypes::parse)` as well

logbook-servlet/src/test/java/org/zalando/logbook/servlet/MimeTypesTest.java
```diff
@@ -2,14 +2,14 @@
...
 
-import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
 final class MimeTypesTest {
 
     @Test
     void shouldFailToParse() {
-        assertThrows(MimeTypeParseException.class, () -> MimeTypes.parse(""));
+        assertEquals(MimeTypes.parse(""), Optional.empty());
```
AlexanderYastrebov: No happy case test?

logbook-servlet/src/main/java/org/zalando/logbook/servlet/RemoteRequest.java
whiskeysierra: :man_facepalming: right...

logbook-servlet/src/test/java/org/zalando/logbook/servlet/MimeTypesTest.java
whiskeysierra: The happy case is tested within the interceptor test

whiskeysierra: Done

logbook-servlet/src/main/java/org/zalando/logbook/servlet/RemoteRequest.java
whiskeysierra: Done

AlexanderYastrebov:
:+1: 

I guess `multipart/form-data` would be considered later on once needed

AlexanderYastrebov:
:+1:

AlexanderYastrebov:
:+1:

PR #214, opened 2018-03-09
Added original request to Precorrelation

### LINKED ISSUE #212: Can "Incoming Request" log under info-level and "Outgoing Request" log under debug-level
It is a limitation of that logbook use one level in all type http-request scene，and it  is also confused for user to config logbook  level in properties meantime coordinate with log4j2.xml or logback.xml

whiskeysierra:
Fixes #212


lukasniemeier-zalando:
:+1:

Issue #219, opened 2018-04-19
SecurityStrategy supporting 403


predragstefanovic:
Hey, using Logbook with Spring Boot and found that request/response is not logged in case of 403, while 401 is logged. How about extending [SecurityStrategy](https://github.com/zalando/logbook/blob/master/logbook-servlet/src/main/java/org/zalando/logbook/servlet/SecurityStrategy.java) to support 403 as well?

whiskeysierra:
Which layer generates the 403? Spring Security?

whiskeysierra:
@lukasniemeier-zalando Do you remember why we have special support for 401 here and not 403?

predragstefanovic:
Yes in this case it is coming from Spring Security layer, same one where 401s are generated.

whiskeysierra:
@predragstefanovic I don't have a lot of time at my hands right now. If you'd be willing to open a pull request that adds support for 403 incl. a test, I'd be more than willing to get this merged and deployed quickly.

whiskeysierra:
Released as 1.7.2.

PR #222, opened 2018-04-20
SecurityStrategy supporting 403 (#219)

fschlesinger:
Proposed changes to add support for logging requests and responses with HTTP 403 status in the default SpringBoot setup, as discussed in  #219.

### DIFF
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/SecurityStrategy.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/SecurityStrategy.java
@@ -31,7 +31,7 @@ final class SecurityStrategy implements Strategy {
 
         chain.doFilter(request, response);
 
-        if (isUnauthorized(response)) {
+        if (isUnauthorizedOrForbibben(response)) {
             final Optional<Correlator> correlator;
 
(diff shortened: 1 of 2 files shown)


logbook-servlet/src/main/java/org/zalando/logbook/servlet/SecurityStrategy.java
```diff
@@ -47,7 +47,8 @@ public void doFilter(final Logbook logbook, final HttpServletRequest httpRequest
     }
 
     private boolean isUnauthorized(final HttpServletResponse response) {
```
whiskeysierra: `isUnauthorizedOrForbidden`?

whiskeysierra:
> Is there any documentation to update?

I don't think so.

logbook-servlet/src/main/java/org/zalando/logbook/servlet/SecurityStrategy.java
fschlesinger: fixed

whiskeysierra:
:+1:

whiskeysierra:
@fschlesinger Thanks for the contribution :tada: I released the fix as 1.7.2

Issue #267, opened 2018-05-18
Add logbook.write.body_max_size as configuration parameters


boure:
Add `logbook.write.body_max_size` as a configuration so that the size of the logged body can be limited to a given number of characters.

## Context
Some of our services return response with body in the order of 10MBs, which can be a load on logging. While we are still interested in reading some of the body content, having this configuration would allow us to use logbook from the configuration only while avoiding over-large bodies (which can break the service).

whiskeysierra:
To clarify: Compared to chunking (what we already support), you're rather looking for truncating, right?

boure:
yes. Chunking is fine but it doesn't prevent display very large bodies in itself.

I hadn't considered that the whole log entry could be truncated - just the body part.
I guess truncating the entry would be more difficult and less desirable.

whiskeysierra:
Truncating the whole log message is actually easy. I believe both logback and log4j2 support it out of the box. But you're right, it's not desirable, since it breaks our format, eg JSON. Truncating just the body makes way more sense.

Would you be willing to propose a pull request? You can take the chunking feature as a blueprint on how to implement it.

boure:
:+1: will do.

PR #272, opened 2018-05-24
add a body max size config for Spring

boure:
This PR aims at adding a Spring configuration field to limit the size of the body of logged requests/responses.

Solves #267

## Description
This PR adds a field `logbook.write.body-max-size` to the Spring configuration with the following behavior:
- if field absent or null, use the default BodyFilter (which hides some fields)
- if n>=0, use a merge of the default filter (applied first), truncates the body up to n-th character and then appends '...'. If n=0, the entire body will be printed as '...'.

## Motivation and Context
This PR aims at giving control over how much of the request/response body should be printed, especially if it is expected to contain a large number of characters.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,6 +450,7 @@ The following tables show the available configuration:
 | `logbook.write.category`       | Changes the category of the [`DefaultHttpLogWriter`](#logger)    | `org.zalando.logbook.Logbook` |
 | `logbook.write.level`          | Changes the level of the [`DefaultHttpLogWriter`](#logger)       | `TRACE`                       |
 | `logbook.write.chunk-size`     | Splits log lines into smaller chunks of size up-to `chunk-size`. | `0` (disabled)                |
+| `logbook.write.body-max-size`  | Truncates the body up to `body-max-size` and appends `...`.      | `null` (disabled)             |
 
 ##### Example configuration
 
--- a/logbook-core/src/main/java/org/zalando/logbook/BodyFilters.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/BodyFilters.java
@@ -8,6 +8,7 @@ import java.util.function.Predicate;
 import java.util.regex.Pattern;
 
 import static java.util.stream.Collectors.joining;
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 import static org.apiguardian.api.API.Status.MAINTAINED;
 import static org.apiguardian.api.API.Status.STABLE;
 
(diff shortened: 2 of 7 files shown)


logbook-core/src/main/java/org/zalando/logbook/BodyFilters.java
```diff
@@ -61,4 +62,9 @@ public static BodyFilter replaceJsonStringProperty(final Set<String> properties,
                 pattern.matcher(body).replaceAll("$1\"" + replacement + "\"") : body;
     }
 
+    @API(status = EXPERIMENTAL)
+    public static BodyFilter truncatedBody(int maxSize) {
```
whiskeysierra: What about just calling it `truncate`?

README.md
```diff
@@ -450,6 +450,7 @@ The following tables show the available configuration:
 | `logbook.write.category`       | Changes the category of the [`DefaultHttpLogWriter`](#logger)    | `org.zalando.logbook.Logbook` |
 | `logbook.write.level`          | Changes the level of the [`DefaultHttpLogWriter`](#logger)       | `TRACE`                       |
 | `logbook.write.chunk-size`     | Splits log lines into smaller chunks of size up-to `chunk-size`. | `0` (disabled)                |
+| `logbook.write.body-max-size`  | Truncates the body up to `body-max-size` and appends `...`.      | `null` (disabled)             |
```
whiskeysierra: We should document somewhere that this will break the JSON-inlining feature of the JSON formatter.

logbook-spring-boot-starter/src/test/java/org/zalando/logbook/spring/WriteBodyMaxSizeTest.java
```diff
@@ -0,0 +1,28 @@
...
+    private BodyFilter bodyFilter;
+
+    @Test
+    void shouldUseBodyMaxSizeFilter() {
+        assertThat(bodyFilter.filter("application/json", "{\"foo\":\"someLongMessage\"}"),
+                is("{\"foo\":\"someLongMess..."));
+    }
+
+    @Test
+    void shouldUseBodyMaxSizeOverDefaultFilter() {
```
whiskeysierra: Nice test!

README.md
boure: alternatively, we could use a flag to enable it

whiskeysierra: A flag to enable JSON-inlining?

boure: Misunderstood your comment. Nevermind.

whiskeysierra: Or we find a way to detect within the JSON formatter that truncating is active and skip inlining then.

whiskeysierra:
:+1:

whiskeysierra:
@boure In the interest of progress, I decided to merge this pull request in its current state. I have a follow up at #274 that fixes some minor issues that were not worth bringing up in this review (although noteworthy: I renamed the key to `max-body-size`), since the feedback loop usually takes too long. Feel free to comment there.

whiskeysierra:
@boure Thanks for the contribution! :tada: I'll let you know as soon as there is a new release available.

PR #280, opened 2018-05-31
Added JSON heuristic

### LINKED ISSUE #279: JSON formatted logs with truncated body are not valid JSON anymore
## Description

By default the body of a request that is not authenticated is replaced with `<skipped>`. This is currently formatted in the logs as `{..., "headers":{...},"body":<skipped>}` which is not valid JSON.

This is an issue when you attempt to parse the lines as JSON in order to get structured details about the content.

## Expected Behavior

The line with a skipped body should be formatted as `{...,"body":"<skipped>"}`.

## Actual Behavior

See above.

## Possible Fix

I had a look at `JsonHttpLogFormatter`'s `tryParseBodyAsJson`. The issue seems to be that the string should be considered invalid JSON and the method should simply return the string as is. However, the `compactJson` method terminates early because `isAlreadyCompacted` only looks for the existence of new lines but not whether it is valid JSON at all. The string is then simply returned and later serialized as a raw value via `JsonBody` which foregoes adding the quotes.

I guess a fix could be to also check for valid JSON syntax in `tryParseBodyAsJson` or similar but I'm not sure what the best approach would be. Maybe the `isAlreadyCompacted` check can be removed; I don't know how much performance it saves. Parsing the content beforehand for syntactical validity might lead to duplicate parsing. Not sure.

## Steps to Reproduce

1. configure application to use `JsonHttpLogFormatter`
2. send unauthenticated request with payload

## Your Environment

* Version used: 1.8.0
* Link to your project:

whiskeysierra:
## Motivation and Context

Fixes #279

## Types of changes

## Checklist:

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/JsonCompactor.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/JsonCompactor.java
@@ -1,0 +1,42 @@
+package org.zalando.logbook;
+
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import lombok.AllArgsConstructor;
+
+import java.io.IOException;
+import java.io.StringWriter;
+
+@AllArgsConstructor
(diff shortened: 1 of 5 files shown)


lukasniemeier-zalando:
:+1:

PR #288, opened 2018-06-14
Added bill of materials

whiskeysierra:
## Description

Logbook has already 5+ modules that produce individual artifacts. Users that depend on them will usually list them all as dependencies in their POMs. Right now we guarantee that version numbers across modules are always the same, i.e. a new release will release each module, even if it wasn't change.
That allows users right now already to use a single maven property `logbook.version` that is used for each dependency. 

This pull requests introduces a BOM (bill of material) as an alternative. The BOM can be imported in the `dependencyManagement` section of a POM. 

## Motivation and Context

The BOM removes the need to specify any version on individual logbook artifacts.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Logbook is ready to use out of the box for most common setups. Even for uncommon
 
 ## Installation
 
-Selectively add the following dependencies to your project:
+Add the following dependency to your project:
 
 ```xml
 <dependency>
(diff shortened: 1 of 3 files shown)


lukasniemeier-zalando:
:+1:

whiskeysierra:
Shouldn't the version be inherited from the parent?

rainermueller:
Compared to `org.springframework:spring-framework-bom` and others every dependency defines the version.

whiskeysierra:
Would you mind creating a pull request?

rainermueller:
Done, see [Update bill of materials (BOM) to define version for each dependency](https://github.com/zalando/logbook/pull/379)

PR #300, opened 2018-06-27
add JAX-RS support (on client and server) for logbook

davidwcarlson:
Support for using Logbook with JAX-RS clients and servers via filters

## Motivation and Context

Support for a Java standard API (JAX-RS)

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ Logbook is ready to use out of the box for most common setups. Even for uncommon
 - OkHttp (optional)
 - Spring 4.x **or 5.x** (optional)
 - Spring Boot 1.x **or 2.x** (optional)
+- JAX-RS 2.x Client and Server (optional)
 
 ## Installation
 
@@ -89,6 +90,10 @@ Alternatively, you can import our *bill of materials*...
 <dependency>
     <groupId>org.zalando</groupId>
     <artifactId>logbook-spring-boot-starter</artifactId>
+</dependency>
+<dependency>
+    <groupId>org.zalando</groupId>
+    <artifactId>logbook-jaxrs</artifactId>
 </dependency>
 ```
 
@@ -495,6 +500,18 @@ logbook:
         category: http.wire-log
         level: INFO
         chunk-size: 1000
+```
+
+### JAX-RS
+
+The `logbook-jaxrs` module contains:
+ - a `ClientLoggingFilter` for use with applications making HTTP requests
+```java
+  client.register(new ClientLoggingFilter(logbook));
+```
+ - a `ServerLoggingFilter` for use with HTTP servers
+```java
+  resourceConfig.register(new ServerLoggingFilter(logbook));
 ```
 
 ## Known Issues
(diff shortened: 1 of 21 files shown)


logbook-jaxrs/src/test/java/org/zalando/logbook/jaxrs/ClientAndServerTest.java
```diff
@@ -0,0 +1,330 @@
...
+        // clientRequest
+        () -> assertEquals("GET", clientRequest.getMethod(), "clientRequest method"),
+        () -> assertEquals(LOCAL, clientRequest.getOrigin(), "clientRequest origin"),
+        () -> assertEquals("localhost", clientRequest.getRemote()),
+        () -> assertEquals("http", clientRequest.getScheme(), "clientRequest scheme"),
+        () -> assertEquals("localhost", clientRequest.getHost(), "clientRequest host"),
+        () -> assertEquals(Optional.of(this.getPort()), clientRequest.getPort()),
+        () -> assertEquals("/testws/testGet/first/textPlain", clientRequest.getPath(), "clientRequest path"),
+        () -> assertEquals("param2=second", clientRequest.getQuery(), "clientRequest query"),
+        () -> assertEquals("Jersey/2.27 (HttpUrlConnection 1.8.0_131)", clientRequest.getHeaders().get("User-Agent").get(0), "clientRequest userAgent"),
```
whiskeysierra: You should remove version numbers here. It fails on travis right now, because it's running a different java version.

whiskeysierra:
Thanks for the contribution! :tada: 

Couple of minor issues that we need to address before we can get this merged and released:
- [ ] license headers
- [ ] codacy (unused imports only, as far as I can see)
- [ ] passing tests (user agent contains concrete java and jersey version right now)

Question: You do use jersey for your tests, but I assume that the module will work with any JAX-RS implementation?

davidwcarlson:
I pushed a few updates to address the issues you mentioned and to put all the Jersey-specific dependencies in test scope.

My intent is that this module be dependent only on JAX-RS, as that should be sufficient.  In order to test, I use Jersey as an implementation since that is what I am most familiar.  I started looking adding a second implementation for another test, but it may take some time to find one.  Also, I wonder whether it would be useful to split the library, jersey and any other implementations into separate sub-modules.

whiskeysierra:
> My intent is that this module be dependent only on JAX-RS, as that should be sufficient.

Great.

> In order to test, I use Jersey as an implementation since that is what I am most familiar.

Makes sense.

> I started looking adding a second implementation for another test, but it may take some time to find one.

I believe jersey is enough for the test. If someone is experiencing issues with resteasy or any other implementation, than we can take a look into this.

> Also, I wonder whether it would be useful to split the library, jersey and any other implementations into separate sub-modules.

I was expecting that there is no jersey-specific code, just jaxrs and therefore no need to have any other module.

whiskeysierra:
:+1:

PR #347, opened 2018-09-11
fix BodyFilter.replaceJsonStringProperty issue with empty values

nsmolenskii:
## Description

BodyFilter `BodyFilters#replaceJsonStringProperty` corrupts json.

## Motivation and Context

When `BodyFilters#replaceJsonStringProperty` filters json containing empty properties to replace it corrupt json. Example:

Filter: 
```java
final BodyFilter unit = BodyFilters.replaceJsonStringProperty(Collections.singleton("foo"), "XXX");
```

Input: 
```json
{"foo":"","bar":"public"}
```
Output: 
```json
{"foo":"XXX"bar":"public"}
```
## Types of changes

## Checklist:

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/BodyFilters.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/BodyFilters.java
@@ -6,6 +6,7 @@ import java.util.HashSet;
 import java.util.Set;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
+import java.util.stream.StreamSupport;
 
 import static java.util.stream.Collectors.joining;
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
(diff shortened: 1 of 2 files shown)


logbook-core/src/main/java/org/zalando/logbook/BodyFilters.java
```diff
@@ -50,13 +51,14 @@ public static BodyFilter accessToken() {
      * @return BodyFilter generated
      */
     @API(status = MAINTAINED)
-    public static BodyFilter replaceJsonStringProperty(final Set<String> properties, final String replacement) {
-        final String regex = properties.stream()
+    public static BodyFilter replaceJsonStringProperty(final Iterable<String> properties, final String replacement) {
+        final String regex = StreamSupport.stream(properties.spliterator(), false)
```
whiskeysierra: Why are you using StreamSupport here?

whiskeysierra: Or rather *Why are you switching from Set to Iterable?*

nsmolenskii: Well, I always prefer to use `Iterable<T>` if that allows to do everything I want it to. 
It's more flexible for callers and it allows to keep method contract simple.

nsmolenskii: That's is actually a trade of between semantics of `Set<T>` and flexibility of `Iterable<T>`.
Changes in method signature could be removed, if it makes no sense. Should I ?

whiskeysierra: I believe it's a breaking API change anyway.

To me Set has the obvious benefit of highlighting to callers what I'm expecting here, i.e. no duplicates, since that wouldn't make any sense.

whiskeysierra:
Thanks for your pull request :tada:! I released the fix as 1.9.1 to central.

PR #357, opened 2018-10-07
Add compacting body filters

nsmolenskii:
## Description

Add body filters to compact most common content types

## Motivation and Context

Custom `HttpLogForrmatter`s (different from `JsonHttpLogFormatter`) have also need to minify message body. To reduce code duplication from formatter to formatter, it could be extracted as delegating body filters.

## Types of changes

## Checklist:

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/XmlCompactingBodyFilter.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/XmlCompactingBodyFilter.java
@@ -1,0 +1,74 @@
+package org.zalando.logbook;
+
+import lombok.extern.slf4j.Slf4j;
+import org.w3c.dom.Document;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
+import javax.annotation.Nullable;
+import javax.xml.parsers.DocumentBuilderFactory;
+import javax.xml.transform.Transformer;
+import javax.xml.transform.TransformerFactory;
+import javax.xml.transform.dom.DOMSource;
+import javax.xml.transform.stream.StreamResult;
+import javax.xml.xpath.XPath;
+import javax.xml.xpath.XPathFactory;
+import java.io.ByteArrayInputStream;
+import java.io.StringWriter;
+import java.util.function.Predicate;
+
+import static javax.xml.transform.OutputKeys.INDENT;
+import static javax.xml.transform.OutputKeys.OMIT_XML_DECLARATION;
+import static javax.xml.xpath.XPathConstants.NODESET;
+import static org.zalando.fauxpas.FauxPas.throwingSupplier;
+
+@Slf4j
+class XmlCompactingBodyFilter implements BodyFilter {
+
+    private final Predicate<String> contentTypes = MediaTypeQuery.compile("*/xml", "*/*+xml");
+    private final Transformer transformer = transformerFactory();
+
+    @Override
+    public String filter(@Nullable final String contentType, final String body) {
+        return contentTypes.test(contentType) ? compact(body) : body;
+    }
+
+    private String compact(final String body) {
+        if (body.trim().isEmpty()) return body;
+        try {
+            final StringWriter output = new StringWriter();
+            final Document document = documentWithoutTextNodes(body);
+            transformer.transform(new DOMSource(document), new StreamResult(output));
+            return output.toString();
+        } catch (Exception e) {
+            log.trace("Unable to compact body, is it a XML?. Keep it as-is: `{}`", e.getMessage());
+            return body;
+        }
+    }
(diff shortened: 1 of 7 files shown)


logbook-core/src/main/java/org/zalando/logbook/XmlCompactingBodyFilter.java
```diff
@@ -0,0 +1,74 @@
...
+            final Document document = documentWithoutTextNodes(body);
+            transformer.transform(new DOMSource(document), new StreamResult(output));
+            return output.toString();
+        } catch (Exception e) {
+            log.trace("Unable to compact body, is it a XML?. Keep it as-is: `{}`", e.getMessage());
+            return body;
+        }
+    }
+
+    private Document documentWithoutTextNodes(final String body) {
```
jhorstmann: This seems rather inefficient, a custom sax or stax parser should have much less overhead.

There are also some security implication when parsing the request payload, see <https://en.wikipedia.org/wiki/XML_external_entity_attack> and <https://www.owasp.org/index.php/XML_External_Entity_(XXE)_Prevention_Cheat_Sheet#JAXP_DocumentBuilderFactory.2C_SAXParserFactory_and_DOM4J> for how to prevent this in java.

nsmolenskii: I would agree with a fact, that sax/stax parsing will be more effective than dom in general.
Unfortunately, potential benefit from introducing custom parser with supporting rules will be miserable, due to fact of parsing already allocated in memory string .

nsmolenskii: I had applied recommended changes, according to XXE

whiskeysierra:
> To reduce code duplication from formatter to formatter, it could be extracted as delegating body filters.

But from what I see there is nothing shared, is there?

nsmolenskii:
Well, it shared not as part of every single formatter, but as part of infrastructural body filters.
E.g.: 
* you have json http log formatter, but some of your endpoints are xml based, you will receive minified xml, as soon as xml compacting body filter will be configured.
* you introduces another http log formatter (like #358) and you don't have to repeat logic to minify json or xml, or your custom format

nsmolenskii:
> I failed to realize that you moved the compacting from the formatter to the filter.
> 
> We could think about removing it from the `JSONHttpLogFormatter` in favor of registering those two filters by default, but that would be a breaking change. Maybe something for 2.x

Due to backward compatibility (`JsonHttpLogFormatter` marked as `STABLE`), I was not going into that direction myself. But I also find it would be a proper change.

logbook-core/src/main/java/org/zalando/logbook/XmlCompactingBodyFilter.java
```diff
@@ -0,0 +1,100 @@
...
+    private final DocumentBuilderFactory documentBuilderFactory = documentBuilderFactory();
+    private final Transformer transformer = transformer();
+
+    @Override
+    public String filter(@Nullable final String contentType, final String body) {
+        return contentTypes.test(contentType) && shouldCompact(body) ? compact(body) : body;
+    }
+
+    private boolean shouldCompact(final String body) {
+        return Stream.of("<?xml", "\n", "  ").anyMatch(body::contains);
```
whiskeysierra: The xml declaration is redundant and should be covered by the content type check, shouldn't it? Spaces would be generally ok, since the whole log message would still fit into a single line.

I'd just change this in the same way as the json compactor:

```
return body.indexOf('\n') != -1;
```

nsmolenskii: It's quite possible to have `xml prolog` inside of the message, but do not have `\n` inside.
So, to not check `<?xml` potentially means to keep it in compacted version (due to compact skip)

nsmolenskii:
So do we have some open tasks, preventing that PR ?

logbook-core/src/main/java/org/zalando/logbook/XmlCompactingBodyFilter.java
whiskeysierra: But if there are no newlines, what exactly would you expect to happen when compacting? Removing other whitespaces?

nsmolenskii: Well I got your point. We have different points of view to minimisation. I consider extra whitespaces and xml prolog as too verbose in logs. But I could leave with, cause single line is much more important

whiskeysierra:
I released your changes as 1.10.0. :tada:

PR #367, opened 2018-10-12
add remove query filter

nsmolenskii:
## Description

Add query filter, that allows removing some query parameters

## Motivation and Context

Support  removing some redundant query parameters:
* too verbose names
* too few meaning
*  debug params

## Types of changes

## Checklist:

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/QueryFilters.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/QueryFilters.java
@@ -5,6 +5,7 @@ import org.apiguardian.api.API;
 import java.util.regex.Pattern;
 
 import static java.util.regex.Pattern.quote;
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 import static org.apiguardian.api.API.Status.MAINTAINED;
 import static org.apiguardian.api.API.Status.STABLE;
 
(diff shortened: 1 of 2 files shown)


logbook-core/src/main/java/org/zalando/logbook/QueryFilters.java
```diff
@@ -33,4 +32,10 @@ public static QueryFilter replaceQuery(final String name, final String replaceme
         return query -> pattern.matcher(query).replaceAll(replacementPattern);
     }
 
+    @API(status = MAINTAINED)
+    public static QueryFilter removeQuery(final String name) {
```
whiskeysierra: Can you describe your use case a bit more?

logbook-core/src/main/java/org/zalando/logbook/QueryFilters.java
```diff
@@ -33,4 +32,10 @@ public static QueryFilter replaceQuery(final String name, final String replaceme
         return query -> pattern.matcher(query).replaceAll(replacementPattern);
     }
 
+    @API(status = MAINTAINED)
```
whiskeysierra: Experimental?

logbook-core/src/main/java/org/zalando/logbook/QueryFilters.java
nsmolenskii: there is some cases, when request contains a lot of query parameters, but you not really interested in them. 
in same time it could be few parameters you would like to keep. So it may make some sense to omit some parameters from logging. currently we using `replaceQuery` and it works quite good so far, until query names are small. unfortunately it not scaling with long query parameter names.

whiskeysierra: I see. That makes sense.

whiskeysierra:
Thanks for the contribution. I just released it as 1.11.0 :tada:

PR #379, opened 2018-11-01
Update bill of materials (BOM) to define version for each dependency

rainermueller:
## Description

To omit the versions in `pom.xml` the dependencies in `org.zalando:logbook-bom` must define `<version>${logbook.version}</version>`

## Motivation and Context

We still have to define the dependency version due to following error

`[ERROR] 'dependencies.dependency.version' for org.zalando:logbook-servlet:jar is missing. @ com.inventage.demo:backend-war:[unknown-version]
`
## Types of changes

## Checklist:

### DIFF
--- a/logbook-bom/pom.xml
+++ b/logbook-bom/pom.xml
@@ -20,30 +20,37 @@
             <dependency>
                 <groupId>org.zalando</groupId>
                 <artifactId>logbook-api</artifactId>
+                <version>1.12.0-SNAPSHOT</version>
             </dependency>
             <dependency>
                 <groupId>org.zalando</groupId>
                 <artifactId>logbook-core</artifactId>
+                <version>1.12.0-SNAPSHOT</version>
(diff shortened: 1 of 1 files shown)


logbook-bom/pom.xml
```diff
@@ -20,30 +20,37 @@
             <dependency>
                 <groupId>org.zalando</groupId>
                 <artifactId>logbook-api</artifactId>
+                <version>1.12.0-SNAPSHOT</version>
```
whiskeysierra: Is it possible to use a property pointing to it's own version? Otherwise we need to make sure to bump this here: 

https://github.com/zalando/logbook/blob/936deddc73ef6e82908a5db137480cbd6f952ee8/release.sh#L11

whiskeysierra:
I released your fix as 1.11.1 to central. Thanks a lot for the contribution! :tada:

Issue #423, opened 2019-02-09
Using JDK 11 causes NoClassDefFoundError


acet:
On the same code base using JDK 8 works as expected, but after switching to JDK 11 a NoClassDefFoundError is thrown 

```
Caused by: java.lang.NoClassDefFoundError: javax/activation/MimeType
	at org.zalando.logbook.servlet.RemoteRequest.isFormRequest(RemoteRequest.java:119)
	at org.zalando.logbook.servlet.RemoteRequest.withBody(RemoteRequest.java:100)
	at org.zalando.logbook.DefaultLogbook.write(DefaultLogbook.java:45)
	at org.zalando.logbook.servlet.NormalStrategy.logRequestIfNecessary(NormalStrategy.java:48)
	at org.zalando.logbook.servlet.NormalStrategy.doFilter(NormalStrategy.java:32)
	at org.zalando.logbook.servlet.LogbookFilter.doFilter(LogbookFilter.java:39)
	at org.zalando.logbook.servlet.HttpFilter.doFilter(HttpFilter.java:31)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
```

Please consider the application at: 
https://github.com/acet/logbook_demo_spring_starter
using both JDK 8 and JDK 11

## Possible Fix
Include 
```
        <dependency>
            <groupId>com.sun.activation</groupId>
            <artifactId>javax.activation</artifactId>
            <version>1.2.0</version>
        </dependency>
```
as a dependency in logback-servlet

## Context
I use JDK 11 and so I must currently include javax.activation in my dependencies currently to work around this

whiskeysierra:
Fix released as 1.13.0

PR #424, opened 2019-02-09
Add javax activation as a dependency to logbook-servlet

acet:
Potential fix for issue #423


whiskeysierra:
Thanks for your contribution! :tada: 

I released this as 1.13.0 to central just now.

PR #437, opened 2019-02-22
2.x Redesign

### LINKED ISSUE #211: Conflict with Spring Boot ErrorPageFilter
Hello ! First, thanks a lot for your awesome library !

I didn't see that this issue has already been reported even if some are close to what I encountered.

I get a conflict with ErrorPageFilter in non-embedded containers, the result is that the Spring Boot Whitelabel Error Page doesn't display. I have an error log similar to this one :
`2018-03-06 22:05:05.752 ERROR 8600 --- [nio-8080-exec-6] o.s.boot.web.support.ErrorPageFilter     : Cannot forward to error page for request [/test] as the response has already been committed. As a result, the response may have the wrong status code. If your application is running on WebSphere Application Server you may be able to resolve this problem by setting com.ibm.ws.webcontainer.invokeFlushAfterService to false`

The issue can easily be reproduced with a plain Spring Boot application like this [one](https://github.com/Anto89/logbook-error-filter.git). Deploy it to Tomcat (non-embedded) and hit a path that is unknown to the server.

### LINKED ISSUE #295: Allow non-text-based formats
## Detailed Description

Right now we strictly separate between formatting and writing, which is good. It allows us to combine formats (json, http, curl, ...) freely with different targets (log, stream, ...). What it doesn't really support well is non-text-based formats, e.g. logging requests/responses to a database.

## Context

#283, #229

## Possible Implementation

- to be defined

### LINKED ISSUE #296: Introduce strategy pattern
## Detailed Description

Right now the hard-coded default behavior consists of the following:
- log requests immediately
- log responses immediately
- log both, request and response, or neither of them

Being a bit more flexible in this regard would allow users to use different patterns/strategies, e.g.:
1. Log requests, but don't log responses
2. Log responses, but don't log requests
3. Only log requests if response status is not 2xx
4. Log requests without body, as long as response status is 2xx
5. Log some requests without body

## Context

#227, #231, #186, #183, #165, #336

## Your Environment

* Version used:
* Link to your project:

### LINKED ISSUE #315: OutOfMemoryError produced by huge response, that must be excluded.
Hello!
I catch java.lang.OutOfMemoryError: Java heap space when I download huge text file (70 Mb) and response body become very big. This request (and response correspondingly) was excluded in property file and doesn't appear in log.
JVM params are -Xmx256m -Xms128m
I tried to use logbook.write.chunk-size with values 10 and 10000 - it didn't work.
Here is the error log: https://pastebin.com/ESTUwqZW
How can I avoid this exception?

* Version used: 1.9.0

### LINKED ISSUE #334: SecurityStrategy logs twice
If I add a second LogBook Filter with Strategy.SECURITY, my calls resulting in 401 get logged twice. Once unfiltered with the whole body by the normal filter and once by the security filter.

I use Spring Boot without Spring Security.
What am I missing?

whiskeysierra:
## Motivation and Context

 
Fixes #211 
Fixes #295
Fixes #296 
Fixes #315 
Fixes #334

## Types of changes

## Checklist:


whiskeysierra:
Docs need to be updated.

logbook-core/src/main/java/org/zalando/logbook/CompositeSink.java
```diff
@@ -0,0 +1,34 @@
...
+    private final Collection<Sink> sinks;
+
+    @Override
+    public boolean isActive() {
+        return sinks.stream().anyMatch(Sink::isActive);
+    }
+
+    @Override
+    public void write(final Precorrelation precorrelation, final HttpRequest request) {
+        sinks.forEach(throwingConsumer(sink -> sink.write(precorrelation, request)));
```
lukasniemeier-zalando: Should these be `sinks.filter(Sink::isActive).forEach`?

logbook-api/src/main/java/org/zalando/logbook/Strategy.java
```diff
@@ -0,0 +1,30 @@
+package org.zalando.logbook;
+
+import org.apiguardian.api.API;
+
+import java.io.IOException;
+
+import static org.apiguardian.api.API.Status.STABLE;
+
+@API(status = STABLE)
+public interface Strategy {
```
lukasniemeier-zalando: TODO documentation? Just saying as the `README` teases this :wink:

README.md
```diff
@@ -121,18 +121,40 @@ or create a customized version using the `LogbookBuilder`:
...
+
+Starting with version 2.0 Logbook now comes with a [Strategy pattern](https://en.wikipedia.org/wiki/Strategy_pattern)
+at its core. Make sure you read the documentation of the [`Strategy`](logbook-api/src/main/java/org/zalando/logbook/Strategy.java)
+interface to understand the implications.
+
+Some example implementations can be found here:
+
+- [`BodyOnlyIfErrorStrategy`](logbook-core/src/test/java/org/zalando/logbook/BodyOnlyIfErrorStrategy.java)
+- [`ErrorResponseOnlyStrategy`](logbook-core/src/test/java/org/zalando/logbook/ErrorResponseOnlyStrategy.java)
+- [`WithoutBodyStrategy`](logbook-core/src/test/java/org/zalando/logbook/WithoutBodyStrategy.java)
```
lukasniemeier-zalando: What do you think of shipping such example strategies? We know that there is a demand for such behavior and we can easily foresee that users may want to apply this to multiple services. 

* `BodyOnlyIfStatusAtLeastStrategy(int)`
* `StatusAtLeastStrategy(int)`
* `WithoutBodyStrategy`

README.md
```diff
@@ -121,18 +121,40 @@ or create a customized version using the `LogbookBuilder`:
...
+
+- Requests/responses are logged separately
+- Requests/responses are logged soon as possible
+- Requests/responses are logged as a pair or not logged at all  
+  (i.e. no partial logging of traffic)
+
+Some of those restrictions could be mitigated with custom [`HttpLogWriter`](#writing)
+implementations, but they were never ideal.
+
+Starting with version 2.0 Logbook now comes with a [Strategy pattern](https://en.wikipedia.org/wiki/Strategy_pattern)
```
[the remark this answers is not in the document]
whiskeysierra: This one in itself doesn't really require a migration, since it's just a new feature. But it's definitely worth having proper release notes for 2.0.

whiskeysierra:
> I see a few `TODO`s on "body replacement/without-body" - are they required for this PR? Or long-term?

I don't have an opinion how we should behave there.

logbook-api/src/main/java/org/zalando/logbook/Strategy.java
whiskeysierra: :man_facepalming:

README.md
whiskeysierra: Yeah, probably worth it. I was hoping that the interface is so easy to use that users can just implement it within 2-5 lines.

logbook-core/src/main/java/org/zalando/logbook/CompositeSink.java
whiskeysierra: Yes... :man_facepalming: :

whiskeysierra: Fixed

logbook-api/src/main/java/org/zalando/logbook/Strategy.java
whiskeysierra: Fixed

README.md
lukasniemeier-zalando: Agreed on the easy-to-be-used, but still gets tedious to put these 5 lines in my 25 micro-services :wink:

whiskeysierra: I created #440 as a follow-up.

lukasniemeier-zalando:
:+1:

Issue #381, opened 2018-11-05
Extract client identifier from authorization token


whiskeysierra:
## Detailed Description

Logbook should enrich requests with the subject from the JWT token in the `Authorization` header, if present.

## Context

Having the id as part of the requests would make it way easier to identify clients which in turn helps when:
- identifying unauthorized access issues
- usage analysis

## Possible Implementation

- introduce the concept of an *attribute*
- attributes are simple key-value pairs (**tbd** type of value)
- a request/response can have multiple attributes
- attributes should be derived/created from requests/responses before any filtering (e.g. obfuscation)
- built-in attribute extractor for `sub` from JWT token
  - detect JWT tokens: Bearer prefix + 3x base64 data separated by dots
  - remove `Bearer` prefix
  - split at `.`
  - base64 decode payload, i.e. the second element
  - parse JSON
  - read properties in order and return the first one that is present
    - `https://identity.zalando.com/managed-id` (Zalando employee tokens)
    - `sub`
  - don't hard code priorities, but rather allow to configure a list of names, defaults to `["sub"]`
- extend `JsonHttpLogFormatter` to include attributes (**tbd**, top level? nested?  name clashes?)

### Employee Token

```json
{
  "sub": "3b66d47c-d886-4c63-a0b9-9ec3cad7e848",
  "https://identity.zalando.com/realm": "users",
  "https://identity.zalando.com/token": "Bearer",
  "https://identity.zalando.com/managed-id": "wschoenborn",
  "azp": "ztoken",
  "https://identity.zalando.com/bp": "810d1d00-4312-43e5-bd31-d8373fdd24c7",
  "auth_time": 1540188140,
  "iss": "https://identity.zalando.com",
  "exp": 1541411248,
  "iat": 1541407638
}
```

### Service Token

```json
{
  "sub": "stups_sales-order-service",
  "https://identity.zalando.com/realm": "services",
  "https://identity.zalando.com/token": "Bearer",
  "azp": "stups_sales-order-service_389e4e16-0695-45df-9afd-d9be0ffab456",
  "https://identity.zalando.com/bp": "810d1d00-4312-43e5-bd31-d8373fdd24c7",
  "iss": "https://identity.zalando.com",
  "exp": 1541411315,
  "iat": 1541407705,
  "https://identity.zalando.com/privileges": [
    "com.zalando::loyalty_point_account.read_all"
  ]
}
```

### Links

- https://jwt.io/
- https://tools.ietf.org/html/rfc7519

[...]

PR #449, opened 2019-03-02
Chunking is now limited to the body itself

### LINKED ISSUE #322: Chunking should only affect the body
## Your Environment

* Version used:
* Link to your project:

whiskeysierra:
## Description

Chunking used to work on the raw string that was the result of the formatting. This fix limits the impact of chunking to the body itself.

## Motivation and Context

Fixes #322

## Types of changes

## Checklist:


logbook-core/src/main/java/org/zalando/logbook/ChunkingSink.java
```diff
@@ -0,0 +1,54 @@
...
+        chunk(original)
+                .map(chunk -> new BodyReplacementHttpRequest(original, chunk))
+                .forEach(throwingConsumer(replaced -> delegate.write(precorrelation, replaced)));
+    }
+
+    @Override
+    public void write(final Correlation correlation, final HttpRequest request, final HttpResponse original) throws IOException {
+        chunk(original)
+                .map(chunk -> new BodyReplacementHttpResponse(original, chunk))
+                .forEach(throwingConsumer(replaced -> delegate.write(correlation, request, replaced)));
```
lukasniemeier-zalando: What about going on step further and hide the headers, like

```
final Iterator<String> chunks = chunk(original).iterator();
final String firstChunk = chunks.next();
delegate.write(correlation, request, new BodyReplacementHttpResponse(original, firstChunk));
chunks.forEachRemaining(chunk -> delegate.write(correlation, request, new BodyReplacementHttpResponse(new HeaderHidingHttpResponse(original), chunk)));
```

whiskeysierra: That would move it a bit closer to be a properly usable feature. Until now it was basically breaking our format, but even with my change it's still a gimmick and not 100% *"production-ready"*.

As an example a parser wouldn't know that this is a chunk. I believe it would be nice to get attribute-support in (see #381) and use that to *tag* individual chunks with their number. Parsers like Scalyr could then exclude anything with `$chunk > 1` when counting requests/responses (as an example).

lukasniemeier-zalando:
:+1:

PR #458, opened 2019-03-08
Initial support for logstash-logback-encoder

skjolber:
Add support for logstash-logback-encoder. 

## Description
Add a new submodule that insert the request/response-logging as a subtree node, so that the structure is something like

```
{
  "@timestamp" : "2019-03-08T09:37:46.239+01:00",
  "@version" : "1",
  "message" : "GET http://localhost/test?limit=1",
  "logger_name" : "org.zalando.logbook.Logbook",
  "thread_name" : "main",
  "level" : "TRACE",
  "level_value" : 5000,
  "http" : {
    "origin" : "remote",
    "type" : "request",
    "correlation" : "3ce91230-677b-11e5-87b7-10ddb1ee7671",
    "protocol" : "HTTP/1.0",
    "remote" : "127.0.0.1",
    "method" : "GET",
    "uri" : "http://localhost/test?limit=1",
    "headers" : {
      "Accept" : [ "application/json" ],
      "Date" : [ "Tue, 15 Nov 1994 08:12:31 GMT" ]
    },
    "body" : {
      "person" : {
        "name" : "Thomas"
      }
    }
  }
}
```

## Motivation and Context
logstash-logback-encoder is a library for structured logging, which is considered best practice. We have been using Logbook in production with logstash-logback-encoder for some time. We're using single log files per application in GCP stackdriver.

## Types of changes

## Checklist:


skjolber:
Implementation note: I would prefer to also add the HTTP code name (i.e. 'OK' in '200 OK') in the LogbackLogstashSink.responseMessage(..) method, but not sure how to get to it. Quite a few libraries have mappers, like Spring's HttpCode, but those are not in the dependencies as far as I can see.

whiskeysierra:
> I would prefer to also add the HTTP code name (i.e. 'OK' in '200 OK') in the LogbackLogstashSink.responseMessage(..) method, but not sure how to get to it. Quite a few libraries have mappers, like Spring's HttpCode, but those are not in the dependencies as far as I can see.

We had the same problem already for the `DefaultHttpLogFormatter` and decided to keep a simple mapping in there:

https://github.com/zalando/logbook/blob/9eca1eb0c9527b330ccbe03ee4c05f78e09ae667/logbook-core/src/main/java/org/zalando/logbook/DefaultHttpLogFormatter.java#L21-L39

Not sure where to put it, but we may find a place where both modules can share it.

logbook-logstash-logback-encoder/src/test/java/org/zalando/logbook/lle/DefaultSinkTest.java
```diff
@@ -0,0 +1,77 @@
...
+import java.io.IOException;
+
+import org.junit.jupiter.api.Test;
+import org.zalando.logbook.Correlation;
+import org.zalando.logbook.HttpRequest;
+import org.zalando.logbook.HttpResponse;
+import org.zalando.logbook.Precorrelation;
+import org.zalando.logbook.StructuredHttpLogFormatter;
+
+class DefaultSinkTest {
```
whiskeysierra: `LogbackLogstashSinkTest`?

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/LogbackLogstashLogWriter.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.io.IOException;
+
+import org.apiguardian.api.API;
+import org.slf4j.Marker;
+import org.zalando.logbook.Correlation;
+import org.zalando.logbook.Precorrelation;
+
+@API(status = EXPERIMENTAL)
+public interface LogbackLogstashLogWriter {
```
whiskeysierra: Do you expect other implementations? I'd actually get rid of this an inline the default implementation into your sink.

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/AutodetectPrettyPrintingMarker.java
```diff
@@ -0,0 +1,45 @@
...
+
+import net.logstash.logback.marker.RawJsonAppendingMarker;
+
+/**
+ * 
+ * Auto-detecting pretty-printing Marker. Pretty-printing cannot be performed in advance.
+ *
+ */
+
+public class AutodetectPrettyPrintingMarker extends RawJsonAppendingMarker {
```
whiskeysierra: Package private? Final?

logbook-logstash-logback-encoder/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+            <groupId>org.zalando</groupId>
+            <artifactId>logbook-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.zalando</groupId>
+            <artifactId>logbook-json</artifactId>
+        </dependency>        
+        <dependency>
+            <groupId>net.logstash.logback</groupId>
+            <artifactId>logstash-logback-encoder</artifactId>
```
whiskeysierra: The *magic* of this guy only works with logback directly? Any chance we can build our module just using slf4j?

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/AutodetectPrettyPrintingMarker.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    protected void writeFieldValue(JsonGenerator generator) throws IOException {
+        PrettyPrinter prettyPrinter = generator.getPrettyPrinter();
+        if(prettyPrinter == null) {
+            super.writeFieldValue(generator);
+        } else {
+            // append to existing tree event by event
+            final JsonParser parser = generator.getCodec().getFactory().createParser((String)super.getFieldValue());
```
whiskeysierra: The logbook-json module contains a pretty printing body filter. If that one is used in conjunction you can ignore it here and just use `generator.writeRaw(..)`.

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/LogbackLogstashLogWriter.java
whiskeysierra: Or alternatively apply the marker within your custom writer but without the need to change the interface, i.e. stick with `HttpLogFormatter`.

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/LogbackLogstashSink.java
```diff
@@ -0,0 +1,49 @@
...
+    private final LogbackLogstashLogWriter logWriter;
+
+    @Override
+    public boolean isActive() {
+        return logWriter.isActive();
+    }
+
+    @Override
+    public void write(final Precorrelation precorrelation, final HttpRequest request) throws IOException {
+        Marker marker = new AutodetectPrettyPrintingMarker(request.getScheme(), formatter.format(precorrelation, request));
```
whiskeysierra: I'd just hardcode `http` as the field name. Wouldn't it be confusing anyway that sometimes this field is called `http` and sometimes `https`?

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/LogbackLogstashLogWriter.java
whiskeysierra: The alternative will probably not work since you need access to the request/response to construct your message. But your use case i pretty much what Sink was designed for.

logbook-logstash-logback-encoder/src/test/java/org/zalando/logbook/lle/LogbackLogstashLogWriterTest.java
```diff
@@ -0,0 +1,30 @@
...
+import org.junit.jupiter.api.Test;
+import org.slf4j.Marker;
+import org.zalando.logbook.Correlation;
+import org.zalando.logbook.Precorrelation;
+
+public class LogbackLogstashLogWriterTest {
+
+    @Test
+    public void testDefaultImplementation() {
+        LogbackLogstashLogWriter writer = new LogbackLogstashLogWriter() {
```
whiskeysierra: You can actually create a mock and use `callRealMethod` as the default answer. Saves you ~20 lines.

logbook-logstash-logback-encoder/src/test/java/org/zalando/logbook/lle/LogbackLogstashTest.java
```diff
@@ -0,0 +1,143 @@
...
+                .assertThat("$.http.protocol", is("HTTP/1.0"))
+                .assertThat("$.http.status", is(200))
+                .assertThat("$.http.headers.*", hasSize(1))
+                .assertThat("$.http.headers['Date']", is(singletonList("Tue, 15 Nov 1994 08:12:31 GMT")))
+                .assertThat("$.http.body.person.name", is("Magnus"))
+                .assertThat("$.http.duration", is(125));
+        }        
+    }
+
+    static class SimplePrecorrelation implements Precorrelation {
```
whiskeysierra: You could just mock those.

logbook-logstash-logback-encoder/src/test/java/org/zalando/logbook/lle/PrettyPrintingStaticAppender.java
```diff
@@ -0,0 +1,40 @@
...
+    public static void reset() {
+        bout.reset();
+    }
+    
+    public static String getLastStatement() {
+        try {
+            String content = bout.toString(StandardCharsets.UTF_8.name());
+            
+            return content.substring(content.lastIndexOf("\n{\n", content.length()) + 1);
+        } catch (UnsupportedEncodingException e) {
```
whiskeysierra: If you use `String content = new String(bout.toByteArray(), UTF_8)` you wouldn't need to catch anything here.

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/AutodetectPrettyPrintingMarker.java
skjolber: Can be final. Must not be private, as it is referred to in configuration. However this is more of a convenience class than anything.

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/AutodetectPrettyPrintingMarker.java
skjolber: That will not work out of the box, because the indent level will one level off. But I'll look into it.

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/LogbackLogstashSink.java
skjolber: Agreed

logbook-logstash-logback-encoder/pom.xml
skjolber: Yes, Logback is the only alternative. From https://github.com/logstash/logstash-logback-encoder: 

`<!-- Your project must also directly depend on either logback-classic or logback-access. -->`

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/LogbackLogstashLogWriter.java
skjolber: Remved interface

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/AutodetectPrettyPrintingMarker.java
skjolber: Actually I think the marker approach is more tidy as it uses the log framework's prettyprinting. There might be a case for improved performance with pretty-printing here, but thats not for production use anyways, so it is not that important (at least not to us). 

I'm not claiming that very many are doing this, but it also keeps the door open for stuff like this:

https://github.com/skjolber/logback-logstash-syntax-highlighting-decorators

logbook-logstash-logback-encoder/src/test/java/org/zalando/logbook/lle/LogbackLogstashTest.java
skjolber: Mocked (originally copied from JsonHttpLogFormatterTest)

logbook-logstash-logback-encoder/pom.xml
skjolber: The Marker is a slf4j interface, I'm not really sure how it is handled in various logging other frameworks. I guess that can be investigated later if support for more JSON logging frameworks is added (?).

logbook-logstash-logback-encoder/src/test/java/org/zalando/logbook/lle/DefaultSinkTest.java
skjolber: Renamed

skjolber:
So the DefaultHttpLogFormatter HTTP codes - is not that enum material for org.zalando.logbook.HttpResponse.getStatus()? Or perhaps (non-breaking change) a corresponding getMessage() / getStatusMessage().

whiskeysierra:
We only want that for the *natural* `http` style since it's also on the wire, but for JSON I found that it's not needed, especially if we just map it from status code anyway.

skjolber:
@whiskeysierra so where can/should I move the HTTP code lookup?

whiskeysierra:
Can I convince you that the status code is enough?

whiskeysierra:
@skjolber Tests are failing with the Spring 4 profile (`-P spring4`)

skjolber:
@whiskeysierra so now it builds. Still would want the HTTP status code message; have yet to meet a dev whom wants to be ops, so would like to make it as painless as possible. Many of the juniors probably dont know the codes by hearth. Alternatives:

Request: 

1. GET /path
2. Request GET /path

 Response:

1. GET /path 200 OK
2. GET /path 200 
3. 200 OK GET /path
4. 200 GET /path
5. GET /path Response 200 OK
6. GET /path Response 200 
7. Response 200 OK GET /path
8. Response 200 GET /path

So I wonder if the most 'readable' ones are 1 and 3.

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/AutodetectPrettyPrintingMarker.java
whiskeysierra: Let's make it `final` then and mark it with `@API(status=INTERNAL)`?

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/AutodetectPrettyPrintingMarker.java
whiskeysierra: Agreed.

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/AutodetectPrettyPrintingMarker.java
```diff
@@ -0,0 +1,46 @@
...
+
+            try {
+                while (parser.nextToken() != null) {
+                    generator.copyCurrentEvent(parser);
+                }
+            } finally {
+                parser.close();
```
whiskeysierra: What about using a *try-with-resource* statement here?

logbook-logstash-logback-encoder/src/test/java/org/zalando/logbook/lle/LogbackLogstashTest.java
whiskeysierra: Maybe worth doing it there too?

whiskeysierra:
> Many of the juniors probably dont know the codes by hearth

Although many of the reason phrases are not really helping much, tbh.

> So I wonder if the most 'readable' ones are 1 and 3.

I'm inclined to agree.

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/LogstashLogbackHttpLogWriter.java
```diff
@@ -0,0 +1,29 @@
...
+import java.io.IOException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.Marker;
+import org.zalando.logbook.Correlation;
+import org.zalando.logbook.Logbook;
+import org.zalando.logbook.Precorrelation;
+
+public class LogstashLogbackHttpLogWriter {
```
whiskeysierra: Package private and final?

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/LogstashLogbackSink.java
```diff
@@ -0,0 +1,49 @@
...
+import org.zalando.logbook.Precorrelation;
+import org.zalando.logbook.Sink;
+
+import lombok.AllArgsConstructor;
+
+@AllArgsConstructor
+public final class LogstashLogbackSink implements Sink {
+
+    private final HttpLogFormatter formatter;
+    private final LogstashLogbackHttpLogWriter logWriter;
```
whiskeysierra: I'd actually inline the logic form this one and drop the `LogstashLogbackHttpLogWriter`.

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/AutodetectPrettyPrintingMarker.java
skjolber: Yes, but was there not a Jacoco issue with those - has it been fixed? See JsonCompactor.

README.md
```diff
@@ -619,6 +624,39 @@ logbook:
...
+```
+appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+   <encoder class="net.logstash.logback.encoder.LogstashEncoder"/>
+/appender>
+```
+
+configure Logbook with a `LogstashLogbackSink`
+
+```
+LogstashLogbackHttpLogWriter logWriter = new LogstashLogbackHttpLogWriter();
```
whiskeysierra: Needs to be updated as well I guess.

logbook-logstash-logback-encoder/src/main/java/org/zalando/logbook/lle/AutodetectPrettyPrintingMarker.java
```diff
@@ -0,0 +1,46 @@
+package org.zalando.logbook.lle;
```
whiskeysierra: Any idea for a more speaking package name?

skjolber: org.zalando.logbook.logback.logstash or org.zalando.logbook.logstash??

whiskeysierra: > org.zalando.logbook.logstash

I'd go with that. Module name (directory/artifactId) would then be `logbook-logstash`?

skjolber: Yeah, lets go with that.

skjolber:
So for the HTTP reason phrases, as in DefaultHttpLogFormatter, can I move it as a default implementation of getMessage() in HttpResponse?

whiskeysierra:
I'd call it `getReasonPhrase()` but in general I'd be OK with that.

skjolber:
Went with a switch statement instead of HashMap, interfaces cannot have static initializers, should be at least as fast. Noticed you supported one of the unofficial nginx http codes, so added the others as well.

README.md
```diff
@@ -105,6 +106,10 @@ Alternatively, you can import our *bill of materials*...
     <groupId>org.zalando</groupId>
     <artifactId>logbook-spring-boot-starter</artifactId>
 </dependency>
+<dependency>
+    <groupId>org.zalando</groupId>
+    <artifactId>logbook-logstash-logback-encoder</artifactId>
```
whiskeysierra: logbook-logstash

logbook-api/src/test/java/org/zalando/logbook/HttpResponseTest.java
```diff
@@ -0,0 +1,72 @@
...
+        assertThat(response.getReasonPhrase(), is("Bad Request"));
+        
+    }
+    
+    @Test
+    public void testEmptyCodes() {
+        HttpResponse response = mock(HttpResponse.class);
+        when(response.getReasonPhrase()).thenCallRealMethod();
+        
+        List<Integer> list = Arrays.stream(codes).boxed().collect(Collectors.toList());
```
whiskeysierra: `Arrays.stream(codes).forEach(code -> ..)`

logbook-logstash/src/main/java/org/zalando/logbook/logstash/LogstashLogbackSink.java
```diff
@@ -0,0 +1,65 @@
...
+import org.zalando.logbook.HttpRequest;
+import org.zalando.logbook.HttpResponse;
+import org.zalando.logbook.Logbook;
+import org.zalando.logbook.Precorrelation;
+import org.zalando.logbook.Sink;
+
+import lombok.AllArgsConstructor;
+
+@AllArgsConstructor
+public final class LogstashLogbackSink implements Sink {
```
whiskeysierra: `@API(status=EXPERIMENTAL)`

PR #463, opened 2019-03-14
Fixed replacement of escaped quotes in string properties

### LINKED ISSUE #462: replaceJsonStringProperty BodyFilter not working for values with quotation marks
## Description
I do the following:

> BodyFilter bodyFilter = BodyFilters.replaceJsonStringProperty(Collections.singleton("password"), "\<secret\>");

With this config, when I have a password which contains a quotation mark in it, e.g.: 

> abc"!?$123

 and I try to hide it, the filter will fail to do so, at least partially. So, when I do the following:

> bodyFilter.filter("application/json", "{\\"password\\": \\"abc\\\\\\"!?$123\\"}");

the output is the following:

> {"password": "\<secret\>"!?$123"}

So, most of the password is still visible - the whole part after the quotation mark. I'd expect the whole value to be hidden, i.e.:

>  {"password": "\<secret\>"}

## Expected Behavior

I'd expect the whole value to be hidden, i.e.:

>  {"password": "\<secret\>"}

## Actual Behavior

The output is the following:

> {"password": "\<secret\>"!?$123"}

## Possible Fix

The regex used for finding the property looks like this:

> final String regex = properties.stream().map(Pattern::quote).collect(joining("|"));
final Pattern pattern = Pattern.compile("(\"(?:" + regex + ")\"\\s*:\\s*)\".*?\"");

I think it might have to be improved somehow.

## Steps to Reproduce

Use the following code:
> BodyFilter bodyFilter = BodyFilters.replaceJsonStringProperty(Collections.singleton("password"), "\<secret\>");
bodyFilter.filter("application/json", "{\\"password\\": \\"abc\\\\\\"!?$123\\"}");

## Context

I wanted to hide sensitive data in logs which might contain special characters, like quotation marks.

## Your Environment

* Version used: logbook-core-1.13.0

whiskeysierra:
## Motivation and Context

Fixes #462

## Types of changes

## Checklist:

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/BodyFilters.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/BodyFilters.java
@@ -69,7 +69,7 @@ public final class BodyFilters {
         final Predicate<String> json = MediaTypeQuery.compile("application/json", "application/*+json");
 
         final String regex = properties.stream().map(Pattern::quote).collect(joining("|"));
-        final Pattern pattern = Pattern.compile("(\"(?:" + regex + ")\"\\s*:\\s*)\".*?\"");
+        final Pattern pattern = Pattern.compile("(\"(?:" + regex + ")\"\\s*:\\s*)\".*?(?<!\\\\)\"");
         final UnaryOperator<String> delegate = body -> pattern.matcher(body).replaceAll("$1\"" + replacement + "\"");
 
         return (contentType, body) -> json.test(contentType) ? delegate.apply(body) : body;
--- a/logbook-core/src/test/java/org/zalando/logbook/BodyFiltersTest.java
+++ b/logbook-core/src/test/java/org/zalando/logbook/BodyFiltersTest.java
@@ -2,8 +2,7 @@ package org.zalando.logbook;
 
 import org.junit.jupiter.api.Test;
 
-import java.util.Collections;
-
+import static java.util.Collections.singleton;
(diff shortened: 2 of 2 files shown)


whiskeysierra:
@lukasniemeier-zalando Any idea how harmful this could be in terms of performance?

logbook-core/src/main/java/org/zalando/logbook/BodyFilters.java
```diff
@@ -69,7 +69,7 @@ public static BodyFilter replaceJsonStringProperty(final Set<String> properties,
         final Predicate<String> json = MediaTypeQuery.compile("application/json", "application/*+json");
 
         final String regex = properties.stream().map(Pattern::quote).collect(joining("|"));
-        final Pattern pattern = Pattern.compile("(\"(?:" + regex + ")\"\\s*:\\s*)\".*?\"");
+        final Pattern pattern = Pattern.compile("(\"(?:" + regex + ")\"\\s*:\\s*)\".*?(?<!\\\\)\"");
```
lukasniemeier-zalando:
```
"password": "abc\\","foo":"!?$123"
```
Hmmm :see_no_evil:

lukasniemeier-zalando: We may want to offer an alternative `ParsingJsonBodyFilter` which allows filtering for field names (by regex? by path? by predicate?). The default `replaceJsonStringProperty` should then document its pitfalls.

PR #490, opened 2019-04-24
Updated docs

### LINKED ISSUE #488: Spring-Boot: Missing logs on HTTP 500
When a controller throws an Exception, not all is logged as expected.

## Description

When a spring boot controller throws an Exception, resulting in a HTTP 500 which returns (in case of a RestController) details about the error, Logbook logging is not what I would expect.

I'm talking about cases where DispatcherServlet is logging the error like this:

    Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed

## Expected Behavior

Basic configuration: log request and response

With status strategy as below: log request and response

    logbook:
      strategy: status-at-least

## Actual Behavior

Basic configuration: only request gets logged

With status strategy as above: neither request nor response are logged

## Steps to Reproduce
1. Spring-boot app with controller throwing an Exception
2. use default configuration for logbook, except for the case with the strategy, then change as above
3. curl the URL

## Context

I want to log internal server errors including the request and response data.

## Your Environment
* Version used: 2.0.0-RC.0
* Spring-Boot 2.0.8

### LINKED ISSUE #489: Documentation for setting up body filter with Spring Boot Auto-configuration
The information of #298 should be part of the README, including necessary imports. It's possible to find it in closed issues, but that's not ideal.

A link to [LogbookAutoConfiguration](https://github.com/zalando/logbook/blob/master/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java) is also helpful to see what's possible outside `application.yml`.

whiskeysierra:
Fixes #488
Fixes #489

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -444,6 +444,9 @@ context.addFilter("LogbookFilter", new LogbookFilter(logbook))
     .addMappingForUrlPatterns(EnumSet.of(REQUEST, ASYNC), true, "/*"); 
 ```
 
+**Beware**: The `ERROR` dispatch is not supported. You're strongly advised to produce error responses within the
+`REQUEST` or `ASNYC` dispatch.
+
 The `LogbookFilter` will, by default, treat requests with a `application/x-www-form-urlencoded` body not different from
 any other request, i.e you will see the request body in the logs. The downside of this approach is that you won't be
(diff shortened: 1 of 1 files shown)


lukasniemeier-zalando:
:+1:

PR #508, opened 2019-05-16
Adjust logging of empty or filtered body

skjolber:
Invalid body values are logged as raw valid JSON values. 

## Description
For the special case of an empty body, resulting from using a filter, log an empty body.

## Motivation and Context
Produces invalid log statements.

## Types of changes

## Checklist:

### DIFF
--- a/logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
@@ -62,11 +62,14 @@ public final class JsonHttpLogFormatter implements StructuredHttpLogFormatter {
     public Optional<Object> prepareBody(final HttpMessage message) throws IOException {
         final String contentType = message.getContentType();
         final String body = message.getBodyAsString();
-
+        if(body.isEmpty()) {
+            return Optional.empty();
+        }
         if (JsonMediaType.JSON.test(contentType)) {
+        	// TODO has this JSON been validated? If not then this might result in invalid log statements
             return Optional.of(new JsonBody(body));
         } else {
-            return Optional.ofNullable(body.isEmpty() ? null : body);
+            return Optional.of(body);
         }
     }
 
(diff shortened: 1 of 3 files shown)


logbook-logstash/src/test/java/org/zalando/logbook/logstash/LogbackLogstashSinkTest.java
```diff
@@ -138,4 +138,64 @@ void shouldLogRequestAndResponse() throws IOException {
         
     }
 
+
+    @Test
+    void shouldLogCorrectJsonForEmptyBody() throws IOException {
```
skjolber: Perhaps this test is not necessary, as the fix is really in JsonHttpLogFormatter.

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
```diff
@@ -62,11 +62,14 @@ public JsonHttpLogFormatter(final ObjectMapper mapper) {
     public Optional<Object> prepareBody(final HttpMessage message) throws IOException {
         final String contentType = message.getContentType();
         final String body = message.getBodyAsString();
-
+        if(body.isEmpty()) {
+            return Optional.of("");
```
whiskeysierra: Now we have three return statements in here and all of them use `Optional.of(..)`. That means we never return an empty Optional now.

The original idea was to return empty for an empty/absent body. The contract of the JSON log output until also was that a request/response without body will produce an absent `body` field in the log message. I'd like to preserve that.

skjolber: Ok, I'd like that too, just make a minimal fix first. So then HttpMessage can return null in getBodyAsString() or what?

whiskeysierra: I believe what we should do is replace line 66 with `return Optional.empty()`.

whiskeysierra:
:+1:

PR #511, opened 2019-05-20
JMH module

skjolber:
Basic JHM module for performance testing

## Description
New module with a few baseline tests and documentation.

## Motivation and Context
We're using this library for logging in our microservices and would like to keep the resource usage and response time low.

## Types of changes

## Checklist:

I added a seperate README for the module. Perhaps this checklist should be updated with an item for how a PR affects the performance.

### DIFF
--- a/logbook-jmh/README.md
+++ b/logbook-jmh/README.md
@@ -1,0 +1,42 @@
+# logbook-jmh
+Module for performance-testing using the [JMH](https://openjdk.java.net/projects/code-tools/jmh/) framework.
+
+## Introduction
+Performance analysis is complicated due to the dynamic nature of the JVM Hotspot implementation. 
+
+For example the JVM Hotspot is affected by the classes it loads and better optimize when there is only one implementation of an interface in use. 
+
+In practice much of the dynamic nature can be handled by using so-called forks and warmups:
+
+ * forks makes sure the individual benchmarks run alone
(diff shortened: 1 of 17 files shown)


skjolber:
Waiting for https://github.com/zalando/logbook/pull/515

whiskeysierra:
> Waiting for #515

Done

skjolber:
@whiskeysierra ready for first round of review

logbook-jmh/README.md
```diff
@@ -0,0 +1,42 @@
+# logbook-jmh
```
whiskeysierra: Would the same setup allow us to compare a simple web app with and without logbook to compare IO throughput/latency?

skjolber: I'm not sure, I've used other tools, like Gatling, for that; it is just not so 'micro' anymore. Ideally you'll want the test driver and the system under test on two systems to be completely fair, the driver will consume considerable resources. 

But I am guessing this can be done, at least for determining the introduced delay when not in saturation.

skjolber: I think I'm going to go with 'no'. What you'd like is a reference point for the current implementation, and then interpolate as performance changes. You'll probably see with and without logging is something like 10-25% worth (loosely based on [this](https://www.greenbird.com/2015/06/11/logging-soap-traffic-in-mule-esb-versions-3-2-to-3-6)).

whiskeysierra:
:+1:

PR #512, opened 2019-05-20
Jackson-based body filter for JSON

skjolber:
New JSON body filter.

## Description
Add filter which uses Jackson to parse/serialize when filtering JSON. Supports both scalar values (i.e. string, number, boolean etc) and structures (tree, object).

## Motivation and Context
Remove sensitive information (GDPR) and also cut down on log statement size (we're using stackdriver and there is a limit on statement size).

## Types of changes

## Checklist:

### DIFF
--- a/logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
@@ -1,7 +1,6 @@
 package org.zalando.logbook.json;
 
 import java.io.StringWriter;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.Set;
@@ -49,7 +48,7 @@ public class JacksonJsonFieldBodyFilter implements BodyFilter {
             
             JsonGenerator generator = factory.createGenerator(writer);            
             try {
-                do {
+                while(true) {
                     JsonToken nextToken = parser.nextToken();
                     if(nextToken == null) {
                         break;
@@ -63,8 +62,7 @@ public class JacksonJsonFieldBodyFilter implements BodyFilter {
                             parser.skipChildren(); // skip children
                         }
                     }
-                    
-                } while(true);
+                }                    
             } finally {
                 parser.close();
                 
--- a/logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
@@ -1,0 +1,82 @@
+package org.zalando.logbook.json;
+
+import java.io.StringWriter;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Set;
+import org.zalando.logbook.BodyFilter;
+
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * 
+ * Thread-safe filter for JSON fields. Filters on property names, 
+ *
+ */
+
+public class JacksonJsonFieldBodyFilter implements BodyFilter {
+
+    private final String replacement;
+    private final Set<String> fields;
+    private final ObjectMapper objectMapper;
+    
+    public JacksonJsonFieldBodyFilter(Collection<String> fieldNames, String replacement, ObjectMapper objectMapper) {
+        this.fields = new HashSet<>(fieldNames); // thread safe for reading
+        this.replacement = replacement;
+        this.objectMapper = objectMapper;
+    }
+
+    public JacksonJsonFieldBodyFilter(Collection<String> fieldNames, String replacement) {
+        this(fieldNames, replacement, new ObjectMapper());
+    }
+
+    @Override
+    public String filter(String contentType, String body) {
+        return JsonMediaType.JSON.test(contentType) ? filter(body) : body;
+    }
+
+    public String filter(final String body) {
+        try {
+            JsonFactory factory = objectMapper.getFactory();
+            final JsonParser parser = factory.createParser(body);
+            
+            StringWriter writer = new StringWriter(body.length() * 2); // rough estimate of final size
+            
+            JsonGenerator generator = factory.createGenerator(writer);            
+            try {
+                do {
+                    JsonToken nextToken = parser.nextToken();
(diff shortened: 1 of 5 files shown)


logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
```diff
@@ -0,0 +1,91 @@
...
+    public String filter(final String body) {
+        try {
+            JsonFactory factory = objectMapper.getFactory();
+            final JsonParser parser = factory.createParser(body);
+            
+            StringWriter writer = new StringWriter(body.length() * 2); // rough estimate of final size
+            
+            JsonGenerator generator = factory.createGenerator(writer);            
+            try {
+                do {
```
whiskeysierra: Why not `while` instead of `do ... while`?

logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
```diff
@@ -0,0 +1,91 @@
...
+                    JsonToken nextToken = parser.nextToken();
+                    if(nextToken == null) {
+                        break;
+                    }
+
+                    generator.copyCurrentEvent(parser);
+                    if(nextToken == JsonToken.FIELD_NAME && fields.contains(parser.getCurrentName())) {
+                        nextToken = parser.nextToken();
+                        if(nextToken.isScalarValue()) {
```
whiskeysierra:
```java
generator.writeString("XXX");

if (!nextToken.isScalarValue()) {
    parser.skipChildren();
}
```

logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
```diff
@@ -0,0 +1,91 @@
...
+                            generator.writeString("XXX");
```
whiskeysierra: Dynamic replacement?

logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * 
+ * Thread-safe filter for JSON fields. Filters on property names, 
+ *
+ */
+
+public class JacksonJsonFieldBodyFilter implements BodyFilter {
+
+    public static JacksonJsonFieldBodyFilter newInstance(String ... fieldNames) {
```
whiskeysierra: Constructor is enough, imo.

logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
```diff
@@ -0,0 +1,80 @@
...
+                }                    
+            } finally {
+                parser.close();
+                
+                generator.close();
+            }
+            
+            return writer.toString();
+        } catch(Exception e) {
+            // ignore
```
whiskeysierra: Should we log something?

https://github.com/zalando/logbook/blob/f6321b4fcf3a726bb0135c6a1945350dcad9db2d/logbook-json/src/main/java/org/zalando/logbook/json/CompactingJsonBodyFilter.java#L40

skjolber: If this code defers logging to upstream, thats perhaps better? If the document is invalid, surely that will be caught during databinding.

whiskeysierra: What kind of exception are we potentially ignoring here right now?

skjolber: Invalid JSON (or content classified as JSON but in another format somehow). Should not happen very often.

logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
```diff
@@ -0,0 +1,80 @@
...
+
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * 
+ * Thread-safe filter for JSON fields. Filters on property names,
```
whiskeysierra: Can we document that the side effect of this filter is compacting of the JSON body and therefore shouldn't be used in conjunction with the compacting body filter.

logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
whiskeysierra: Still, if it does seeing a log message would be nice.

logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
skjolber: Ok, I guess then adding compacting logic on the exception handling makes sense.

skjolber: So I'll wait for https://github.com/zalando/logbook/pull/528/files

skjolber: Added compacting filter on exception

logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
skjolber: Added log message.

logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
```diff
@@ -0,0 +1,87 @@
...
+                }                    
+            } finally {
+                parser.close();
+                
+                generator.close();
+            }
+            
+            return writer.toString();
+        } catch(Exception e) {
+            log.trace("Unable to filter body, is it a JSON?. Keep as-is. `{}`", e.getMessage());
```
whiskeysierra: Message doesn't really match what's happening.

skjolber: Better?

whiskeysierra:
:+1:

whiskeysierra:
Thanks for the contribution! :tada:

PR #513, opened 2019-05-20
Support for filtering request path

### LINKED ISSUE #499: Support for filtering path
Add hooks so that the path part of an URI can be filtered 

/a/b/c

prints as

/a/xxx/b

## Detailed Description
Something like the QueryFilter, but just for the path instead. 

## Context
Some of our legacy systems are passing values which we would like to filter out as part of the URI path.

## Possible Implementation
Something like the QueryFilter, but just for the path instead. We currently have an implementation which uses RequestFilter and a PathFilterHttpRequest (with delegate HttpRequest). The first attempt was using RawRequestFilter but that did not pan out.

## Your Environment
Mostly Spring with Java, Kotlin and Scala.

skjolber:
Support for filtering the request path.

## Description
Add PathFilter interface (much liker HeaderFilter and QueryFilter). 

## Motivation and Context
REST services often expose important values directly in their paths. Add support for filtering these. 
Fixes #499 

## Types of changes

## Checklist:


logbook-api/src/main/java/org/zalando/logbook/LogbookCreator.java
```diff
@@ -38,6 +39,10 @@ private static Logbook create(
                 .reduce(QueryFilter::merge)
                 .orElse(null);
 
+        @Nullable final PathFilter pathFilter = pathFilters.stream()
```
whiskeysierra: I'm a big fan of the [Open-closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle). What is the benefit of introducing a new filter type? This could be implemented as a specialized request and response filter, right?

skjolber: The primary benefit is being more developer-friendly. I am pretty sure it is more intuitive for developers with a path-filter when there also is a header, body and query filter; these four are the essential parts of HTTP after all. Also a working filter implementation will save time. 

So while much can be done with the RequestFilter and ResponseFilter, those are more 'catch all' in my view.

logbook-api/src/main/java/org/zalando/logbook/PathFilter.java
```diff
@@ -0,0 +1,21 @@
...
+
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+
+import org.apiguardian.api.API;
+
+@API(status = EXPERIMENTAL)
+@FunctionalInterface
+public interface PathFilter {
+
+    String filter(final String path);
```
skjolber: The path could be split into its segments, is that better?

whiskeysierra: Could we offer two versions of `filter`, one with `String path` the other with a `List<String> segments`?

skjolber: Sure, and the default implementation of the one with string is to just split into segments and call the other method?

whiskeysierra: Is there some ready-to-use library that we already have available? That default implementation should go in core, not api.

skjolber: @whiskeysierra Or perhaps just having a single filter method for `List<String> segments` is better?

whiskeysierra: Also fine

logbook-api/src/main/java/org/zalando/logbook/LogbookCreator.java
whiskeysierra: Fair enough.

logbook-api/src/main/java/org/zalando/logbook/PathFilter.java
skjolber: `List<String> segments` turned out to be a dead end, especially since it would not be a noop for the passthrough filter.

skjolber:
Moved away from the regexp filter - was able to make a simple default filter which should be lightning fast. It accepts path expressions on the form `/myApp/orders/{secret}/order`, which are commonly found in frameworks like Spring REST (i.e. when using @PathVariable).

logbook-core/src/main/java/org/zalando/logbook/FilteredHttpRequest.java
```diff
@@ -42,11 +45,17 @@ public String getQuery() {
...
-
+    
     @Override
     public Map<String, List<String>> getHeaders() {
         return headers;
     }
+    
+    @Override
+    public String getPath() {
+        final String path = request.getPath();
```
whiskeysierra: Should this be cached, similar to headers?

logbook-core/src/test/java/org/zalando/logbook/DefaultPathFilterTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+	@Test
+	public void testFilterOneReplacement() {
+		String path = "/profiles/123456789/user.json";
+		String expr = "/profiles/{organization}/user.json";
+		
+		PathFilter regexpPathUriFilter = new DefaultPathFilter("XXX", expr);
+		
+		String result = regexpPathUriFilter.filter(path);
+        assertThat(result, is("/profiles/XXX/user.json"));
```
whiskeysierra: Indent?

logbook-core/src/test/java/org/zalando/logbook/DefaultPathFilterTest.java
```diff
@@ -0,0 +1,132 @@
+package org.zalando.logbook;
+
+import static org.hamcrest.CoreMatchers.sameInstance;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+import org.junit.jupiter.api.Test;
+
+public class DefaultPathFilterTest {
```
whiskeysierra: Reformatting the whole class should fix most issues.

logbook-core/src/test/java/org/zalando/logbook/FilteredHttpRequestTest.java
```diff
@@ -19,8 +19,11 @@
             .withHeaders(MockHeaders.of(
                     "Authorization", "Bearer 9b7606a6-6838-11e5-8ed4-10ddb1ee7671",
                     "Accept", "text/plain"))
-            .withBodyAsString("My secret is s3cr3t"),
+            .withBodyAsString("My secret is s3cr3t")
+            .withPath("/endpoint/secret/action")
+            ,
```
whiskeysierra: Move to previous line?

logbook-core/src/main/java/org/zalando/logbook/FilteredHttpRequest.java
skjolber: Why not, also adding caching for query to be consistent then.

whiskeysierra:
:+1:

PR #524, opened 2019-06-04
Improve performance of DefaultHttpLogFormatter

skjolber:
Improve performance of DefaultHttpLogFormatter.

## Description
Construct log statement in a single 'streaming' fasion.

## Motivation and Context
The previous implementation used a two-step approach via StructuredHttpLogFormatters format + prepare. Switching to a single step approach improves performance considerably, run HttpLogFormatterBenchmark for comparisons.

## Types of changes

## Checklist:

### DIFF
--- a/logbook-api/src/main/java/org/zalando/logbook/RequestURI.java
+++ b/logbook-api/src/main/java/org/zalando/logbook/RequestURI.java
@@ -10,33 +10,33 @@ import static org.zalando.logbook.RequestURI.Component.PATH;
 import static org.zalando.logbook.RequestURI.Component.QUERY;
 import static org.zalando.logbook.RequestURI.Component.SCHEME;
 
-public final class RequestURI {
+final class RequestURI {
 
     private RequestURI() {
 
     }
 
-    public static enum Component {
+    enum Component {
         SCHEME, AUTHORITY, PATH, QUERY
     }
 
-    public static String reconstruct(final HttpRequest request) {
+    static String reconstruct(final HttpRequest request) {
         final StringBuilder url = new StringBuilder();
         reconstruct(request, url);
         return url.toString();
     }
 
-    public static void reconstruct(final HttpRequest request, StringBuilder output) {
+    static void reconstruct(final HttpRequest request, StringBuilder output) {
         reconstruct(request, EnumSet.allOf(Component.class), output);
     }
 
-    public static String reconstruct(final HttpRequest request, final Component... components) {
+    static String reconstruct(final HttpRequest request, final Component... components) {
         final StringBuilder url = new StringBuilder();
         reconstruct(request, EnumSet.copyOf(asList(components)), url);
         return url.toString();
     }
 
-    public static String reconstruct(final HttpRequest request, final Set<Component> components) {
+    static String reconstruct(final HttpRequest request, final Set<Component> components) {
         final StringBuilder url = new StringBuilder();
         reconstruct(request, components, url);
         return url.toString();
--- a/logbook-api/src/main/java/org/zalando/logbook/RequestURI.java
+++ b/logbook-api/src/main/java/org/zalando/logbook/RequestURI.java
@@ -10,35 +10,47 @@ import static org.zalando.logbook.RequestURI.Component.PATH;
 import static org.zalando.logbook.RequestURI.Component.QUERY;
 import static org.zalando.logbook.RequestURI.Component.SCHEME;
 
-final class RequestURI {
+public final class RequestURI {
 
     private RequestURI() {
 
     }
 
-    enum Component {
+    public static enum Component {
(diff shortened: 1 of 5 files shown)


logbook-api/src/main/java/org/zalando/logbook/RequestURI.java
```diff
@@ -10,35 +10,47 @@
 import static org.zalando.logbook.RequestURI.Component.QUERY;
 import static org.zalando.logbook.RequestURI.Component.SCHEME;
 
-final class RequestURI {
+public final class RequestURI {
```
skjolber: Open for suggestions how to do this better

whiskeysierra: Why does it need to be public?

whiskeysierra:
> Improve performance of DefaultHttpLogFormatter.

The README states that

> [..] provided by the DefaultHttpLogFormatter. It is primarily designed to be used for local development and debugging, **not for production use**.
>
> https://github.com/zalando/logbook#http

I'm not saying that we shouldn't take low-hanging fruits and make it faster, but if it requires semi-significant changes to our API (seems to me that it does) then we should re-consider.

logbook-api/src/main/java/org/zalando/logbook/RequestURI.java
whiskeysierra: :ping_pong: :wink:

logbook-api/src/main/java/org/zalando/logbook/RequestURI.java
```diff
@@ -16,29 +16,41 @@ private RequestURI() {
 
     }
 
-    enum Component {
+    static enum Component {
```
whiskeysierra: Not needed for enums

logbook-api/src/main/java/org/zalando/logbook/RequestURI.java
```diff
@@ -16,29 +16,41 @@ private RequestURI() {
...
         SCHEME, AUTHORITY, PATH, QUERY
     }
 
-    static String reconstruct(final HttpRequest request) {
-        return reconstruct(request, EnumSet.allOf(Component.class));
+    public static String reconstruct(final HttpRequest request) {
```
whiskeysierra: All methods can drop the public modified as well, right?

skjolber: Yes, forgot.

skjolber:
Visibility changes to API should be reverted now, must had leaked from previously adapting JsonHttpLogFormatter in the same branch. :eyes:

My initial description is incorrect, DefaultHttpLogFormatter does not currently use StructuredHttpLogFormatter, so this PR is essentially moving to directly writing with a StringBuilder instead of collecting a list of lines and appending them at the end. 

I'll admit the performance of the HttpLogFormatter is not that important, but at least it is quite a lot faster now (like 10x). We're actually running both the HttpLogFormatter and the JsonHttpLogFormatter for build / local development, one is printed to console and the other is used for log statement unit testing.

skjolber:
BTW: Has Ansi coding been discussed?

whiskeysierra:
> I'll admit the performance of the HttpLogFormatter is not that important, but at least it is quite a lot faster now (like 10x). We're actually running both the HttpLogFormatter and the JsonHttpLogFormatter for build / local development, one is printed to console and the other is used for log statement unit testing.

I actually don't mind having this change. The code is still readable overall. 10x is quite appealing, tbh.

> BTW: Has Ansi coding been discussed?

What do you mean?

logbook-api/src/main/java/org/zalando/logbook/RequestURI.java
```diff
@@ -63,13 +75,11 @@ private static String reconstruct(final HttpRequest request, final Set<Component
...
             url.append('?').append(query);
         }
-
-        return url.toString();
     }
 
     private static boolean isNotStandardPort(final String scheme, final int port) {
-        return "http".equals(scheme) && port != 80 ||
-                "https".equals(scheme) && port != 443;
+        return ("http".equals(scheme) && port != 80) ||
```
whiskeysierra: Those are not really needed, are they?

skjolber: Strictly, no, it was intended as an improvement in readability, forgot to comment.

skjolber:
Ansi coding as in colors for improved readablity for local development and debugging. I guess it could be a future improvement (i.e. enabled by a config parameter).

whiskeysierra:
> Ansi coding as in colors for improved readablity for local development and debugging. I guess it could be a future improvement (i.e. enabled by a config parameter).

Let's move this to a separate issue.

whiskeysierra:
:+1:

PR #530, opened 2019-06-05
Added primitive json number replacement filter

### LINKED ISSUE #478: Add a way to replace non String Json Properties
It would be convenient to be able to replace numeric values in json the same way string values are replaced 

## Detailed Description
public static BodyFilter replaceJsonStringProperty(final Set<String> properties, final String replacement)
could become 
public static BodyFilter replaceJsonProperties(final Set<String> properties, final String stringReplacement, float numericReplacement)

## Context
I need to obfuscate numerci data in my json request and responses because they are sensitive data i don't want to log.
I believe I wouldn't be the only one with this need as it is foreseen for strings but not numeric values

BTW: why not allowing inheritance with final classes and why the package private classes ?
Thanks for this library that could really fulfill our needs if this feature request was implemented.

whiskeysierra:
## Motivation and Context

Fixes #478

## Types of changes

## Checklist:

### DIFF
--- a/logbook-json/src/main/java/org/zalando/logbook/json/JsonBodyFilters.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/JsonBodyFilters.java
@@ -9,7 +9,9 @@ import java.util.Set;
 import java.util.function.UnaryOperator;
 import java.util.regex.Pattern;
 
+import static java.util.regex.Pattern.compile;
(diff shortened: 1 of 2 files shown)


whiskeysierra:
@lukasniemeier-zalando :+1::question:

whiskeysierra:
@AlexanderYastrebov :+1::question:

fatroom:
👍

PR #536, opened 2019-06-14
Improve performance of JsonHttpLogFormatter: StringBuilder

skjolber:
Full optimization of JsonHttpLogFormatter.

Original:
Benchmark                                Mode  Cnt       Score     Error  Units
HttpLogFormatterBenchmark.jsonRequest thrpt 5 **291434**,027 ± 614,715 ops/s
HttpLogFormatterBenchmark.jsonResponse thrpt 5 **324309**,826 ± 565,692 ops/s

Optimized:
Benchmark                                Mode  Cnt       Score      Error  Units
HttpLogFormatterBenchmark.jsonRequest   thrpt    5  **447403**,477 ±  521,810  ops/s
HttpLogFormatterBenchmark.jsonResponse  thrpt    5  **388240**,457 ± 1964,092  ops/s

So in other words approximately +54% for request and ~~+40%~~ +20% for response.

## Description
StringBuilder with various inlining, including JSON output. Improves performance at the cost of rougher code.

## Motivation and Context
Improve performance. Even if the code changes are considered a too high price for the performance benefit, at least it is now known how much there is to gain from the original implementation.

## Types of changes

## Checklist:

### DIFF
--- a/logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
@@ -1,89 +1,205 @@
 package org.zalando.logbook.json;
 
-import com.fasterxml.jackson.annotation.JsonRawValue;
-import com.fasterxml.jackson.annotation.JsonValue;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import lombok.AllArgsConstructor;
-import org.apiguardian.api.API;
-import org.zalando.logbook.HttpLogFormatter;
-import org.zalando.logbook.HttpMessage;
-import org.zalando.logbook.StructuredHttpLogFormatter;
-
-import java.io.IOException;
-import java.util.Map;
-import java.util.Optional;
-
 import static org.apiguardian.api.API.Status.STABLE;
 
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.Map.Entry;
+
+import org.apiguardian.api.API;
+import org.zalando.logbook.Correlation;
+import org.zalando.logbook.HttpLogFormatter;
+import org.zalando.logbook.HttpMessage;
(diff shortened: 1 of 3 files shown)


whiskeysierra:
> Even if the code changes are considered a too high price for the performance benefit, at least it is now known how much there is to gain from the original implementation.

Code changes are one thing but we're actually loosing functionality that has been requested a couple of times which is a way to extend the data structure that is produced by the formatter.

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
```diff
@@ -1,92 +1,204 @@
...
+        encode(request.getProtocolVersion(), builder);
+        builder.append("\",\"remote\":\"");
+        encode(request.getRemote(), builder);
+        builder.append("\",\"method\":\"");
+        encode(request.getMethod(), builder);
+        builder.append("\",\"uri\":\"");
+        reconstruct(request, builder);
+        builder.append('"');
 
-    private final ObjectMapper mapper;
```
whiskeysierra: Instead of going from `ObjectMapper` (relatively high-level) to `StringBuilder` (lowest level) maybe we could try to re-implement this using `JsonGenerator`: https://github.com/FasterXML/jackson-docs/wiki/JacksonStreamingApi#creating-generators

Extending the data structure can be achieved by allowing to register callbacks to be invoked with the generator before closing the outer most object.

skjolber: I tried that and was not too impressed with the improvement, but I guess I can dig up the code and see if it can be tweaked further.

skjolber: See https://github.com/zalando/logbook/pull/541, it is better than the original, but overall a bit disappointing result.

skjolber:
Closed in favor of https://github.com/zalando/logbook/pull/541

PR #541, opened 2019-06-22
Improve performance of JsonHttpLogFormatter: Jackson JsonFactory

skjolber:
Optimization of JsonHttpLogFormatter. See https://github.com/zalando/logbook/pull/536.

Original
Benchmark                                   Mode  Cnt       Score      Error  Units
HttpLogFormatterBenchmark.defaultRequest   thrpt    5  756914,492 ± 2507,324  ops/s
HttpLogFormatterBenchmark.defaultResponse  thrpt    5  565435,993 ± 1866,147  ops/s
HttpLogFormatterBenchmark.jsonRequest      thrpt    5  **291434**,027 ±  614,715  ops/s
HttpLogFormatterBenchmark.jsonResponse     thrpt    5  **324309**,826 ±  565,692  ops/s

Optimized:
Benchmark                                   Mode  Cnt       Score      Error  Units
HttpLogFormatterBenchmark.defaultRequest   thrpt    5  749361,362 ± 1422,486  ops/s
HttpLogFormatterBenchmark.defaultResponse  thrpt    5  565017,412 ± 2198,724  ops/s
HttpLogFormatterBenchmark.jsonRequest      thrpt    5  **370198**,126 ±  569,101  ops/s
HttpLogFormatterBenchmark.jsonResponse     thrpt    5  **323354**,622 ± 1911,860  ops/s

So in other words approximately +27% for request and no improvement for response.
## Description
Jackson JsonFactory with various inlining, including JSON output. Improves performance at the cost of rougher code.

## Motivation and Context
Improve performance.

## Types of changes

## Checklist:

Note: Will add full coverage and extension point if this PR is preferred over the StringBuilder variant.


skjolber:
@whiskeysierra so which do you guys prefer, this or the StringBuilder equivalent?

whiskeysierra:
I prefer this, because it's a good compromise between speed and usability

skjolber:
@whiskeysierra  So about 

> Code changes are one thing but we're actually loosing functionality that has been requested a couple of times which is a way to extend the data structure that is produced by the formatter.

this is quite lost when `JsonHttpLogFormatter` is a final class. Should the class be non-final with protected methods, or go for final with private methods?

whiskeysierra:
I'd favor this one over #536. The numbers are hugely different and the jackson based version to me is more maintainable and extensible.

whiskeysierra:
@skjolber Do you mind rebasing?

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
```diff
@@ -1,92 +1,191 @@
...
+
+        StringBuilderWriter writer = new StringBuilderWriter(body.length() + 2048);
+
+        JsonGenerator generator = jsonFactory.createGenerator(writer);
+        try {
+            generator.writeStartObject();
+            
+            generator.writeFieldName("origin");
+            if(request.getOrigin() == Origin.LOCAL) {
+                generator.writeRawValue("\"local\"");
```
whiskeysierra: Why not `writeString("local")`?

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
```diff
@@ -1,92 +1,191 @@
...
+
+        final String scheme = request.getScheme();
+        builder.append(scheme);
+        builder.append("://"); // forward slash escaping is optional, so don't escape
+        
+        // host legal characters: Letters, Numbers 0-9 and Hyphen - none of which are escaped in JSON
+        builder.append(request.getHost()); 
+
+        final Optional<Integer> port = request.getPort();
+        if(port.isPresent() && isNotStandardPort(scheme, port.get())) {
```
whiskeysierra:
```suggestion
        if (port.isPresent() && isNotStandardPort(scheme, port.get())) {
```

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
```diff
@@ -1,92 +1,191 @@
...
+            builder.append(':').append(port.get());
         }
+        
+        JsonStringEncoder jsonStringEncoder = BufferRecyclers.getJsonStringEncoder();
+        jsonStringEncoder.quoteAsString(request.getPath(), builder); // escape
```
whiskeysierra: The `JsonGenerator` doesn't give you this?

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
```diff
@@ -1,92 +1,191 @@
...
+        
+        if(!headers.isEmpty()) {
+            // implementation note:
+            // for some unclear reason, manually iterating over the headers
+            // while writing performs worse than letting Jackson do the job.
+            builder.writeObjectField("headers", headers);
+        }
+    }
+    
+    private void reconstruct(final HttpRequest request, JsonGenerator generator, StringBuilderWriter writer) throws IOException {
```
whiskeysierra:
```suggestion
    private void reconstructUri(final HttpRequest request, JsonGenerator generator, StringBuilderWriter writer) throws IOException {
```

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
```diff
@@ -1,92 +1,191 @@
...
+        
+        final String correlationId = correlation.getId();
+        
+        String body = response.getBodyAsString();
+
+        StringBuilderWriter writer = new StringBuilderWriter(body.length() + 2048);
+
+        JsonGenerator generator = jsonFactory.createGenerator(writer);
+
+        try {
```
whiskeysierra:
```suggestion
        try (final JsonGenerator generator = jsonFactory.createGenerator(writer))  {
```

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
```diff
@@ -1,92 +1,191 @@
...
+            generator.writeStringField("protocol", response.getProtocolVersion());
+            generator.writeNumberField("duration", correlation.getDuration().toMillis());
+            generator.writeNumberField("status", response.getStatus());
+    
+            writeHeaders(generator, response);
+            writeBody(response, body, generator);
+            
+            generator.writeEndObject();
+        } finally {
+            generator.close();
```
whiskeysierra:
```suggestion
```

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
skjolber: Marginally faster; no escaping necessary.

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
skjolber: It does, however I've tried to avoid escaping whenever possible. So reconstructUri is a bit ugly. Would be possible to make it a bit less low-level and see how that affects performance.

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
```diff
@@ -1,92 +1,184 @@
...
     }
+    
+    public JsonHttpLogFormatter(JsonFactory jsonFactory) {
+        this.jsonFactory = jsonFactory;
+    }
 
-    public JsonHttpLogFormatter(final ObjectMapper mapper) {
-        this.mapper = mapper;
+    public JsonHttpLogFormatter(ObjectMapper objectMapper) {
+        this.jsonFactory = objectMapper.getFactory();
```
whiskeysierra:
```suggestion
        this(objectMapper.getFactory());
```

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
```diff
@@ -1,92 +1,184 @@
...
         this(new ObjectMapper());
     }
+    
+    public JsonHttpLogFormatter(JsonFactory jsonFactory) {
+        this.jsonFactory = jsonFactory;
+    }
 
-    public JsonHttpLogFormatter(final ObjectMapper mapper) {
-        this.mapper = mapper;
+    public JsonHttpLogFormatter(ObjectMapper objectMapper) {
```
whiskeysierra: Should swap places with the constructor above.

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
```diff
@@ -1,92 +1,184 @@
...
+    }
+
+    private void writeBody(HttpMessage message, String body, JsonGenerator builder) throws IOException {
+        if(!body.isEmpty()) {
+            builder.writeFieldName("body");
+
+            final String contentType = message.getContentType();
+
+            if (JsonMediaType.JSON.test(contentType)) {
+                builder.writeRawValue(body);
```
whiskeysierra: This is now way more expressive than in the original version that used `@JsonRawValue`. :+1:

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
```diff
@@ -1,92 +1,184 @@
...
+        
+        if(!headers.isEmpty()) {
+            // implementation note:
+            // for some unclear reason, manually iterating over the headers
+            // while writing performs worse than letting Jackson do the job.
+            builder.writeObjectField("headers", headers);
+        }
     }
+    
+    private void reconstructUri(final HttpRequest request, JsonGenerator generator, StringBuilderWriter writer) throws IOException {
```
whiskeysierra: Passing two *sinks* here is a bit strange.

logbook-json/src/main/java/org/zalando/logbook/json/StringBuilderWriter.java
```diff
@@ -0,0 +1,125 @@
...
+ * Adapted from Apache commons-io equivalent. 
+ * <p>
+ * <strong>NOTE:</strong> This implementation, as an alternative to
+ * <code>java.io.StringWriter</code>, provides an <i>un-synchronized</i>
+ * (i.e. for use in a single thread) implementation for better performance.
+ * For safe usage with multiple {@link Thread}s then
+ * <code>java.io.StringWriter</code> should be used.
+ *
+ */
+class StringBuilderWriter extends Writer implements Serializable {
```
whiskeysierra: Why `Serializable`?

whiskeysierra:
```suggestion
final class StringBuilderWriter extends Writer implements Serializable {
```

skjolber: It was in the original source code. I'll remove it.

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
skjolber: I can't remember how much of a difference the current reconstructUri(..) version did, so I'll give it a go later and compare against just doing this 'the normal way'.

whiskeysierra:
You're missing some code coverage there.

skjolber:
Re-ran the benchmarks and the StringBuilderWriter did not make such a difference as my initial test indicated (must have been something wrong with it). So that simplifies things a bit.

whiskeysierra:
Some lines are not covered yet.

skjolber:
I've got another minor PR after this is merged, making sure all JSON parse / serialize is streaming + using JsonFactory instead of ObjectMapper if possible.

skjolber:
Final numbers:
jmhJsonHttpLogFormatter3:
Benchmark                                   Mode  Cnt       Score      Error  Units
HttpLogFormatterBenchmark.defaultRequest   thrpt    5  887362.552 ± 2880.713  ops/s
HttpLogFormatterBenchmark.defaultResponse  thrpt    5  678167.697 ± 6525.555  ops/s
HttpLogFormatterBenchmark.jsonRequest      thrpt    5  407298.683 ± 3392.139  ops/s
HttpLogFormatterBenchmark.jsonResponse     thrpt    5  375502.031 ± 3487.851  ops/s

Master:
Benchmark                                   Mode  Cnt       Score      Error  Units
HttpLogFormatterBenchmark.defaultRequest   thrpt    5  816553.280 ± 6368.931  ops/s
HttpLogFormatterBenchmark.defaultResponse  thrpt    5  672247.206 ± 1982.301  ops/s
HttpLogFormatterBenchmark.jsonRequest      thrpt    5  349667.127 ± 3373.876  ops/s
HttpLogFormatterBenchmark.jsonResponse     thrpt    5  333637.318 ± 2923.868  ops/s

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
```diff
@@ -1,92 +1,169 @@
...
- *         return delegate.format(response);
- *      }
- *
- * }
- * }
- * </pre>
+ * A custom {@link HttpLogFormatter} that produces JSON objects. 
  */
 @API(status = STABLE)
-public final class JsonHttpLogFormatter implements StructuredHttpLogFormatter {
```
whiskeysierra: Technically that's a breaking change since the public API of the class changed.

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
```diff
@@ -1,92 +1,169 @@
...
 
-    @AllArgsConstructor
-    private static final class JsonBody {
-        String json;
+    protected void writeHeaders(JsonGenerator builder, HttpMessage httpMessage) throws IOException {
+        Map<String, List<String>> headers = httpMessage.getHeaders();
+        
+        if(!headers.isEmpty()) {
+            // implementation note:
+            // for some unclear reason, manually iterating over the headers
```
whiskeysierra: If that is the case for headers, maybe sticking to a map-based approach (see `StructuredHttpLogFormatter`) would yield similar results without introducing a breaking API change?

skjolber: Not 100% sure, but I think it has something to do with a smart way of serializing arrays/Lists in a Map. I'm able to make something thats almost as fast. I'm sure there is no magic, but did not pursue it further due to time, so could be a future improvement.

But to answer your question, no, then the performance improvement would be gone?

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
skjolber: You're technically correct, however

 * StructuredHttpLogFormatter is still status experimental, so JsonHttpLogFormatter should be too?
 * You're about to do a new major version, some API changes are to be expected?
 * The class is final, so there exist no subclasses which would (potentially silently) break?

So can I leave it as is?

whiskeysierra: > * StructuredHttpLogFormatter is still status experimental, so JsonHttpLogFormatter should be too?
> * You're about to do a new major version, some API changes are to be expected?

Both true.

> * The class is final, so there exist no subclasses which would (potentially silently) break?

Those methods are public and the preferred usage is composition where users are allowed to call the public methods, e.g. by augmenting the map before passing it down.

skjolber: So your point is that those who now change (add / remove / transform) the output JSON format are now unable to do so using the same JsonHttpLogFormatter. 

Letting JsonHttpLogFormatter implement StructuredHttpLogFormatter is still possible, then HttpLogFormatter#format methods run the optimized code, while the StructuredHttpLogFormatter#prepare / #format still runs as today. The will be some small overlap in functionality, but not that much really.

whiskeysierra: > The will be some small overlap in functionality, but not that much really.

Overlap in functionality without overlap in code, i.e. there is no reuse between the two approaches. In that case I'd actually rather split those two use cases up into two classes (banking on the fact that we're going to release a major version anyway, so breaking changes are easy right now).

whiskeysierra:
I added #580 as an alternative.

skjolber:
Very good, lets go with #580. And sorry this has taken so long to finish, it has been summer holiday.

PR #586, opened 2019-08-23
GH-585 Fixed PUT without body

### LINKED ISSUE #585: PUT method with no body
When a PUT request has no body set, logbook throw an error

## Description
In module logbook httpclient, when a PUT request is logged, it expects a body to be present. But in some cases, a PUT request can have no body (especially true when creating indices on elasticsearch)

  

## Expected Behavior
Just treat a PUT request with no body set as a request without body

## Possible Fix
https://gist.github.com/hcapitaine/0e92994ceb1706a166c2a7db5df9332c

Method withBody:134, add conditional test and treat it as withoutBody when no body set.

## Steps to Reproduce
See this test as an example:
https://gist.github.com/hcapitaine/f05a522dce89fa9d961a3610df19db86

whiskeysierra:
## Motivation and Context

Fixes #585

## Types of changes

## Checklist:


whiskeysierra:
:+1:

PR #617, opened 2019-10-09
Fixed missing flush for writer

### LINKED ISSUE #603: Logbook removes body from response
I encountered an issue in combination with CXF and Spring Security where the body of an response vanishes (=is null) when using Logbook.

## Description
In Logbook version 1.13.0 the body of the response vanishes when the Writer Level is set to something that would lead to requests not getting logged (=INFO).

Using Logbook version 2.0.0-RC.33 the behaviour is inverted, e. g. the body disappears when the level would actually log it (=TRACE).

The Logbook Writer Level setting influences this behaviour. 

The setup is very simple:

The Spring Boot App exposes a CXF service list under /services. This resource is accessible for everyone (permitAll).

GET requests to /services do not yield any body when logging is turned _off_ by changing the log level of Logbook. When I turn the logging _on_, the body is contained in the response as it would be without Logbook.

## Expected Behavior
The response should contain the original body.

## Actual Behavior
The body of the response is null.

## Possible Fix
I have no idea, sadly.

## Steps to Reproduce
Reproduce erroneus behaviour :
1. Check out https://github.com/Breeki32/logbooktest113
2. Start app and send GET request to /services
or simply run the tests

Reproduce correct behaviour :
1. Check out https://github.com/Breeki32/logbooktest113
2. set logging.level.org.zalando property to TRACE
2. Start app and send GET request to /services
or simply run the tests

Reproduce erroneus behaviour with **2.0.0-RC.3**:
1. Check out https://github.com/Breeki32/logbooktest113
2. set Logbook version to 2.0.0-RC.3 in pom.xml
2. set logging.level.org.zalando property to TRACE
or simply run the tests with given version

## Context
I´ve noticed that certain things did not work in an web application when the log level was set to INFO. I tracked it down to the combination of CXF, Spring Security and Logbook (the minimal setup in the example project). The use case was logging the traffic to and from SOAP services for development purposes - but I noticed missing data in the service list servlet. 

## Your Environment

* Version used: Logbook 1.13.0 - 2.0.0-RC3
* Spring Boot 2.1.6 - 2.1.8
* CXF 3.3.2 - 3.3.3
* Link to your project: https://github.com/Breeki32/logbooktest113

Thank you for looking into this!

whiskeysierra:
## Description

Turns out that my fake controller in tests behaved too correctly. It always closed (and implicitly flushed) the writer it was using. But that is in fact not done by everyone out there in the wild. Ultimately the `flushBuffer` method is invoked on the response, but that wasn't intercepted and propagated properly. The noticable effect was that users which use `getWriter()` without an explicit `close` or `flush` would not see the body getting logged and even worse wouldn't see the body on the wire either.

## Motivation and Context

Fixes #603

## Types of changes

## Checklist:

### DIFF
--- a/logbook-servlet/src/test/java/org/zalando/logbook/servlet/ExampleController.java
+++ b/logbook-servlet/src/test/java/org/zalando/logbook/servlet/ExampleController.java
@@ -13,7 +13,6 @@ import javax.servlet.ServletOutputStream;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
-import java.io.PrintWriter;
 import java.nio.CharBuffer;
 import java.util.Objects;
 import java.util.concurrent.Callable;
@@ -95,9 +94,7 @@ public class ExampleController {
 
     @RequestMapping(path = "/reader", produces = TEXT_PLAIN_VALUE)
     public void reader(final HttpServletRequest request, final HttpServletResponse response) throws IOException {
-        try (final PrintWriter writer = response.getWriter()) {
-            copy(request.getReader(), writer);
-        }
+        copy(request.getReader(), response.getWriter());
     }
 
     @RequestMapping(path = "/binary", produces = MediaType.APPLICATION_OCTET_STREAM_VALUE)
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
@@ -107,6 +107,14 @@ final class LocalResponse extends HttpServletResponseWrapper implements HttpResp
     }
 
     @Override
+    public void flushBuffer() throws IOException {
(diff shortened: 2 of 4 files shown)


logbook-servlet/src/test/java/org/zalando/logbook/servlet/ExampleController.java
```diff
@@ -95,9 +94,7 @@ public void stream(final HttpServletRequest request, final HttpServletResponse r
 
     @RequestMapping(path = "/reader", produces = TEXT_PLAIN_VALUE)
     public void reader(final HttpServletRequest request, final HttpServletResponse response) throws IOException {
-        try (final PrintWriter writer = response.getWriter()) {
-            copy(request.getReader(), writer);
-        }
+        copy(request.getReader(), response.getWriter());
```
whiskeysierra: That change was enough to reproduce the issue in a test.

lukasniemeier-zalando:
:+1:

PR #699, opened 2020-01-26
Use @Configuration(proxyBeanMethods=false) to avoid unnecessary gener…

PascalSchumacher:
…ation of cglib proxies. Cglib proxy creation can be an expensive operation.

Similar to https://github.com/spring-projects/spring-boot/issues/9068

I think this is a breaking change if somebody calls methods on the configuration classes.


whiskeysierra:
Thanks for the PR! I'll need to fix the coverage requirement/action first before I can merge this.

whiskeysierra:
:+1:

whiskeysierra:
:tada: Thanks for the contribution!

PR #710, opened 2020-02-11
Made form request mode configurable via configuration properties

### LINKED ISSUE #701: "Required request body is missing" when POSTing form
When POSTing to an endpoint with form content, "Required request body is missing" exception is thrown and a 400 error is returned.

## Expected Behavior
The request should be logged and a 200 returned

## Actual Behavior
An exception is thrown and a 400 is returned

## Possible Fix
It seems that setting the system property `logbook.servlet.form-request` to "PARAMETER" works around this issue, but an exception shouldn't be thrown with the default setup. Maybe calling getParameterMap before getInputStream in RemoteRequest would solve this.

It may also be better to read the `logbook.servlet.form-request` from application properties, similar to how `logbook.filter.enabled` is set.

## Steps to Reproduce
I've provided a minimal example to reproduce this issue at https://github.com/jbonzowh/logbook_bug

Run the project then run this curl command `curl http://localhost:8080/message -H 'Content-Type: application/x-www-form-urlencoded; charset=UTF-8' --data 'accountId=90240'`

## Your Environment
* Version used: logbook-spring-boot-starter 1.13.0, spring-boot 2.2.4.RELEASE, Java 8
* Link to your project: https://github.com/jbonzowh/logbook_bug

whiskeysierra:
Fixes #701

## Motivation and Context

Made form request mode configurable using configuration properties.

## Types of changes

## Checklist:

### DIFF
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LogbookFilter.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LogbookFilter.java
@@ -1,5 +1,7 @@
 package org.zalando.logbook.servlet;
 
+import lombok.AllArgsConstructor;
+import lombok.With;
 import org.apiguardian.api.API;
 import org.zalando.logbook.HttpRequest;
 import org.zalando.logbook.Logbook;
@@ -17,9 +19,11 @@ import java.io.IOException;
 import java.util.UUID;
 
 import static javax.servlet.DispatcherType.ASYNC;
+import static lombok.AccessLevel.PRIVATE;
 import static org.apiguardian.api.API.Status.STABLE;
 
 @API(status = STABLE)
+@AllArgsConstructor(access = PRIVATE)
 public final class LogbookFilter implements HttpFilter {
 
     /**
@@ -31,6 +35,9 @@ public final class LogbookFilter implements HttpFilter {
     private final Logbook logbook;
     private final Strategy strategy;
 
+    @With
+    private final FormRequestMode formRequestMode;
+
     public LogbookFilter() {
         this(Logbook.create());
     }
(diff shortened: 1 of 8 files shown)


logbook-servlet/src/main/java/org/zalando/logbook/servlet/LogbookFilter.java
```diff
@@ -17,9 +19,11 @@
 import java.util.UUID;
 
 import static javax.servlet.DispatcherType.ASYNC;
+import static lombok.AccessLevel.PRIVATE;
 import static org.apiguardian.api.API.Status.STABLE;
 
 @API(status = STABLE)
+@AllArgsConstructor(access = PRIVATE)
```
lukasniemeier-zalando: Intentionally hiding the constructor? Is there no use case for anyone programatically instantiating a `LogbookFilter` with a explicit `FormRequestMode`?

logbook-servlet/src/test/java/org/zalando/logbook/servlet/FormRequestTest.java
```diff
@@ -0,0 +1,118 @@
...
+ * Verifies that {@link LogbookFilter} delegates to {@link HttpLogWriter} correctly.
+ */
+final class FormRequestTest {
+
+    private final HttpLogFormatter formatter = spy(new ForwardingHttpLogFormatter(new DefaultHttpLogFormatter()));
+    private final HttpLogWriter writer = mock(HttpLogWriter.class);
+
+    @BeforeEach
+    void setUp() {
+        reset(formatter, writer);
```
lukasniemeier-zalando: Why is such `reset` required?

logbook-servlet/src/main/java/org/zalando/logbook/servlet/LogbookFilter.java
whiskeysierra: I prefer to expose only the constructor with required arguments. So that a user get's the default by invoking the constructor but can subsequently override/configure things as needed using `with*` methods (convention over configuration).

logbook-servlet/src/test/java/org/zalando/logbook/servlet/FormRequestTest.java
whiskeysierra: Nope, not needed. Probably a left over from an even earlier version of this test. Or just copy&pasted from somewhere else without giving it another though.

lukasniemeier-zalando: Ok thanks, I had a short panic attack that with JUnit 5 the test class life cycle has changed :-D

logbook-servlet/src/main/java/org/zalando/logbook/servlet/LogbookFilter.java
lukasniemeier-zalando: Ah I've missed the `@With` annotation on the field, makes sense :+1:

lukasniemeier-zalando:
:+1:

PR #711, opened 2020-02-11
Fixed old code example in javadoc

### LINKED ISSUE #691: JavaDoc of JsonHttpLogFormatter isn't in line with method signatures
Currently, it's only possible to log request/response's in isolation. That is, the request contains the URI etc and the response contains the duration and status. It would be nice to be able to log request's details (mainly the URI) alongside the response for easier lockup.

## Detailed Description
Filtering logs by uri is only possible on the request logs. Therefore, it's not easily possible to grap "all failed responses against _some URI_" etc. Due to the way the API is designed neither [`HttpLogFormat#format(Correlation correlation, HttpResponse response)`](https://github.com/zalando/logbook/blob/eb15e9adb829ec840dcc93358236b1d2c98dee3d/logbook-api/src/main/java/org/zalando/logbook/HttpLogFormatter.java#L13) nor [`HttpResponse`](https://github.com/zalando/logbook/blob/eb15e9adb829ec840dcc93358236b1d2c98dee3d/logbook-api/src/main/java/org/zalando/logbook/HttpResponse.java) allow access to the associated request details, which rules out custom extensions of the response format (see possible hack below).

It seems from [`JsonHttpLogFormatter`'s (now) incorrect Javadoc](https://github.com/zalando/logbook/blob/master/logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java#L32-L42), that the desired functionality used to be possible at some point. Or maybe the JavaDoc happened to be out-of-sync with the code from the beginning?

## Context
Often, when analysing certain patterns in log it's the responses that are looked after. Unfortunately, there's currently no way to extend the `JsonHttpLogFormat` to include the `uri` within the resulting JSON.

## Possible Implementation
1. Add `HttpRequest` as parameter to the format method (it is available on the [callside](https://github.com/zalando/logbook/blob/79cd28a2af44ceca7199bc405c68ee4347e30cc2/logbook-core/src/main/java/org/zalando/logbook/DefaultSink.java#L26))
2. Add `uri` to `HttpResponse`
3. (hacky workaround, haven't tried) Wrap `HttpResponse` in `Sink#write(final Correlation correlation, final HttpRequest request, final HttpResponse response)` before calling the formatter and implement a custom formatter that unwraps the response 🙄 
4. Won't fix. Use a more sophisticated logging tool.

## Your Environment
Logbook 2.0.0-RC.8

whiskeysierra:
Fixes #691

## Motivation and Context

Fixes #691 

## Types of changes

## Checklist:

### DIFF
--- a/logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
@@ -25,20 +25,20 @@ import static org.apiguardian.api.API.Status.STABLE;
  *
  *     private final JsonHttpLogFormatter delegate;
  *
- *     public CustomsFormatter(final ObjectMapper mapper) {
+ *     public CustomsFormatter(ObjectMapper mapper) {
  *         this.delegate = new JsonHttpLogFormatter(mapper);
  *     }
  *
- *     public String format(final Precorrelation<HttpRequest> precorrelation) throws IOException {
- *         Map<String, Object> request = delegate.prepare(precorrelation);
+ *     public String format(Precorrelation precorrelation, HttpRequest request) throws IOException {
+ *         Map<String, Object> content = delegate.prepare(precorrelation, request);
  *         // modify request here
- *         return delegate.format(request);
+ *         return delegate.format(content);
  *     }
  *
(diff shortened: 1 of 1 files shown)


lukasniemeier-zalando:
:+1:

PR #758, opened 2020-05-13
build: Maven CI Friendly Versions

bro0k:
http://maven.apache.org/maven-ci-friendly.html

## Types of changes

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,10 +5,11 @@
 ### Maven ###
 target/
 /.mvn/wrapper/*.jar
+.flattened-pom.xml
 
 *.log
 
-### Eclipse ### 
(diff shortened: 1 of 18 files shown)


whiskeysierra:
I fail to see the motivation for this? CI-friendly maven builds are meant for cases where the CI server builds, packages and deploys artifacts. In order to do that one would need to pass a generated version number, e.g. the build/run number from your CI/pipeline.

But that's not the approach that were pursuing with Logbook. Releasing a new version is a human decision and our tooling is the `release.sh` in the root of the repository.

Unless there is some convincing argument that I just didn't see yet, I'm inclined to reject this.

PR #811, opened 2020-08-12
Add logstash into spring autoconfiguration and provide logstash raw marker config possibility

pbouillet:
## Description
2 changes overall:
* Provide a possibility to change the logstash raw marker field name which contains all the information about the requests and responses
* Add a Spring configuration possibility to use the logstash sink and configure the raw marker field name

## Motivation and Context
The first change is motivated by an issue in our logging environment (ELK). As we use it for many applications, the http field name clashes with an already existing one which is indexed and defined as a different type. Because of this, we cannot even see the requests and responses logged by logbook. For this, I introduced a possibility to change this field name.
The second change is more of a convenient change for Spring autoconfiguration. Currently, we need to define an own Spring bean to be able to use this Logstash sink. I'm not sure if you have some requirements on if and how configuration changes should be for your framework, so feedback is pretty much appreciated (and if needed, I could roll back the second change if you don't feel this is a good way)
I added both changes in a way that the defaults are as before and no breaking changes are introduced.

## Types of changes

## Checklist:

### DIFF
--- a/logbook-logstash/src/main/java/org/zalando/logbook/logstash/LogstashLogbackSink.java
+++ b/logbook-logstash/src/main/java/org/zalando/logbook/logstash/LogstashLogbackSink.java
@@ -25,6 +25,12 @@ public final class LogstashLogbackSink implements Sink {
 
     private final HttpLogFormatter formatter;
 
+    private final String baseField;
+
+    public LogstashLogbackSink(final HttpLogFormatter formatter) {
+        this(formatter, "http");
+    }
+
     @Override
     public boolean isActive() {
         return log.isTraceEnabled();
@@ -32,7 +38,7 @@ public final class LogstashLogbackSink implements Sink {
 
     @Override
     public void write(final Precorrelation precorrelation, final HttpRequest request) throws IOException {
-        final Marker marker = new AutodetectPrettyPrintingMarker("http", formatter.format(precorrelation, request));
+        final Marker marker = new AutodetectPrettyPrintingMarker(baseField, formatter.format(precorrelation, request));
         log.trace(marker, requestMessage(request));
     }
 
@@ -43,7 +49,7 @@ public final class LogstashLogbackSink implements Sink {
     @Override
     public void write(final Correlation correlation, final HttpRequest request,
             final HttpResponse response) throws IOException {
-        final Marker marker = new AutodetectPrettyPrintingMarker("http", formatter.format(correlation, response));
+        final Marker marker = new AutodetectPrettyPrintingMarker(baseField, formatter.format(correlation, response));
 
         log.trace(marker, responseMessage(request, response));
     }
--- a/logbook-logstash/src/test/java/org/zalando/logbook/logstash/LogbackLogstashSinkTest.java
+++ b/logbook-logstash/src/test/java/org/zalando/logbook/logstash/LogbackLogstashSinkTest.java
@@ -44,100 +44,7 @@ class LogbackLogstashSinkTest {
 
     @Test
     void shouldLogRequestAndResponse() throws IOException {
-        final String correlationId = "3ce91230-677b-11e5-87b7-10ddb1ee7671";
-        final int duration = 125;
-
-        final Precorrelation precorrelation = mock(Precorrelation.class);
-        final Correlation correlation = mock(Correlation.class);
-
-        when(precorrelation.getId()).thenReturn(correlationId);
-        when(correlation.getId()).thenReturn(correlationId);
-        when(correlation.getDuration()).thenReturn(Duration.ofMillis(duration));
-
-        final HttpLogFormatter formatter = new JsonHttpLogFormatter();
-        final LogstashLogbackSink sink = new LogstashLogbackSink(formatter);
-
-        assertTrue(sink.isActive());
-
-        final HttpRequest request = MockHttpRequest.create()
-                .withProtocolVersion("HTTP/1.0")
-                .withOrigin(REMOTE)
-                .withPath("/test")
-                .withQuery("limit=1")
-                .withHeaders(HttpHeaders.empty()
-                        .update("Accept", "application/json")
-                        .update("Date", "Tue, 15 Nov 1994 08:12:31 GMT"))
-                .withContentType("application/json")
-                .withBodyAsString("{\"person\":{\"name\":\"Thomas\"}}");
-
-        sink.write(precorrelation, request);
-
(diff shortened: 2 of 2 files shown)


pbouillet:
Seems like there is new identified vulnerability in the kotlin-stdlib library that is used by logbook-okhttp: https://nvd.nist.gov/vuln/detail/CVE-2020-15824. As this is not related to my PR, I leave it as it is right now. Guess there is a change needed to either ignore or update to a newer version.

whiskeysierra:
Yes, I'll need to deal with this separately.

whiskeysierra:
@pbouillet I fixed the CVE issue in master. If you rebase onto the latest version the build error should be gone. I'll take a look at the code now.

pbouillet:
@whiskeysierra done, thanks!

logbook-logstash/src/main/java/org/zalando/logbook/logstash/LogstashLogbackSink.java
```diff
@@ -17,22 +16,32 @@
 
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 
-@AllArgsConstructor
```
whiskeysierra, line 20: I'd leave that in, so you don't have to implement it by hand.

logbook-logstash/src/main/java/org/zalando/logbook/logstash/LogstashLogbackSink.java
```diff
@@ -17,22 +16,32 @@
...
 @API(status = EXPERIMENTAL)
 public final class LogstashLogbackSink implements Sink {
 
     private final static Logger log = LoggerFactory.getLogger(Logbook.class);
 
     private final HttpLogFormatter formatter;
 
+    private String baseField = "http";
```
whiskeysierra, line 26:
```suggestion
    private final String fieldName;
```

logbook-logstash/src/main/java/org/zalando/logbook/logstash/LogstashLogbackSink.java
```diff
@@ -17,22 +16,32 @@
...
+
+    public LogstashLogbackSink(final HttpLogFormatter formatter) {
+        this.formatter = formatter;
```
whiskeysierra, line 29:
```suggestion
        this(formatter, "http");
```

logbook-logstash/src/main/java/org/zalando/logbook/logstash/LogstashLogbackSink.java
```diff
@@ -17,22 +16,32 @@
...
+    }
+
+    public LogstashLogbackSink(final HttpLogFormatter formatter, final String baseField) {
+        this.formatter = formatter;
+        this.baseField = baseField;
+    }
```
whiskeysierra, line 35:
```suggestion
```

logbook-logstash/src/main/java/org/zalando/logbook/logstash/LogstashLogbackSink.java
```diff
@@ -17,22 +16,32 @@
...
+
```
whiskeysierra, line 36:
```suggestion
```

logbook-logstash/src/test/java/org/zalando/logbook/logstash/LogbackLogstashSinkTest.java
```diff
@@ -200,4 +200,101 @@ void shouldLogCorrectJsonForEmptyBody() throws IOException {
         }
 
     }
+
+    @Test
+    void shouldUsePassedBaseField() throws IOException {
```
whiskeysierra, line 205: Is this a full copy of the original test? I'd either make this test way smaller so that it's laser-focused on what it tries to test (just the name of the field) or as an alternative, extract the common parts of both tests into a separate, private method and call that twice with a different parameter:


```java
@Test
void shouldUsePassedBaseField() throws IOException {
    logsAs("test")
}
```

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -245,12 +246,22 @@ public Strategy withoutBody() {
     @API(status = INTERNAL)
     @Bean
     @ConditionalOnMissingBean(Sink.class)
-    public Sink sink(
+    @ConditionalOnProperty(value = "logbook.sink.style", havingValue = "default", matchIfMissing = true)
```
whiskeysierra, line 249: I'd move this to a separate PR. Until now the only way to configure Sinks is programmatically, by registering a bean. I have to think about whether we can (and should?) offer that also for other Sink implementations.

whiskeysierra: (By *this*, I'm referring to the whole Spring Boot Auto Configuration support aspect of this PR.)

logbook-logstash/src/main/java/org/zalando/logbook/logstash/LogstashLogbackSink.java
pbouillet: Fine for me with your suggestion! Thanks

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
pbouillet: Alright, will do

logbook-logstash/src/test/java/org/zalando/logbook/logstash/LogbackLogstashSinkTest.java
pbouillet: I like the private method approach, will go for it

pbouillet:
PR has been updated to only focus the change in the logback sink. Let me know if more needs to be changed!

whiskeysierra:
:+1:

whiskeysierra:
Thank you for your contribution! :tada:
I'm planning on releasing on Monday the latest.

whiskeysierra:
I released it as 2.2.0 just now. Should be available in central soon.

pbouillet:
Awesome, thank you! Let me know if you need the other part as a PR in the future!

PR #846, opened 2020-09-29
Fixes chunking of empty bodies.

### LINKED ISSUE #842: ChunkingSink does not write Requests/Responses without body
## Detailed Description

ChunkingSink should write requests/responses without a body. It should use the delegate sink directly as it only chunks the bodies.

## Your Environment

    * Version used: 2.1.0

## Reproducing
```java
// create a sink which prints to console
var sink = new ChunkingSink(new DefaultSink(new DefaultHttpLogFormatter(), new HttpLogWriter() {

	@Override
	public void write(Precorrelation precorrelation, String request) throws IOException {
		System.out.println(request);

	}


	@Override
	public void write(Correlation correlation, String response) throws IOException {
		System.out.println(response);
	}

}), 100);
// just use whatever does not matter for this test
var corr = new Correlation() {

	@Override
	public String getId() {
		return UUID.randomUUID().toString();
	}


	@Override
	public Instant getStart() {
		return Instant.now();
	}


	@Override
	public Instant getEnd() {
		return Instant.now();
	}


	@Override
	public Duration getDuration() {
		return Duration.ofSeconds(5);
	}

};
// will be printed
sink.write(corr, MockHttpRequest.create().withBodyAsString("bodyAsString"));
sink.write(corr, MockHttpRequest.create().withBodyAsString("bodyAsString"), MockHttpResponse.create().withBodyAsString("bodyAsString"));
// will not be printed
sink.write(corr, MockHttpRequest.create());
sink.write(corr, MockHttpRequest.create(), MockHttpResponse.create());
```

whiskeysierra:
Fixes #842

## Motivation and Context

Fixes #842 

## Types of changes

## Checklist:

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/ChunkingSink.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/ChunkingSink.java
@@ -48,6 +48,10 @@ public final class ChunkingSink implements Sink {
     }
 
     private Stream<String> chunk(final HttpMessage message) throws IOException {
+        if (message.getBodyAsString().isEmpty()) {
+            return Stream.of("");
+        }
+
         return stream(new ChunkingSpliterator(message.getBodyAsString(), minChunkSize, maxChunkSize), false);
     }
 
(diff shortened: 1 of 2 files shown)


whiskeysierra:
👍

Issue #755, opened 2020-05-06
Partial masking


Sam-Kruglov:
Given I have this body
```json
{
  "cardNumber": "1111-2222-3333-4444"
}
```

I want to turn it into this
```json
{
  "cardNumber": "****-****-****-4444"
}
```

I.e. I don't want to mask it completely, but I do want to mask it somehow to hide the sensitive data.

I suggest that `BodyFilters` and `QueryFilters` have this method:
```java

 public static *Filter replaceStringProperty(
    final BiFunction<String, String, String> replacement
) {
    return ...;
}
```

So that I can do this:
```java
*Filters.replaceStringProperty(
    (name, value) -> {
        if (shouldMaskPartiallyPredicate.test(name)){
            return maskPartially(value);
        }
        if (shouldMaskFullyPredicate.test(name)){
            return "***";
        }
        return value;
    };
);
```

whiskeysierra:
Sounds like a useful extension. We have something like that for headers already, if I remember correctly. API looks also good. Do you care to implement it?

Sam-Kruglov:
I probably will but not sure when

Sam-Kruglov:
Switched jobs, now I most likely won't do this one

dnandola-wyze:
Hey @whiskeysierra , this is something I can work on. I need this function especially in Json Body filter to replace body value to hash of that value rather than a generic replacement parameter. Adding a function here to do the replacement would be a helpful extension IMO. Will send out a pull request soon.

whiskeysierra:
Follow up issue is #851

PR #855, opened 2020-10-08
Add feature in JsonBodyFilters to replace value by custom function.

### LINKED ISSUE #851: Provide function in JSON Body filter to obfuscate values rather than replacing completely.
I need a mechanism where I can hash the values in JSON body rather than replace them completely with "XXX". One way of doing that is to provide a function in JSONBodyFilters like:

@API(status = EXPERIMENTAL)
    public static BodyFilter replaceJsonStringProperty(
            final Predicate<String> predicate, final Function<String, String> replacement) {

        return replaceStringFunction(predicate, replacement);
    }

I can submit a PR for this and have the code ready (but unable to push a new branch upstream).

dnandola-wyze:
Summary:
Currently Json Body filters obfusicate the property values with a replacement string. eg. "XXX". However, users may want to do a custom obfusication of these values for eg. Hash of the given value. In order to make it extensible, add a function in JSonBodyFilter to convert a primitive JSON using a custom function provided by the user.

 Description:
The following changes are made:

- Make PrimitiveJsonPropertyBodyFilter as abstract class, where child class will provide replacement function.
- Change Pattern regular expression to add groupname "propertyValue" for grouping the value.
- Add two subclass for string replacement and function replacement.

## Motivation and Context

Adds extension to help users create a custom filter function.
Fixes #851
https://github.com/zalando/logbook/issues/755

## Types of changes

## Checklist:

- [ ] My change requires a change to the documentation. (Not sure about this)

### DIFF
--- a/logbook-json/src/main/java/org/zalando/logbook/json/JsonBodyFilters.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/JsonBodyFilters.java
@@ -6,7 +6,7 @@ import org.zalando.logbook.BodyFilter;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
-import java.util.function.BiFunction;
+import java.util.function.BinaryOperator;
 import java.util.function.Predicate;
 
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
@@ -83,7 +83,7 @@ public final class JsonBodyFilters {
     @API(status = API.Status.EXPERIMENTAL)
     public static BodyFilter replacePrimitiveJsonProperty(
             final Predicate<String> predicate,
-            final BiFunction<String, String, String> replacement) {
+            final BinaryOperator<String> replacement) {
         return replacePrimitiveFunction(predicate, replacement);
     }
 }
--- a/logbook-json/src/main/java/org/zalando/logbook/json/JsonBodyFilters.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/JsonBodyFilters.java
@@ -6,12 +6,14 @@ import org.zalando.logbook.BodyFilter;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.function.BiFunction;
 import java.util.function.Predicate;
 
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 import static org.apiguardian.api.API.Status.MAINTAINED;
 import static org.zalando.logbook.json.PrimitiveJsonPropertyBodyFilter.replaceNumber;
 import static org.zalando.logbook.json.PrimitiveJsonPropertyBodyFilter.replacePrimitive;
+import static org.zalando.logbook.json.PrimitiveJsonPropertyBodyFilter.replacePrimitiveFunction;
 import static org.zalando.logbook.json.PrimitiveJsonPropertyBodyFilter.replaceString;
 
 public final class JsonBodyFilters {
@@ -78,4 +80,10 @@ public final class JsonBodyFilters {
         return replacePrimitive(predicate, replacement);
     }
 
+    @API(status = API.Status.EXPERIMENTAL)
+    public static BodyFilter replacePrimitiveJsonProperty(
+            final Predicate<String> predicate,
+            final BiFunction<String, String, String> replacement) {
+        return replacePrimitiveFunction(predicate, replacement);
+    }
 }
--- a/logbook-json/src/main/java/org/zalando/logbook/json/PrimitiveJsonPropertyBodyFilter.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/PrimitiveJsonPropertyBodyFilter.java
@@ -5,6 +5,8 @@ import lombok.With;
 import org.zalando.logbook.BodyFilter;
 
 import javax.annotation.Nullable;
+import java.util.function.BiFunction;
+import java.util.function.BinaryOperator;
 import java.util.function.Predicate;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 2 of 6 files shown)


whiskeysierra:
I'd prefer an easier change:

- Leave the structure of `PrimitiveJsonPropertyBodyFilter` as it is
- Change `private final String replacement` to `private final BinaryOperator<String> replacement`
- Change all static methods that are calling `new PrimitiveJsonPropertyBodyFilter(..., replacement)` to `new PrimitiveJsonPropertyBodyFilter(..., (name, value) -> replacement)`

whiskeysierra:
In other words, make the current default implementation (string replacement) a special case of the new default implementation (function-based replacement).

dnandola-wyze:
> In other words, make the current default implementation (string replacement) a special case of the new default implementation (function-based replacement).

I actually had implemented it that way initially. The issue with that is it will break the BodyFilter merge functionality. Because it will become a bit complicated to compare replacement function instead:

public boolean compatibleWith(final PrimitiveJsonPropertyBodyFilter that) {	
        return pattern.equals(that.pattern)	
                && replacement.equals(that.replacement);	
    }

whiskeysierra:
That works if you have a non-anonymous function. So instead of my suggestion from above (`(name, value) -> replacement`) you need a real class that overrides equals/hashCode and then it works:

```java
@AllArgsConstructor
private static final class StaticReplacement implements BinaryOperator<String> {

    private final String replacement;

    // TODO apply
    // TODO equals/hashCode
}
```

dnandola-wyze:
Thanks for the feedback. I hadn't thought of that. I have updated the PR with the changes.

logbook-json/src/main/java/org/zalando/logbook/json/PrimitiveJsonPropertyBodyFilter.java
```diff
@@ -98,7 +106,8 @@ public String filter(@Nullable final String contentType, final String body) {
                 if (predicate.test(matcher.group("property"))) {
                     // this preserves whitespaces around properties
                     matcher.appendReplacement(result, "${key}");
-                    result.append(replacement);
+                    result.append(this.replacement.apply(matcher.group("property"),
```
whiskeysierra, line 109:
```suggestion
                    result.append(replacement.apply(matcher.group("property"),
```

logbook-json/src/main/java/org/zalando/logbook/json/StaticParameterReplacementOperator.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.function.BinaryOperator;
+
+import static org.zalando.logbook.json.PrimitiveJsonPropertyBodyFilter.quote;
+
+@AllArgsConstructor
+@EqualsAndHashCode
+public class StaticParameterReplacementOperator<T> implements BinaryOperator<String> {
+    private final T replacement;
+
+    private final boolean addQuotation;
```
whiskeysierra, line 15: If you split this class into two, you'll see that you no longer need this boolean switch. A fully polymorphic construct (three classes then) would be a lot cleaner.

logbook-json/src/main/java/org/zalando/logbook/json/StaticParameterReplacementOperator.java
```diff
@@ -0,0 +1,21 @@
...
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+
+import java.util.function.BinaryOperator;
+
+import static org.zalando.logbook.json.PrimitiveJsonPropertyBodyFilter.quote;
+
+@AllArgsConstructor
+@EqualsAndHashCode
+public class StaticParameterReplacementOperator<T> implements BinaryOperator<String> {
```
whiskeysierra, line 12:
```suggestion
final class StaticParameterReplacementOperator<T> implements BinaryOperator<String> {
```

Always try to limit the surface of your API to the absolute minimum. Gives you freedom to refactor and change stuff later on.

logbook-json/src/main/java/org/zalando/logbook/json/StringFunctionReplacementOperator.java
```diff
@@ -0,0 +1,20 @@
...
+import lombok.EqualsAndHashCode;
+
+import java.util.function.BiFunction;
+import java.util.function.BinaryOperator;
+
+import static org.zalando.logbook.json.PrimitiveJsonPropertyBodyFilter.quote;
+
+@AllArgsConstructor
+@EqualsAndHashCode
+public class StringFunctionReplacementOperator implements BinaryOperator<String> {
```
whiskeysierra, line 13:
```suggestion
final class StringFunctionReplacementOperator implements BinaryOperator<String> {
```

logbook-json/src/main/java/org/zalando/logbook/json/JsonBodyFilters.java
```diff
@@ -78,4 +80,10 @@ public static BodyFilter replacePrimitiveJsonProperty(
         return replacePrimitive(predicate, replacement);
     }
 
+    @API(status = API.Status.EXPERIMENTAL)
+    public static BodyFilter replacePrimitiveJsonProperty(
+            final Predicate<String> predicate,
+            final BiFunction<String, String, String> replacement) {
```
whiskeysierra, line 86: I'd use a `BinaryOperator`here.

logbook-json/src/main/java/org/zalando/logbook/json/PrimitiveJsonPropertyBodyFilter.java
```diff
@@ -63,25 +65,31 @@
...
 
     static BodyFilter replaceNumber(
             final Predicate<String> predicate, final Number replacement) {
-        return create(NUMBER, predicate, String.valueOf(replacement));
+        return create(NUMBER, predicate, new StaticParameterReplacementOperator<>(replacement, false));
     }
 
     static BodyFilter replacePrimitive(
             final Predicate<String> predicate, final String replacement) {
-        return create(PRIMITIVE, predicate, quote(replacement));
```
whiskeysierra, line 84: Interesting. I didn't realize that we always quoted here. Means there is no way to replace booleans and numbers because they will become strings.

whiskeysierra:
Looks good. Let me fix the issue in the master first. Might take a couple of days.

logbook-json/src/main/java/org/zalando/logbook/json/QuotedStringReplacementOperator.java
```diff
@@ -0,0 +1,19 @@
...
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+
+import java.util.function.BinaryOperator;
+
+import static org.zalando.logbook.json.PrimitiveJsonPropertyBodyFilter.quote;
+
+@AllArgsConstructor
+@EqualsAndHashCode
+final class QuotedStringReplacementOperator<T> implements BinaryOperator<String> {
```
whiskeysierra, line 12: There is no need for this class to be generic, is there?

logbook-json/src/main/java/org/zalando/logbook/json/StaticParameterReplacementOperator.java
```diff
@@ -0,0 +1,17 @@
+package org.zalando.logbook.json;
+
+import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
+
+import java.util.function.BinaryOperator;
+
+@AllArgsConstructor
+@EqualsAndHashCode
+final class StaticParameterReplacementOperator<T> implements BinaryOperator<String> {
```
whiskeysierra, line 10: Same here. `T` is always `String`, no need to be generic.

whiskeysierra:
If you rebase onto main, then I can fix the final review comments myself before releasing.

logbook-json/src/main/java/org/zalando/logbook/json/QuotedStringReplacementOperator.java
dnandola-wyze: Yeah. Not in current use cases, I guess. It can be removed if you want.

logbook-json/src/main/java/org/zalando/logbook/json/StaticParameterReplacementOperator.java
dnandola-wyze: Not in this case. It could be number or boolean.

dnandola-wyze:
I rebased it with main.

whiskeysierra:
👍

Issue #740, opened 2020-03-30
Spring boot starter for reactive web environments


jbellmann:
The current [logbook-spring-boot-starter](https://github.com/zalando/logbook/tree/master/logbook-spring-boot-starter) / [logbook-spring-boot-autoconfigure](https://github.com/zalando/logbook/blob/master/logbook-spring-boot-autoconfigure) do not yet support reactive environments using Netty (eg. spring-cloud-gateway).

## Detailed Description
It would be nice if the same developer experience would be supported like currently available for servlet based web environments.

## Possible Implementation
Extend current autoconfigure module for configuration depending on [type of webapplication condition](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/condition/ConditionalOnWebApplication.Type.html).

Create two separate starter (eg. logbook-web-spring-boot-starter, logbook-netty-spring-boot-starter) to manage dependencies.

## Your Environment
* Version used: 2.1.0-SNAPSHOT

I could try to provide a PR, but I would like to disucuss it first (same starter to be used, different starters for clear dependency separation, ...).

jbellmann:
I just realized there was a release 2 days ago. But version in pom.xml in master still shows 2.1.0-SNAPSHOT. I will try to validate this request against latest released version.

whiskeysierra:
> I just realized there was a release 2 days ago. But version in pom.xml in master still shows 2.1.0-SNAPSHOT

#739 is not merged yet. I believe 2.1.0 only includes the netty module, but no specific Spring Boot Webflux support.

whiskeysierra:
@jbellmann Have you solved this internally and could contribute back a webflux auto configuration?

PR #863, opened 2020-10-14
Spring webflux autoconfiguration

sokomishalov:
Related to https://github.com/zalando/logbook/issues/740

### DIFF
--- a/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
+++ b/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
@@ -57,6 +57,7 @@ import org.zalando.logbook.servlet.LogbookFilter;
 import org.zalando.logbook.servlet.SecureLogbookFilter;
 
 import javax.servlet.Filter;
+import javax.servlet.Servlet;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -326,6 +327,10 @@ public class LogbookAutoConfiguration {
 
     @Configuration(proxyBeanMethods = false)
     @ConditionalOnWebApplication(type = Type.SERVLET)
+    @ConditionalOnClass({
+            Servlet.class,
+            LogbookFilter.class
+    })
     static class ServletFilterConfiguration {
 
         private static final String FILTER_NAME = "logbookFilter";
@@ -346,7 +351,7 @@ public class LogbookAutoConfiguration {
                     .withFormRequestMode(properties.getFilter().getFormRequestMode());
             return newFilter(filter, FILTER_NAME, Ordered.LOWEST_PRECEDENCE);
         }
-        
+
         static FilterRegistrationBean newFilter(final Filter filter, final String filterName, final int order) {
             @SuppressWarnings("unchecked") // as of Spring Boot 2.x
             final FilterRegistrationBean registration = new FilterRegistrationBean(filter);
--- a/logbook-bom/pom.xml
+++ b/logbook-bom/pom.xml
@@ -78,6 +78,11 @@
             </dependency>
             <dependency>
                 <groupId>org.zalando</groupId>
+                <artifactId>logbook-spring-boot-webflux-autoconfigure</artifactId>
+                <version>2.5.0-SNAPSHOT</version>
+            </dependency>
+            <dependency>
+                <groupId>org.zalando</groupId>
                 <artifactId>logbook-spring-boot-starter</artifactId>
                 <version>2.5.0-SNAPSHOT</version>
             </dependency>
--- a/logbook-spring-boot-webflux-autoconfigure/pom.xml
+++ b/logbook-spring-boot-webflux-autoconfigure/pom.xml
@@ -1,0 +1,140 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.zalando</groupId>
+        <artifactId>logbook-parent</artifactId>
+        <version>2.5.0-SNAPSHOT</version>
+    </parent>
+
+    <profiles>
+        <profile>
+            <id>spring4</id>
+            <!-- spring-webflux had been introduced only in spring v5 and spring-boot
+            v2 so we have to prevent testing this module against 4-th spring -->
+            <properties>
+                <spring.version>${spring5.version}</spring.version>
+                <spring-boot.version>${spring-boot2.version}</spring-boot.version>
+                <spring-security.version>${spring-security5.version}</spring-security.version>
+            </properties>
+        </profile>
+    </profiles>
+
+    <artifactId>logbook-spring-boot-webflux-autoconfigure</artifactId>
+    <description>Spring Boot Webflux Auto Configuration for Logbook</description>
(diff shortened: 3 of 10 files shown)


whiskeysierra:
First off, thank you for your contribution. It's highly appreciated!

Now, you may have noticed the failing build... It's an indicator to have a slightly more sophisticated setup.
As of today, all Logbook modules are binary compatible to Spring 4 and 5, as well as Spring Boot 1.5 and 2 respectively.
The build has a dedicated profile to switch between the two:

https://github.com/zalando/logbook/blob/d9bb6f6ec56373138f6c416a2be885c8b162e8b9/pom.xml#L606-L613

The problem with Spring WebFlux is that it just didn't exist back in Spring 4.
I'd like to preserve the Spring 4 compatibility of the `logbook-spring-boot-autoconfigure` and `logbook-spring-boot-starter` modules.

We already have a two-phase build:

https://github.com/zalando/logbook/blob/02ad199aab98cd787037b22cdcaa38306e7945f3/.github/workflows/build.yaml#L31-L34

1. We compile against Spring 5/Spring Boot 2
2. We run all tests twice
  1. Once with Spring 5/Spring Boot 2
  2. Once more with Spring 4/Spring Boot 1.5

What's missing from my perspective is:

 - [ ] Conditional classpath, don't put WebFlux/Netty on the classpath when `spring4` profile is active (not sure how to do that)
 - [ ] Conditional tests, they should only run if WebFlux/Netty is on the classpath
 - [ ] Find a way to satisfy code coverage requirements without completely excluding the class (not sure how to do that either)

An alternative to all that would be to introduce an additional module, e.g. `logbook-spring-webflux-autoconfigure`.

sokomishalov:
Oh no, my bad, haven't noticed either failed build or compatibility with spring 4.
Well, it will be quite difficult to make all this stuff be compatible with spring 4.
IMO adding the new module makes sense. And even more I think that it makes sense to separate all spring stuff:
- `logbook-spring-boot-autoconfigure` - for core `logbook` stuff like sinks, formatters, strategies, etc.
- `logbook-spring-boot-webmvc-autoconfigure` - for servlet `logbook`(filters, secure filters)
- `logbook-spring-boot-webflux-autoconfigure` - for webflux `logbook`(netty customizers)
- and maybe wrap them all with parent module `logbook-spring`

Separating modules like this would prevent mixing and polluting classpaths of "spring" users.

whiskeysierra:
> Separating modules like this would prevent mixing and polluting classpaths of "spring" users.

Technically all dependencies of autoconfigure modules should be optional anyway, so pollution shouldn't be an issue.
I do like the separation, but it would be a breaking change.

whiskeysierra:
Maybe we could add the separate webflux autoconfigure module now, but already outline the vision and next steps to ultimately end up with 3+ autoconfigure modules. On the road towards Logbook 3.x then.

sokomishalov:
Fair enough, I agree with that
Will move to webflux module soon

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -377,4 +383,31 @@ private static FilterRegistrationBean newFilter(final Filter filter, final Strin
...
+    static class WebFluxServerConfiguration {
+
+        static final String CUSTOMIZER_NAME = "logbookNettyServerCustomizer";
+
+        @Bean
+        @ConditionalOnProperty(name = "logbook.filter.enabled", havingValue = "true", matchIfMissing = true)
+        @ConditionalOnMissingBean(name = CUSTOMIZER_NAME)
+        public NettyServerCustomizer logbookNettyServerCustomizer(final Logbook logbook) {
+            return httpServer -> httpServer.tcpConfiguration(tcpServer -> tcpServer.doOnConnection(connection -> connection.addHandlerLast(new LogbookServerHandler(logbook))));
+        }
```
sergey-volkov-lm, line 398: Seems like this does not work nicely together with spring-cloud-starter-sleuth (2.2.5), specifically traceId & spanId MDC variables are empty in logbook's request & response lines.
Being new to webflux, I don't have any ideas how to fix this, but thought it's probably worth sharing 🤔

whiskeysierra: > MDC variables

MDC is just a fancy name for a thread-local map. ThreadLocal in general doesn't work with reactive systems (Netty, Reactor, Rx, etc) because their approach to multi-threading does not guarantee a that the same thread is being used.

- https://projectreactor.io/docs/core/release/reference/#faq.mdc
- https://www.novatec-gmbh.de/en/blog/how-can-the-mdc-context-be-used-in-the-reactive-spring-applications/
- https://github.com/reactor/reactor-core/issues/1985

It's an issue that is not directly related to Logbook, but more to the very nature of reactive systems. 
I have to admit, that I'm not fully in the topic, so if someone has productive suggestions and contributions to Logbook that help with that in any way, I'm very much interested in that.

whiskeysierra: @sergey-volkov-lm Can you take this discussion to a separate issue, if you're still interested in that? It's not too strictly related to this PR.

sergey-volkov-lm: @whiskeysierra about sleuth & MDC: it claims support for reactive applications, seems to work.
I suppose that in order to integrate with sleuth (and spring reactive ecosystem in general) one have to implement WebFilter, so this relates to spring-auto-configuration. Would you like me to move this thread into original issue #740 or create new one?

whiskeysierra:
Do you have some links/references?

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
sergey-volkov-lm: > Do you have some links/references?

(docs) [sleuth integrations: reactor](https://docs.spring.io/spring-cloud-sleuth/docs/current-SNAPSHOT/reference/html/integrations.html#sleuth-reactor-integration)
(source code) [Sleuth's TraceWebFilter](https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebFilter.java)
(stackoverflow) [question about logging request/response bodies answered by Spring member](https://stackoverflow.com/a/45280764/6629674)

whiskeysierra: If I'm reading those correctly, then you can actually fix those outside of Logbook by augmenting/configuring sleuth or their filter differently. I'd prefer to create a completely new issue for this.

sokomishalov:
Sorry, haven't fixed this pr for a very long time.
image

I've moved webflux autoconfiguration to a separate module. 
Module `spring-boot-autoconfigure` is a compile dependency with some exclusions for `spring-boot-webflux-autoconfigure` to inherit all needed stuff.
I propose to merge these modules into one in future (3.x release) and drop compatibility for the 4-th spring.

whiskeysierra:
👍

whiskeysierra:
Thank you very much for your contribution! 📣 

I'm preparing a new release right now and it will be 2.5.0.

magnusheino:
2.5.0 was never released...? Are there snapshots available somewhere?

whiskeysierra:
Sorry, the release failed because sonatype had some issues. Retrying right now.

whiskeysierra:
Done, 2.5.0 should be available in central soon. Sorry for the delay.

magnusheino:
Works perfectly, thanks for the help!

PR #878, opened 2020-11-02
Externalize a handler to write json fields

rjmveloso:
Allows to contribute with a custom JsonFieldWriter to have full freedom on how to write headers and body (naming and ordering).

### DIFF
--- a/logbook-json/src/main/java/org/zalando/logbook/json/FastJsonHttpLogFormatter.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/FastJsonHttpLogFormatter.java
@@ -1,9 +1,12 @@
 package org.zalando.logbook.json;
 
-import com.fasterxml.jackson.core.JsonFactory;
-import com.fasterxml.jackson.core.JsonGenerator;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import lombok.AllArgsConstructor;
+import static org.apiguardian.api.API.Status.STABLE;
+
+import java.io.IOException;
+import java.io.StringWriter;
+import java.util.List;
+import java.util.Map;
+
 import org.apiguardian.api.API;
 import org.zalando.logbook.Correlation;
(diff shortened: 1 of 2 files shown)


logbook-json/src/main/java/org/zalando/logbook/json/FastJsonHttpLogFormatter.java
```diff
@@ -12,24 +15,23 @@
...
  */
 @API(status = STABLE)
-@AllArgsConstructor
-public final class FastJsonHttpLogFormatter implements HttpLogFormatter {
+@RequiredArgsConstructor
+public final class FastJsonHttpLogFormatter implements HttpLogFormatter, JsonFieldWriter {
 
     private final JsonFactory factory;
 
+    private JsonFieldWriter delegate = this; // default to this implementation
```
whiskeysierra, line 33: By implementing this interface directly you're exposing that implementation detail to the outside world because `FastJsonHttpLogFormatter` is public. I'd hide that with a static inner class.

logbook-json/src/main/java/org/zalando/logbook/json/FastJsonHttpLogFormatter.java
```diff
@@ -38,59 +40,31 @@ public FastJsonHttpLogFormatter(final ObjectMapper mapper) {
         this(mapper.getFactory());
     }
 
-    @Override
-    public String format(
-            final Precorrelation precorrelation,
-            final HttpRequest request) throws IOException {
+    public FastJsonHttpLogFormatter(final ObjectMapper mapper,
```
whiskeysierra, line 43: I guess this constructor is not covered by tests. If you follow my approach from above you get that for free, because you would actually use it then.

whiskeysierra:
👍

whiskeysierra:
I'm releasing this as 2.4.0 as we speak. Thanks for the contribution! 🎉

Issue #881, opened 2020-11-04
ClassNotFoundException: javax.servlet.Filter thrown during Spring Reactive Web application startup


jstanik:
During startup of a Spring Reactive Web application a `java.lang.ClassNotFoundException: javax.servlet.Filter` is thrown and the application cannot start.

## Description

The `ClassNotFoundException` is thrown when Spring Boot processes `LogbookAutoConfiguration` class. Even though there is a static inner class `ServletFilterConfiguration` which is processed conditionally if the `javax.servlet.Filter` class is present on the classpath there is one static method directly in the `LogbookAutoConfiguration` class

```
private static FilterRegistrationBean newFilter(final Filter filter, final String filterName, final int order) 
```
which references the `Filter`. This `newFilter(...)` was probably meant to be part of the static inner class `ServletFilterConfiguration` and not `LogbookAutoConfiguration`.

## Expected Behavior
Application should start and `LogbookAutoConfiguration` should be processed without causing `ClassNotFoundExcetpion`.

## Actual Behavior
When Spring Boot processes the `LogbookAutoConfiguration` then `ClassNotFoundException` is thrown:

```
2020-11-04 14:15:36.438 ERROR 1136 --- [           main] o.s.boot.SpringApplication               : Application run failed

java.lang.IllegalStateException: Error processing condition on org.zalando.logbook.autoconfigure.LogbookAutoConfiguration.correlationId
	at org.springframework.boot.autoconfigure.condition.SpringBootCondition.matches(SpringBootCondition.java:60) ~[spring-boot-autoconfigure-2.3.5.RELEASE.jar:2.3.5.RELEASE]
	at org.springframework.context.annotation.ConditionEvaluator.shouldSkip(ConditionEvaluator.java:108) ~[spring-context-5.2.10.RELEASE.jar:5.2.10.RELEASE]
	at org.springframework.context.annotation.ConfigurationClassBeanDefinitionReader.loadBeanDefinitionsForBeanMethod(ConfigurationClassBeanDefinitionReader.java:184) ~[spring-context-5.2.10.RELEASE.jar:5.2.10.RELEASE]
	at org.springframework.context.annotation.ConfigurationClassBeanDefinitionReader.loadBeanDefinitionsForConfigurationClass(ConfigurationClassBeanDefinitionReader.java:144) ~[spring-context-5.2.10.RELEASE.jar:5.2.10.RELEASE]
[...]

PR #882, opened 2020-11-04
Fix ClassNotFoundException when LogbookAutoConfiguration is processed

jstanik:
To avoid `ClassNotFoundException: javax.servlet.Filter` a code referencing the `Filter` is moved to a static inner class which is processed only conditionally when `javax.servlet.Filter` is on the classpath.

## Description

Factory method for creating `FilterRegistrationBean` moved from `LogbookAutoConfiguration` to conditionally processed class `ServletFilterConfiguration`.

## Motivation and Context
Fixes issue https://github.com/zalando/logbook/issues/881

## Types of changes

## Checklist:

### DIFF
--- a/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
+++ b/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
@@ -345,11 +345,20 @@ public class LogbookAutoConfiguration {
                     .withFormRequestMode(properties.getFilter().getFormRequestMode());
             return newFilter(filter, FILTER_NAME, Ordered.LOWEST_PRECEDENCE);
         }
+        
+        private static FilterRegistrationBean newFilter(final Filter filter, final String filterName, final int order) {
+            @SuppressWarnings("unchecked") // as of Spring Boot 2.x
+            final FilterRegistrationBean registration = new FilterRegistrationBean(filter);
+            registration.setName(filterName);
+            registration.setDispatcherTypes(REQUEST, ASYNC);
+            registration.setOrder(order);
+            return registration;
+        }
 
     }
 
     @Configuration(proxyBeanMethods = false)
-    @ConditionalOnClass(SecurityFilterChain.class)
+    @ConditionalOnClass({ SecurityFilterChain.class, Filter.class })
     @ConditionalOnWebApplication
     @AutoConfigureAfter(name = {
             "org.springframework.boot.autoconfigure.security.SecurityFilterAutoConfiguration", // Spring Boot 1.x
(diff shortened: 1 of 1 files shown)


logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -345,11 +345,20 @@ public FilterRegistrationBean logbookFilter(final Logbook logbook) {
...
+            registration.setOrder(order);
+            return registration;
+        }
 
     }
 
     @Configuration(proxyBeanMethods = false)
-    @ConditionalOnClass(SecurityFilterChain.class)
+    @ConditionalOnClass({ SecurityFilterChain.class, Filter.class })
     @ConditionalOnWebApplication
```
sergey-volkov-lm, line 362: Maybe `@ConditionalOnWebApplication(type = SERVLET)` to be more explicit, like it's done in https://github.com/zalando/logbook/pull/863

whiskeysierra: Yes, that's a good idea.

jstanik: I have added the web application type `SERVLET` to  both the `SecurityServletFilterConfiguration` and `ServletFilterConfiguration`.

Now I am not sure if keeping the condition `@ConditionalOnClass(Filter.class)` makes sense for these.

whiskeysierra:
I believe it's redundant now. Should be fine to remove those.

jstanik:
Ok, I removed those redundant conditions.

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -345,12 +345,21 @@ public FilterRegistrationBean logbookFilter(final Logbook logbook) {
...
+            registration.setOrder(order);
+            return registration;
+        }
 
     }
 
     @Configuration(proxyBeanMethods = false)
-    @ConditionalOnClass(SecurityFilterChain.class)
-    @ConditionalOnWebApplication
+    @ConditionalOnClass({ SecurityFilterChain.class })
```
whiskeysierra, line 361: Why the curly braces here?

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -363,18 +372,7 @@ public FilterRegistrationBean logbookFilter(final Logbook logbook) {
         @ConditionalOnProperty(name = "logbook.secure-filter.enabled", havingValue = "true", matchIfMissing = true)
         @ConditionalOnMissingBean(name = FILTER_NAME)
         public FilterRegistrationBean secureLogbookFilter(final Logbook logbook) {
-            return newFilter(new SecureLogbookFilter(logbook), FILTER_NAME, Ordered.HIGHEST_PRECEDENCE + 1);
+            return ServletFilterConfiguration.newFilter(new SecureLogbookFilter(logbook), FILTER_NAME, Ordered.HIGHEST_PRECEDENCE + 1);
```
whiskeysierra, line 375: A static import would not make it a bit shorter (same as it was in fact).

whiskeysierra:
👍

whiskeysierra:
I'm currently releasing this fix as 2.4.1. Thank you for your contribution! 🎉

PR #888, opened 2020-11-10
New module to handler with SOAP logs

jelas99:
## Description
New module to handler with SOAP logs

## Motivation and Context
This change allows to add a soap handler to log the request/responses

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.

PR #895, opened 2020-11-25
Preliminary JAX-WS module

whiskeysierra:
## Motivation and Context

Fixes #888 

## Types of changes

## Checklist:


logbook-jaxws/src/main/java/org/zalando/logbook/jaxws/Handler.java
```diff
@@ -0,0 +1,71 @@
...
+        memorize(stage);
+    }
+
+    void handleResponse(final Origin origin) {
+        final ResponseProcessingStage stage = recall();
+        write(stage, new Response(new HttpMessageContext(context), origin));
+    }
+
+    private void memorize(final ResponseProcessingStage stage) {
+        context.put(ATTRIBUTE, stage);
```
rjmveloso, line 45: I have seen a similar strategy across the library, but this expose the PresponseProcessingStage under context/request.
Using an alternative of ThreadLocal allows to hide any library concept, avoiding get the reference of ResponseProcessingStage in any place of an application with access to javax.xml.ws.handler.soap.SOAPMessageContext/javax.servlet.http.HttpServletRequest.

whiskeysierra: Yes, technically it's being exposed, but those facilities exist exactly for our use case, to store context information between requests and responses. ThreadLocal is a smell, that I'm trying to avoid, because I'd rather delegate the job of keeping track of the context to the framework, then imposing  the requirement to have the same thread process the request and response (which is not ensured).

whiskeysierra:
I took another look and it turns out that there is no viable way to extract the necessary information on the client side, afaict.

Regarding the server side, we already have support for Servlets. I'd suggest to use that, if a servlet-based JAX-WS implementation is being used.

Issue #915, opened 2020-12-25
Separate host/path from uri 


nhomble:
```StructuredHttpLogFormatter``` sets ```content.put("uri", request.getRequestUri())``` and I think this should be broken down.

## Context

When I am searching through logs in some log monitor like splunk, I find it helpful to separate larger strings like this for simpler searches. If I am doing searches across nonprod environments, the host may be different so instead of doing
```uri=*/some/path```
I can just do
```path=/some/path```

Also in cases of client side load balancing, I might be interested to see how we are distributed across hosts.
```search path=/some/api | stats count by host```

## Possible Implementation
I suggest we separate the two into different k/v pairs. 
```java
content.put("host", request.getHost());
content.put("path", request.getPath());
```
For backwards compatibility, we could keep the redundant ```uri``` field.

## Your Environment

No bug, just writing a spring boot web service. We can always write a custom implementation and extend ```prepare()``` if needed.

nhomble:
@whiskeysierra for the query param parsing, is there a preference for picking some third party lib into `logbook-api` or do we want to rely on the underlying http lib in each sibling module? 

If I had to imagine the rationale for your original reasoning for keeping the `QueryFilter` based off of a `String`, library consumers can always implement their own filter and utilize their http lib to parse the query string. Similarly, library consumers could implement `StructuredHttpLogFormatter` and add query params.

whiskeysierra:
Multiple reasons, actually:
- there is no official RFC that actually specifies query parameters
- Logbook's primary purpose is to log requests "as-is", with as little transformation as possible
- minimal dependencies
- no desire to implement **and maintain** a custom implementation

nhomble:
Yeah that makes sense to me. Then I think we agree that supporting query params as suggested in the pr won't be implemented? 

Ppl who need it can implement the interfaces as I mentioned above

whiskeysierra:
> Then I think we agree that supporting query params as suggested in the pr won't be implemented?

Yes 👍

PR #916, opened 2020-12-25
adds host and path to structured log formats

nhomble:
Proposed change for https://github.com/zalando/logbook/issues/915

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,7 +303,11 @@ Requires the following dependency:
   "protocol": "HTTP/1.1",
   "sender": "127.0.0.1",
   "method": "GET",
-  "path": "http://example.org/test",
+  "uri": "http://example.org/test",
+  "host": "example.org",
+  "path": "/test",
+  "scheme": "http",
+  "port": "NULL",
   "headers": {
     "Accept": ["application/json"],
     "Content-Type": ["text/plain"]
@@ -385,7 +389,7 @@ requests and response as key-value pairs.
 ###### Request
 
 ```text
-origin=remote type=request correlation=2d66e4bc-9a0d-11e5-a84c-1f39510f0d6b protocol=HTTP/1.1 sender=127.0.0.1 method=POST path=http://example.org/test headers={Accept=[application/json], Content-Type=[text/plain]} body=Hello world!
+origin=remote type=request correlation=2d66e4bc-9a0d-11e5-a84c-1f39510f0d6b protocol=HTTP/1.1 sender=127.0.0.1 method=POST uri=http://example.org/test host=example.org scheme=http port=NULL path=/test headers={Accept=[application/json], Content-Type=[text/plain]} body=Hello world!
 ```
 
 ###### Response
--- a/README.md
+++ b/README.md
@@ -307,7 +307,7 @@ Requires the following dependency:
   "host": "example.org",
   "path": "/test",
   "scheme": "http",
-  "port": "NULL",
+  "port": null,
   "headers": {
     "Accept": ["application/json"],
     "Content-Type": ["text/plain"]
@@ -389,7 +389,8 @@ requests and response as key-value pairs.
 ###### Request
 
 ```text
-origin=remote type=request correlation=2d66e4bc-9a0d-11e5-a84c-1f39510f0d6b protocol=HTTP/1.1 sender=127.0.0.1 method=POST uri=http://example.org/test host=example.org scheme=http port=NULL path=/test headers={Accept=[application/json], Content-Type=[text/plain]} body=Hello world!
+origin=remote type=request correlation=2d66e4bc-9a0d-11e5-a84c-1f39510f0d6b protocol=HTTP/1.1 sender=127.0.0.1 method=POST uri=http://example.org/test host=example.org scheme=http port=null path=/test headers={Accept=[application/json], Content-Type=[text/plain]} body=Hello world!
+
 ```
 
 ###### Response
--- a/logbook-api/src/main/java/org/zalando/logbook/StructuredHttpLogFormatter.java
+++ b/logbook-api/src/main/java/org/zalando/logbook/StructuredHttpLogFormatter.java
@@ -3,6 +3,7 @@ package org.zalando.logbook;
 import org.apiguardian.api.API;
 
 import java.io.IOException;
+import java.net.URLEncoder;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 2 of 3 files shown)


whiskeysierra:
Thanks for your pull request! 📣 

Looks good so far. Just one thing to add:
This change affects both, the `JsonHttpLogFormatter` and the `SplunkHttpLogFormatter `. The docs for both should be adjusted accordingly:

- https://github.com/zalando/logbook#json
- https://github.com/zalando/logbook#splunk

Another idea, while we're at it: What about scheme, port and query? If we break down the request uri into its components, why not apply it all the way?

nhomble:
Sorry for missing the docs, I'll definitely update. 

I also agree that it makes sense to add all of the components. One question, should we be consistent with the `prepareHeaders`/`prepareBody` pattern and ignore `port` when optional is empty and ignore `query` when query is empty?

whiskeysierra:
Good question. I already thought about it. I guess absent would be more consistent with the existing features. But null values or defaults could be easier to use on the log analytics site of things.

nhomble:
I think it would be reasonable to put `null` as a default port. And then if users want to remap to some numeric default they can do that in their query. 
I'm also thinking for query, it would be nice to parse the map out of the string and insert into content similar to headers. Beyond making searching easier, I think if we make this a utility method of `HttpRequest` then the `QueryFilter` implementation could become easier.

nhomble:
Since the other keys are pretty straightforward, I'm thinking we tackle the query params separate from the other simpler fields (happy to follow up with another pr). I was also a little confused by the readme since it mentioned `path` in both the json and splunk examples. If there is another transformation happening somewhere, I could use help seeing where that happens.

whiskeysierra:
> I'm also thinking for query, it would be nice to parse the map out of the string and insert into content similar to headers. Beyond making searching easier, I think if we make this a utility method of `HttpRequest` then the `QueryFilter` implementation could become easier.

Yes, that would be nice.

> Since the other keys are pretty straightforward, I'm thinking we tackle the query params separate from the other simpler fields (happy to follow up with another pr).

Sounds good.

> I was also a little confused by the readme since it mentioned `path` in both the json and splunk examples. If there is another transformation happening somewhere, I could use help seeing where that happens.

I guess it's just out of date with the implementation. Maybe I had this in mind at some point but never implemented it.

README.md
```diff
@@ -303,7 +303,11 @@ Requires the following dependency:
   "protocol": "HTTP/1.1",
   "sender": "127.0.0.1",
   "method": "GET",
-  "path": "http://example.org/test",
+  "uri": "http://example.org/test",
+  "host": "example.org",
+  "path": "/test",
+  "scheme": "http",
+  "port": "NULL",
```
whiskeysierra, line 310: This should be a proper JSON null value.

```suggestion
  "port": null,

```

README.md
```diff
@@ -385,7 +389,7 @@ requests and response as key-value pairs.
 ###### Request
 
 ```text
-origin=remote type=request correlation=2d66e4bc-9a0d-11e5-a84c-1f39510f0d6b protocol=HTTP/1.1 sender=127.0.0.1 method=POST path=http://example.org/test headers={Accept=[application/json], Content-Type=[text/plain]} body=Hello world!
+origin=remote type=request correlation=2d66e4bc-9a0d-11e5-a84c-1f39510f0d6b protocol=HTTP/1.1 sender=127.0.0.1 method=POST uri=http://example.org/test host=example.org scheme=http port=NULL path=/test headers={Accept=[application/json], Content-Type=[text/plain]} body=Hello world!
```
whiskeysierra, line 392:
```suggestion
origin=remote type=request correlation=2d66e4bc-9a0d-11e5-a84c-1f39510f0d6b protocol=HTTP/1.1 sender=127.0.0.1 method=POST uri=http://example.org/test host=example.org scheme=http port=null path=/test headers={Accept=[application/json], Content-Type=[text/plain]} body=Hello world!

```

logbook-api/src/main/java/org/zalando/logbook/StructuredHttpLogFormatter.java
```diff
@@ -92,6 +97,12 @@ default String format(final Correlation correlation, final HttpResponse response
         return content;
     }
 
+    default String preparePort(final HttpRequest request){
+        return request.getPort()
+                .map(Object::toString)
+                .orElse("NULL");
+    }
```
whiskeysierra, line 104:
```suggestion
    @Nullable
    default String preparePort(final HttpRequest request){
        return request.getPort()
                .map(Object::toString)
                .orElse(null);
    }
```

logbook-api/src/test/java/org/zalando/logbook/StructuredHttpLogFormatterTest.java
```diff
@@ -89,9 +90,23 @@ void prepareRequest() throws IOException {
                 .containsEntry("method", "GET")
                 .containsEntry("uri", "https://www.example.org/search?q=example")
                 .containsEntry("headers", singletonMap("Test", emptyList()))
+                .containsEntry("host", "www.example.org")
+                .containsEntry("path", "/search")
+                .containsEntry("scheme", "https")
+                .containsEntry("port", "NULL")
```
whiskeysierra, line 96:
```suggestion
                .containsEntry("port", null)
```

README.md
```diff
@@ -385,7 +389,8 @@ requests and response as key-value pairs.
 ###### Request
 
 ```text
-origin=remote type=request correlation=2d66e4bc-9a0d-11e5-a84c-1f39510f0d6b protocol=HTTP/1.1 sender=127.0.0.1 method=POST path=http://example.org/test headers={Accept=[application/json], Content-Type=[text/plain]} body=Hello world!
+origin=remote type=request correlation=2d66e4bc-9a0d-11e5-a84c-1f39510f0d6b protocol=HTTP/1.1 sender=127.0.0.1 method=POST uri=http://example.org/test host=example.org scheme=http port=null path=/test headers={Accept=[application/json], Content-Type=[text/plain]} body=Hello world!
```
whiskeysierra, line 392: I'm not an expert on splunk - maybe an empty value would be cleaner?

whiskeysierra: Looks like it https://community.splunk.com/t5/Splunk-Search/What-is-the-best-way-to-search-for-blank-null-fields-in-a-search/m-p/211083

whiskeysierra:
👍

whiskeysierra:
Thanks for your contribution! 🎉 

I'll wait for your other PRs to make one big release, unless you need it faster.

nhomble:
@whiskeysierra thanks! I think I should have bandwidth to finish our open prs and propose something for the query params.

PR #917, opened 2020-12-27
adds LogbookClientHttpRequestInterceptor for spring RestTemplate

nhomble:
## Description

Creates a new `ClientHttpRequestInterceptor` that composes existing apache interceptor functionality but is more pluggable for spring users.

## Motivation and Context

Spring `RestTemplate` users are typically abstracted away from `HttpClient` specific configurations. We operate at the `RestTemplateBuilder` level. 

## Types of changes

## Checklist:


logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -6,11 +6,7 @@
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
```
nhomble, line 9: sorry intellij must have autoformatted, I'll fix

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/interceptors/LogbookClientHttpRequestInterceptor.java
```diff
@@ -0,0 +1,67 @@
...
+import org.zalando.logbook.httpclient.LogbookHttpResponseInterceptor;
+
+import java.io.IOException;
+
+/**
+ * Uses {@link LogbookHttpResponseInterceptor} and {@link LogbookHttpResponseInterceptor} to log the start and
+ * completion of client http requests.
+ */
+@API(status = API.Status.EXPERIMENTAL)
+public class LogbookClientHttpRequestInterceptor implements ClientHttpRequestInterceptor {
```
whiskeysierra, line 23: This implementation is now has a compile-time dependency to apache http-client, even if someone is using the RestTemplate with another HTTP client implementation.

A cleaner way would be to implement it completely separate and also to move it to its own module, e.g. `logbook-spring`.

whiskeysierra: (For those users that are just using Spring, but not Spring Boot)

logbook-spring/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.zalando</groupId>
+            <artifactId>logbook-core</artifactId>
+        </dependency>
+        <dependency>
```
whiskeysierra, line 40: This module and the implementation still depends on the Apache HTTP Client implementation now.

nhomble: maybe it makes less sense for me to reuse `LogbookHttpRequestInterceptor` and `LogbookHttpResponseInterceptor` and instead just adapt spring's `HttpRequest` and `HttpResponse` into the logbook equivalents.

logbook-spring/src/main/java/org/zalando/logbook/spring/LogbookClientHttpRequestInterceptor.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.http.client.ClientHttpRequestExecution;
+import org.springframework.http.client.ClientHttpRequestInterceptor;
+import org.springframework.http.client.ClientHttpResponse;
+import org.zalando.logbook.HttpResponse;
+import org.zalando.logbook.Logbook;
+
+import java.io.IOException;
+
+@API(status = API.Status.EXPERIMENTAL)
+public class LogbookClientHttpRequestInterceptor implements ClientHttpRequestInterceptor {
```
whiskeysierra, line 14:
```suggestion
public final class LogbookClientHttpRequestInterceptor implements ClientHttpRequestInterceptor {
```

logbook-spring/src/main/java/org/zalando/logbook/spring/LogbookClientHttpRequestInterceptor.java
```diff
@@ -0,0 +1,34 @@
...
+
+    private final Logbook logbook;
+
+    public LogbookClientHttpRequestInterceptor(Logbook logbook) {
```
whiskeysierra, line 18: Can be generated by lombok

README.md
```diff
@@ -630,6 +630,7 @@ Logbook comes with a convenient auto configuration for Spring Boot users. It set
 - Header-/Parameter-/Body-Filters
 - HTTP-/JSON-style formatter
 - Logging writer
+- A `RestTemplate`-friendly `ClientHttpRequestInterceptor`
```
whiskeysierra, line 633: I'd say the new module deserves a separate section in the README, similar to the other integrations.

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -323,6 +324,11 @@ public LogbookHttpResponseInterceptor logbookHttpResponseInterceptor() {
             return new LogbookHttpResponseInterceptor();
         }
 
+        @Bean
+        @ConditionalOnMissingBean(LogbookClientHttpRequestInterceptor.class)
```
whiskeysierra, line 328: This should be mutually exclusive to the Apache HTTP Client setup (two bean methods above), shouldn't it?

Otherwise users will see requests and responses twice in the logs.

logbook-spring/src/main/java/org/zalando/logbook/spring/ByteStreams.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    static byte[] toByteArray(final InputStream in) throws IOException {
+        final ByteArrayOutputStream out = new ByteArrayOutputStream();
+        copy(in, out);
+        return out.toByteArray();
+    }
+
+    static void copy(final InputStream from, final OutputStream to) throws IOException {
+        Objects.requireNonNull(from);
```
whiskeysierra, line 22: I'd leave null checks out. Callers will see a NPE anyway.

logbook-spring/src/main/java/org/zalando/logbook/spring/LocalRequest.java
```diff
@@ -0,0 +1,119 @@
...
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+final class LocalRequest implements HttpRequest {
+
+    private final org.springframework.http.HttpRequest request;
+    private final byte[] body;
+
+    private boolean withBody = false;
+
+    LocalRequest(org.springframework.http.HttpRequest request, byte[] body) {
```
whiskeysierra, line 23: Same as `@RequiredArgsConstructor`.

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -304,6 +304,12 @@ public HttpLogWriter writer() {
         return new DefaultHttpLogWriter();
     }
 
+    @Bean
+    @ConditionalOnMissingBean(LogbookClientHttpRequestInterceptor.class)
+    public LogbookClientHttpRequestInterceptor logbookClientHttpRequestInterceptor(Logbook logbook) {
```
whiskeysierra, line 309: How would a usual integration look like on the user side? Inject this by hand and call a `RestTemplateBuilder`?

nhomble: With the builder, should look like

```java
private final RestTemplate rt;
MyClient(RestTemplateBuilder builder, LogbookClientHttpRequestInterceptor interceptor){
  this.rt = builder
    .additionalInterceptors(interceptor)
    .build();
}
```

whiskeysierra: Should we add that to the README as an example?

nhomble: that makes sense but it'll go under a `spring-boot` section since that's where `RestTemplateBuilder` comes from. I didn't make that assumption for ppl pulling the spring dependency directly

whiskeysierra:
👍

PR #948, opened 2021-02-08
Fix empty response body when using Spring RestTemplate with LogbookClientHttpRequestInterceptor

thowimmer:
This PR fixes that null response bodies are returned for non-empty responses when using the new LogbookClientHttpRequestInterceptor with a Spring RestTemplate.

## Description
The response extractor of the Spring RestTemplate considers the position of the response body stream.
This leads to a null response body, since the buffered RemoteResponse was not resetting the streams position after copying the response body bytes.

## Motivation and Context
Currently RestTemplates without any special configuration are returning null response entities when using the new LogbookClientHttpRequestInterceptor.

## Types of changes

## Checklist:

### DIFF
--- a/logbook-spring/src/main/java/org/zalando/logbook/spring/RemoteResponse.java
+++ b/logbook-spring/src/main/java/org/zalando/logbook/spring/RemoteResponse.java
@@ -9,6 +9,7 @@ import org.zalando.logbook.Origin;
 
 import javax.annotation.Nullable;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.charset.Charset;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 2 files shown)


whiskeysierra:
👍

whiskeysierra:
Thank you for your contribution! 🎉 
I'll release it asap.

thowimmer:
Cool thanks @whiskeysierra :-) 


### For you notice

This few lines of code changes how RestTemplate behaves when deserializing the response body in the end.

I had to change the response type of the first test to String so that it **passes** with my changes !
I'm unsure whether the intended behavior when calling _getForObject(url, Void.class)_ is that it fails in case the server returns a response body of a certain content type.

Even though I think this behavior is the correct one, it might break some clients which have different assumptions in this respect - This might be a very limited side effect for clients which want to do the following thing:
Call getForObject(...) on an endpoint which returns a certain content type but do not want to interpret it further.

whiskeysierra:
Understood. I'd rather fix the obvious issue here rather than staying compatible with a rather unusual behavior. Totally agree with your assessment there.

whiskeysierra:
I released your fix as 2.4.2 to central.

Issue #954, opened 2021-02-10
Empty or truncated servlet http response on async request  


nobo728x:
servlet http reponse is empty or truncated

## Description
The bug is in org.zalando.logbook.servlet.LogbookFilter that assumes that if request.getDispatcherType() == ASYNC implies request.isAsyncStarted() but it's not the case, because the code after the filter (graphql servlet in my case) can start the async context and because of this the reponse is never flushed.

## Expected Behavior
Http response as writed by the servlet 

## Actual Behavior
empty http reponse

## Possible Fix
I will try the fix it 

## Steps to Reproduce
I wrote a unit test shouldResponseNotEmptyAsync (whitch fails) 
https://github.com/nobo728x/logbook/blob/main/logbook-servlet/src/test/java/org/zalando/logbook/servlet/AsyncDispatchTest.java#L114

## Context
spring-boot + graphql-java application 

## Your Environment
* Version 2.5.0 (2.4.X are affected)

PascalSchumacher:
I guess this issue should be closed, as it was fixed with #954 and released with `2.6.2`?

PR #985, opened 2021-03-23
capture response where servlet code programmatically starts async dispatch

noffke:
## Description
The GraphQL servlet from https://github.com/graphql-java-kickstart/graphql-java-servlet programmatically starts an async request by calling `startAsync` on the `HttpServletRequest`. In this case, logbook misses the reply because the logbook filter checks `request.isAsyncStarted()` before logging. In order to fix this issue, an `AsyncListener` is being registered whenever `startAsync` is being called.

Regarding `LogbookAsyncListener`, would it make sense to also implement the other methods besides `onComplete`? If yes, how?

## Motivation and Context
Fix for issue #954

## Types of changes

## Checklist:

### DIFF
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LogbookFilter.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LogbookFilter.java
@@ -1,5 +1,14 @@
 package org.zalando.logbook.servlet;
 
+import java.io.IOException;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicBoolean;
+import javax.annotation.Nullable;
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
 import lombok.AllArgsConstructor;
 import lombok.With;
 import org.apiguardian.api.API;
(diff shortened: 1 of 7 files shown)


whiskeysierra:
First off, thank you for your contribution! 📣 

> Regarding `LogbookAsyncListener`, would it make sense to also implement the other methods besides `onComplete`? If yes, how?

I can't really judge, tbh.

logbook-servlet/pom.xml
```diff
@@ -108,4 +108,18 @@
...
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.jacoco</groupId>
+                <artifactId>jacoco-maven-plugin</artifactId>
+                <configuration>
+                    <excludes>
+                        <exclude>org/zalando/logbook/servlet/LogbookAsyncListener.class</exclude>
```
whiskeysierra, line 119: Why the exclude?

logbook-servlet/src/main/java/org/zalando/logbook/servlet/LogbookFilter.java
```diff
@@ -30,7 +30,8 @@
      * Unique per instance so we don't accidentally share stages between filter
      * instances in the same chain.
      */
-    private final String name = ResponseProcessingStage.class.getName() + "-" + UUID.randomUUID();
+    private final String responseProcessingStageName = ResponseProcessingStage.class.getName() + "-" + UUID.randomUUID();
+    private final String responseWritingStageName = ResponseWritingStage.class.getName() + "-" + UUID.randomUUID();
```
whiskeysierra, line 34: The name suggests that the attribute value would be of type `ResponseWritingStage`, but it seems to be an `AtomicBoolean`.

logbook-servlet/pom.xml
noffke: @whiskeysierra thank you for the hint, I've added a simple test for LogbookAsyncListener.

logbook-servlet/src/main/java/org/zalando/logbook/servlet/LogbookFilter.java
noffke: Thanks, I've renamed it to `ResponseProcessingStage-Synchronization-<UUID>`

whiskeysierra:
👍

whiskeysierra:
I released your fix as 2.6.2. Thanks a lot for your contribution! 🎉

PR #992, opened 2021-04-01
fix: README PathFilter based on property name typo

brnhrdt:
Only a small typo fix for PathFilter description in the LogbookAutoconfig section


fatroom:
Thanks for catching. Seems that PathFilter was copied from headers filters. Looking on source code here would you mind to fix this reference as well? https://github.com/zalando/logbook/blob/51b8abc3ce465a6959aa122895ca22624d0c0532/logbook-api/src/main/java/org/zalando/logbook/PathFilter.java#L18 (should be path instead of headers)

brnhrdt:
It's fixed. Thanks guys for the great work making logbook! I'm happy I could contribute a tiny bit :)

fatroom:
👍

whiskeysierra:
Thanks for your contribution! 📣

PR #997, opened 2021-04-05
Logbook-netty more extractions

sokomishalov:
According to Norman Maurer (netty maintainer), this is the only way to extract scheme:
https://stackoverflow.com/questions/36280376/get-base-uri-of-io-netty-handler-codec-http-httprequest#comment109701966_36281043

### DIFF
--- a/logbook-core/src/test/java/org/zalando/logbook/CommonsLogFormatSinkTest.java
+++ b/logbook-core/src/test/java/org/zalando/logbook/CommonsLogFormatSinkTest.java
@@ -1,5 +1,6 @@
 package org.zalando.logbook;
 
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.mockito.ArgumentCaptor;
 import org.zalando.logbook.DefaultLogbook.SimpleCorrelation;
@@ -9,9 +10,11 @@ import java.io.IOException;
 import java.time.Clock;
 import java.time.Instant;
 import java.time.ZoneOffset;
+import java.util.Locale;
 
 import static java.time.Clock.fixed;
 import static java.time.ZoneOffset.UTC;
+import static java.util.Locale.US;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
@@ -25,6 +28,11 @@ import static org.mockito.Mockito.when;
 final class CommonsLogFormatSinkTest {
 
     private final HttpLogWriter writer = mock(HttpLogWriter.class);
+
+    @BeforeEach
+    void setUp() {
+        Locale.setDefault(US);
+    }
 
     @Test
     void shouldDelegateActive() {
--- a/logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
+++ b/logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
@@ -1,20 +1,29 @@
 package org.zalando.logbook.netty;
 
 import io.netty.buffer.ByteBuf;
+import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.http.HttpRequest;
+import io.netty.handler.codec.http.HttpUtil;
+import io.netty.handler.ssl.SslHandler;
+
+import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import java.net.URI;
 import java.nio.charset.Charset;
-import java.nio.charset.StandardCharsets;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
 import javax.annotation.Nullable;
+
 import lombok.AllArgsConstructor;
 import org.zalando.logbook.HttpHeaders;
 import org.zalando.logbook.Origin;
-import static io.netty.handler.codec.http.HttpHeaderNames.CONTENT_TYPE;
+
 import static io.netty.handler.codec.http.HttpHeaderNames.HOST;
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static lombok.AccessLevel.PRIVATE;
+import static org.zalando.logbook.Origin.LOCAL;
 
 @AllArgsConstructor(access = PRIVATE)
 final class Request implements org.zalando.logbook.HttpRequest, HeaderSupport {
@@ -56,19 +65,34 @@ final class Request implements org.zalando.logbook.HttpRequest, HeaderSupport {
 
     @Override
     public String getScheme() {
-        // TODO pick the real one
-        return "http";
+        final SslHandler handler = context.channel().pipeline().get(SslHandler.class);
+        return handler == null ? "http" : "https";
     }
 
     @Override
     public String getHost() {
-        return request.headers().get(HOST, "unknown");
+        final String host = request.headers().get(HOST);
+        if (host == null) {
+            return extractAddress().map(InetSocketAddress::getHostString).orElse("unknown");
+        } else {
+            return stripPortIfNecessary(host);
+        }
     }
 
     @Override
     public Optional<Integer> getPort() {
-        // TODO implement
-        return Optional.empty();
+        return extractAddress().map(InetSocketAddress::getPort);
+    }
+
+    private String stripPortIfNecessary(String host) {
+        final int separator = host.indexOf(":");
+        return separator == -1 ? host : host.substring(0, separator);
+    }
+
+    private Optional<InetSocketAddress> extractAddress() {
+        final Channel channel = context.channel();
+        final SocketAddress address = origin == LOCAL ? channel.remoteAddress() : channel.localAddress();
+        return address instanceof InetSocketAddress ? Optional.of((InetSocketAddress) address) : Optional.empty();
     }
 
     @Override
@@ -89,13 +113,12 @@ final class Request implements org.zalando.logbook.HttpRequest, HeaderSupport {
     @Nullable
     @Override
     public String getContentType() {
-        return request.headers().get(CONTENT_TYPE);
+        return Objects.toString(HttpUtil.getMimeType(request), null);
     }
 
     @Override
     public Charset getCharset() {
-        // TODO pick the real one
-        return StandardCharsets.UTF_8;
+        return HttpUtil.getCharset(request, UTF_8);
     }
 
     @Override
--- a/logbook-netty/src/main/java/org/zalando/logbook/netty/Response.java
+++ b/logbook-netty/src/main/java/org/zalando/logbook/netty/Response.java
@@ -6,6 +6,8 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.concurrent.atomic.AtomicReference;
 import javax.annotation.Nullable;
+
+import io.netty.handler.codec.http.HttpUtil;
 import lombok.AllArgsConstructor;
 import org.zalando.logbook.HttpHeaders;
 import org.zalando.logbook.Origin;
@@ -49,8 +51,7 @@ final class Response
 
     @Override
     public Charset getCharset() {
-        // TODO pick the real one
-        return StandardCharsets.UTF_8;
+        return HttpUtil.getCharset(response, StandardCharsets.UTF_8);
     }
 
     @Override
--- a/logbook-netty/src/test/java/org/zalando/logbook/netty/RequestUnitTest.java
+++ b/logbook-netty/src/test/java/org/zalando/logbook/netty/RequestUnitTest.java
@@ -1,0 +1,64 @@
+package org.zalando.logbook.netty;
+
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelPipeline;
+import io.netty.channel.local.LocalAddress;
+import io.netty.handler.codec.http.DefaultHttpHeaders;
+import io.netty.handler.codec.http.HttpMethod;
+import io.netty.handler.codec.http.HttpRequest;
+import io.netty.handler.codec.http.HttpVersion;
+import io.netty.handler.ssl.SslHandler;
+import org.junit.jupiter.api.Test;
+
+import static io.netty.handler.codec.http.HttpHeaderNames.CONTENT_TYPE;
+import static io.netty.handler.codec.http.HttpHeaderNames.HOST;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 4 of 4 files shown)


whiskeysierra:
Can you rebase onto main? I had to fix a CVE there which also breaks your build.

sokomishalov:
@whiskeysierra sure, done

whiskeysierra:
Ok, now the tests are failing due to coverage:

```
Warning:  Rule violated for class org.zalando.logbook.netty.Request: lines covered ratio is 0.9, but expected minimum is 1.0
Warning:  Rule violated for class org.zalando.logbook.netty.Request: branches covered ratio is 0.5, but expected minimum is 1.0
```

sokomishalov:
Added unit test to cover `org.zalando.logbook.netty.Request`
P.S. Fixed `CommonsLogFormatSinkTest` on hosts with another default host locale (caused wrong datetime formatting)
P.S.S. Replaced deprecated reactor-netty API in tests

logbook-netty/src/test/java/org/zalando/logbook/netty/LogbookClientHandlerTest.java
```diff
@@ -48,11 +44,8 @@
             )
             .bindNow();
 
-    @SuppressWarnings("deprecation") // the alternative method is not available prior to 1.0.0
     private final HttpClient client = HttpClient.create()
-            .tcpConfiguration(tcpClient ->
-                    tcpClient.doOnConnected(connection ->
-                            connection.addHandlerLast(new LogbookClientHandler(logbook))))
+            .doOnConnected(connection -> connection.addHandlerLast(new LogbookClientHandler(logbook)))
```
whiskeysierra, line 48: The suppression was important, because the replacement method breaks backwards compatibility with prior versions of netty.

logbook-netty/src/test/java/org/zalando/logbook/netty/LogbookClientHandlerTest.java
```diff
@@ -33,11 +33,7 @@
 final class LogbookClientHandlerTest {
 
     private final HttpLogWriter writer = mock(HttpLogWriter.class);
-
-    private final Logbook logbook = Logbook.builder()
-            .strategy(new TestStrategy())
-            .sink(new DefaultSink(new DefaultHttpLogFormatter(), writer))
-            .build();
+    private final Logbook logbook = Logbook.builder().strategy(new TestStrategy()).sink(new DefaultSink(new DefaultHttpLogFormatter(), writer)).build();
```
whiskeysierra, line 36: Can you revert the formatting?

logbook-netty/src/test/java/org/zalando/logbook/netty/LogbookServerHandlerTest.java
```diff
@@ -33,17 +33,10 @@
 final class LogbookServerHandlerTest {
 
     private final HttpLogWriter writer = mock(HttpLogWriter.class);
+    private final Logbook logbook = Logbook.builder().strategy(new TestStrategy()).sink(new DefaultSink(new DefaultHttpLogFormatter(), writer)).build();
```
whiskeysierra, line 36: Same here, reformatting is not needed, is it?

logbook-netty/src/test/java/org/zalando/logbook/netty/LogbookServerHandlerTest.java
```diff
@@ -33,17 +33,10 @@
...
-            .strategy(new TestStrategy())
-            .sink(new DefaultSink(new DefaultHttpLogFormatter(), writer))
-            .build();
-
-    @SuppressWarnings("deprecation") // the alternative method is not available prior to 1.0.0
     private final DisposableServer server = HttpServer.create()
-            .tcpConfiguration(tcpServer ->
-                    tcpServer.doOnConnection(connection ->
-                            connection.addHandlerLast(new LogbookServerHandler(logbook))))
+            .doOnConnection(connection -> connection.addHandlerLast(new LogbookServerHandler(logbook)))
```
whiskeysierra, line 39: Suppression was intentional to preserve backwards compatibility.

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
```diff
@@ -114,6 +131,8 @@ void buffer(final ByteBuf content) {
         state.updateAndGet(state -> state.buffer(request, content));
     }
 
+
+
```
whiskeysierra, line 135:
```suggestion
```

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
```diff
@@ -56,19 +64,29 @@ public String getMethod() {
...
-        return "http";
+        SslHandler handler = context.channel().pipeline().get(SslHandler.class);
+        return handler == null ? "http" : "https";
     }
 
     @Override
     public String getHost() {
-        return request.headers().get(HOST, "unknown");
+        String host = request.headers().get(HOST);
+        if (host != null) {
```
whiskeysierra, line 74: I'd prefer to invert the expression/if.

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
```diff
@@ -56,19 +64,29 @@ public String getMethod() {
...
+            int separator = host.indexOf(":");
+            return separator == -1 ? host : host.substring(0, separator);
```
whiskeysierra, line 76: Would be easier to understand if it were a small private method with a descriptive name, e.g. `stripPortIfNecessary`.

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
```diff
@@ -56,19 +64,29 @@ public String getMethod() {
...
+        } else {
+            return extractAddress().map(NetUtil::getHostname).orElse("unknown");
```
whiskeysierra, line 78: Is this safe to do? Does it to a DNS query? Would it actually report a hostname that the client didn't use?

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
```diff
@@ -56,19 +64,29 @@ public String getMethod() {
...
 
     @Override
     public Optional<Integer> getPort() {
-        // TODO implement
-        return Optional.empty();
+        return extractAddress().map(InetSocketAddress::getPort);
+    }
+
+    private Optional<InetSocketAddress> extractAddress() {
+        SocketAddress address = origin == Origin.LOCAL ? context.channel().remoteAddress() : context.channel().localAddress();
```
whiskeysierra, line 88:
```suggestion
        final Channel channel = context.channel();
        final SocketAddress address = origin == LOCAL ? channel.remoteAddress() : channel.localAddress();
```

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
```diff
@@ -56,19 +64,29 @@ public String getMethod() {
...
+        SslHandler handler = context.channel().pipeline().get(SslHandler.class);
+        return handler == null ? "http" : "https";
     }
 
     @Override
     public String getHost() {
-        return request.headers().get(HOST, "unknown");
+        String host = request.headers().get(HOST);
+        if (host != null) {
+            int separator = host.indexOf(":");
```
whiskeysierra, line 75:
```suggestion
            final int separator = host.indexOf(":");
```

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
```diff
@@ -56,19 +64,29 @@ public String getMethod() {
...
-        // TODO pick the real one
-        return "http";
+        SslHandler handler = context.channel().pipeline().get(SslHandler.class);
+        return handler == null ? "http" : "https";
     }
 
     @Override
     public String getHost() {
-        return request.headers().get(HOST, "unknown");
+        String host = request.headers().get(HOST);
```
whiskeysierra, line 73:
```suggestion
        final String host = request.headers().get(HOST);
```

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
```diff
@@ -56,19 +64,29 @@ public String getMethod() {
 
     @Override
     public String getScheme() {
-        // TODO pick the real one
-        return "http";
+        SslHandler handler = context.channel().pipeline().get(SslHandler.class);
```
whiskeysierra, line 67:
```suggestion
        final SslHandler handler = context.channel().pipeline().get(SslHandler.class);
```

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
whiskeysierra: Static import for Origin allows to use LOCAL directly.

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
```diff
@@ -89,13 +107,12 @@ public HttpHeaders getHeaders() {
     @Nullable
     @Override
     public String getContentType() {
-        return request.headers().get(CONTENT_TYPE);
+        return Optional.ofNullable(HttpUtil.getMimeType(request)).map(CharSequence::toString).orElse(null);
```
whiskeysierra, line 110:
```suggestion
        return Objects.toString(HttpUtil.getMimeType(request), null);
```

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
sokomishalov: fixed

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
sokomishalov: fixed

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
sokomishalov: done

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
sokomishalov: fixed

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
sokomishalov: done

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
sokomishalov: Yes, it is
This method delegates to `InetSocketAddress::getHostString` on java >= 1.7
Replaced with explicit usage of this method.

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
sokomishalov: fixed

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
sokomishalov: fixed

logbook-netty/src/test/java/org/zalando/logbook/netty/LogbookClientHandlerTest.java
sokomishalov: sure

logbook-netty/src/test/java/org/zalando/logbook/netty/LogbookClientHandlerTest.java
sokomishalov: OK, I've reverted

logbook-netty/src/test/java/org/zalando/logbook/netty/LogbookServerHandlerTest.java
sokomishalov: reverted

logbook-core/src/test/java/org/zalando/logbook/CommonsLogFormatSinkTest.java
```diff
@@ -26,6 +28,11 @@
 
     private final HttpLogWriter writer = mock(HttpLogWriter.class);
 
+    @BeforeEach
+    void setUp() {
+        Locale.setDefault(Locale.ENGLISH);
```
whiskeysierra, line 33: Do you happen to have the error (message) at hand that you experienced without this?

logbook-netty/src/test/java/org/zalando/logbook/netty/LogbookClientHandlerTest.java
```diff
@@ -39,6 +39,7 @@
             .sink(new DefaultSink(new DefaultHttpLogFormatter(), writer))
             .build();
 
+
```
whiskeysierra, line 42:
```suggestion
```

logbook-netty/src/main/java/org/zalando/logbook/netty/Request.java
whiskeysierra: I'm still thinking, under which circumstances would this code execute and what does the hostname actually represent in that case? It's essentially a heuristic for the case that we didn't receive a `Host` header, right? We then try to do a reverse lookup of the channel's ip address to find the most likely hostname?

logbook-core/src/test/java/org/zalando/logbook/CommonsLogFormatSinkTest.java
sokomishalov: image
As you can see, month letters are cyrillic on my machine with the default locale (Russian in my case) and it causes test failures.
And I decided that changing formatter locale in `CommonsLogFormatSink` would be breaking change so I added this setup.

whiskeysierra: Actually, I believe you discovered a bug. CLF is using US date time format, afaik.

sokomishalov: Well, I supposed the same way as you. But specification says [here](https://en.wikipedia.org/wiki/Common_Log_Format) and [here](https://strftime.org/):
```
[10/Oct/2000:13:55:36 -0700] is the date, time, and time zone that the request was received, by default in strftime format %d/%b/%Y:%H:%M:%S %z.
.....
%b - Month as **locale’s** abbreviated name (i.e. Sep)
```

whiskeysierra:
👍

whiskeysierra:
@sokomishalov Thank you, for your contributions! It's highly appreciated. Sorry that it took so long and thanks for engaging in the review! 📣 

I just released your changes as 2.7.0 to maven central.

Issue #963, opened 2021-02-24
RestTemplate returns NULL data while using LogbookClientHttpRequestInterceptor


sergkunz:
## Description

Hi, I'm getting NULL data while using LogbookClientHttpRequestInterceptor

Example (Kotlin):

            val restTemplate = RestTemplate()
            val interceptor = LogbookClientHttpRequestInterceptor(logbook)
            restTemplate.interceptors.add(interceptor)
            val response: ResponseEntity<String> = restTemplate.getForEntity("http://localhost:8090/test", String::class.java)
            println(response.body.toString()) // prints null

response.body returns NULL

If I remove LogbookClientHttpRequestInterceptor, response.body returns non-null data

            val restTemplate = RestTemplate()
            val response: ResponseEntity<String> = restTemplate.getForEntity("http://localhost:8090/test", String::class.java)
            println(response.body.toString()) // prints non-null data

Is it a bug or I am doing something wrong?

## Expected Behavior

RestTemplate should return non-null data while using LogbookClientHttpRequestInterceptor

## Actual Behavior
RestTemplate returns NULL data

## Steps to Reproduce
See example above

## Your Environment

* Version used: logbook v2.5.0 (org.zalando:logbook-spring-boot-starter:2.5.0), Spring boot v2.3.0.RELEASE, Kotlin v1.3.72

fmedisov:
It seems, this bug is fixed in scope of [Fix empty response... #948 ](https://github.com/zalando/logbook/pull/948) changes in v2.4.2.

sergkunz:
Hi, thank you for quick response. 
But now I'm getting an ResourceAccessException in v2.4.2

org.springframework.web.client.ResourceAccessException: I/O error on GET request for "http://localhost:8090/test": mark/reset not supported; nested exception is java.io.IOException: mark/reset not supported
	at org.springframework.web.client.RestTemplate.doExecute(RestTemplate.java:748)
	at org.springframework.web.client.RestTemplate.execute(RestTemplate.java:674)
	at org.springframework.web.client.RestTemplate.getForEntity(RestTemplate.java:342)
	at ru.lamoda.wmsprinting.service.SomeRestClient.doRestQuery(SomeRestClient.kt:49)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #998, opened 2021-04-08
fix the RestTemplate interceptor & confirm with wiremock

nhomble:
- Want to help resolve: https://github.com/zalando/logbook/issues/963
- Seems like the `MockRestServiceServer` isn't verifying enough

## Description

The previous attempt to fix was calling `.reset()` which seems dangerous since not every child class impls this method. On my [repro](https://github.com/nhomble/logbook-resttemplate-error-repro), I had `EofSensorInputStream`. 

Instead of trying to reset the response body directly, we can leverage spring's wrappers. This requires no changes to logbook and just an update to documentation. The new test with `wiremock` confirms this behavior. 

## Motivation and Context

https://github.com/zalando/logbook/issues/963
## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -632,8 +632,11 @@ The `logbook-spring` module contains a `ClientHttpRequestInterceptor` to use wit
 ```java
     LogbookClientHttpRequestInterceptor interceptor = new LogbookClientHttpRequestInterceptor(logbook);
     RestTemplate restTemplate = new RestTemplate();
+    restTemplate.setRequestFactory(new BufferingClientHttpRequestFactory(restTemplate.getRequestFactory()));
     restTemplate.getInterceptors().add(interceptor);
 ```
+
+**note** if you use the `spring-boot-starter` then you can enable this with spring properties `logbook.write.interceptors-enabled=true`
 
 ### Spring Boot Starter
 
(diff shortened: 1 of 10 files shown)


whiskeysierra:
> This requires no changes to logbook and just an update to documentation.

That's a breaking change though. If buffering is the only viable solution, then the interceptor should just do that internally, without requiring users to change their code.

nhomble:
> > This requires no changes to logbook and just an update to documentation.
> 
> That's a breaking change though. If buffering is the only viable solution, then the interceptor should just do that internally, without requiring users to change their code.

I don't think we'll get a solution out of the interceptor since it just consumes `ClientHttpResponse` object from spring, so proxying that isn't obvious to me. Maybe we introduce one more customizer at the `RestTemplate` level that configures both the interceptor and the request factory. We could write a `LogbookRestTemplateCustomizer`. 

The `.reset()` just breaks for me so I kept this pr light with a new test - but I am happy add the customizer here for you to see as well.

kyle18th:
@nhomble

kyle18th:
@nhomble How about add a **LOWEST_PRECEDENCE** `RestTemplateCustomizer` to set `BufferingClientHttpRequestFactory` if `RestTemplate` has a `LogbookClientHttpRequestInterceptor`?  If so, we don't need to set a propertity.

```
@Bean
@Order
public RestTemplateCustomizer restTemplateBufferingRequestFactoryCustomizer() {
    return restTemplate -> {
        boolean hasLogbookInterceptor = restTemplate.getInterceptors().stream()
                .anyMatch(LogbookClientHttpRequestInterceptor.class::isInstance);
        if (hasLogbookInterceptor) {
            restTemplate.setRequestFactory(new BufferingClientHttpRequestFactory(restTemplate.getRequestFactory()));
        }
    };
}
```

nhomble:
> @nhomble How about add a **LOWEST_PRECEDENCE** `RestTemplateCustomizer` to set `BufferingClientHttpRequestFactory` if `RestTemplate` has a `LogbookClientHttpRequestInterceptor`? If so, we don't need to set a propertity.

Nice idea @kyle18th, I am open to either approach. Personally, I prefer the customizer to work by existence and not conditional on something else but if you think your approach is a better practice I'm happy to learn.

mrResident:
> @nhomble How about add a **LOWEST_PRECEDENCE** `RestTemplateCustomizer` to set `BufferingClientHttpRequestFactory` if `RestTemplate` has a `LogbookClientHttpRequestInterceptor`? If so, we don't need to set a propertity.
> 
> ```
> @Bean
> @Order
> public RestTemplateCustomizer restTemplateBufferingRequestFactoryCustomizer() {
>     return restTemplate -> {
>         boolean hasLogbookInterceptor = restTemplate.getInterceptors().stream()
>                 .anyMatch(LogbookClientHttpRequestInterceptor.class::isInstance);
>         if (hasLogbookInterceptor) {
>             restTemplate.setRequestFactory(new BufferingClientHttpRequestFactory(restTemplate.getRequestFactory()));
>         }
>     };
> }
> ```

If getting response with HTTP code 401 then may catch exception

"cannot retry due to server authentication, in streaming mode; nested exception is java.net.HttpRetryException: cannot retry due to server authentication, in streaming mode"

whiskeysierra:
This is essentially the same fix as #1041

PR #1013, opened 2021-04-29
append empty headers matching support - for example, if header is abs…

Mikhail-Polivakha:
Append empty headers matching support 

## Description
In general, I have made it possible to apply predicate to header even if it is absent considering header as empty string. Below u can see my triuble with that

## Motivation and Context
II have faced an issue I want to exclude with logbook some requests, that has contain empty Authorization header or **does not contain it at all**. As u can understand header name here is unimportant, it could be easily some another header. Unfortunately, by default, if logbook see that header is absent it is not apply predicate to it at all. This it quite sadly, because what I really need to tell logbook is something like this : "Hey, if Authorization header is absent, or consists of an empty string, that you should not log request and corresponding response". 

## Types of changes

The changes I have made is quite simple:
1) Adjust 2 methods in **Conditions** - I think changes is self descriptive, I have made possible to compare header even if it is absent (in this case header is considered as empty string) with predicate  
2) I have obviously run unit tests and fix one that have failed because of my change. This test btw describes my idea - even if header is empty it possibly could satisfy predicate
3) Append apache commons dependency, to just simplify some ordinary stuff, hope we will make use of this latter also

## Checklist:
I think my changes should be somewhere documented. If you will procure me place where this change should be documented I will certainly supply you with one

Best regards,
Mikhail

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/Conditions.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/Conditions.java
@@ -1,11 +1,14 @@
 package org.zalando.logbook;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apiguardian.api.API;
 import org.zalando.logbook.common.Glob;
 import org.zalando.logbook.common.MediaTypeQuery;
 
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
 import java.util.function.BiPredicate;
 import java.util.function.Function;
 import java.util.function.Predicate;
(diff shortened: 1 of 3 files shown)


whiskeysierra:
Thanks for your contribution!

Unfortunately, the way how it's implemented can't be merged for two reasons:

1. It's a breaking change, i.e. it changes behavior
2. All parameters are non-null by default. See https://github.com/zalando/logbook/blob/main/logbook-core/src/main/java/org/zalando/logbook/package-info.java

Instead of sneaking new behavior into an existing method, why not introducing a new one to test for absent headers?

```java
Logbook.builder()
    .condition(exclude(
        withoutHeader("Authorization").or(header("Authorization", ""))
    ))
   .build()
```

Mikhail-Polivakha:
Uh! I just does not know that I am able to use **withoutHeader()** Condition, so of course, in this case it is not necessary at all. Maybe I am a bit inattentive, but consider doc this opprotunity :) Because I have not seen this functional during me research before merge request (maybe I am just a bit remiss). But thank you!

whiskeysierra:
That condition doesn't exist yet, but it would be a better addition, imo.

Mikhail-Polivakha:
I can take care of this, if I may

whiskeysierra:
Contributions are highly appreciated! 📣

Issue #1003, opened 2021-04-12
Allow filtering of objects and arrays inside the body


eusourafael:
Sometimes a request/response, can contain some properties that are not important for logging purposes.
There could be an option in the library to allow us to filter out those properties (objects or arrays).

## Detailed Description
There could be an option to specify a property name or even better a JSON Path to the properties that should be filtered out from the log.

## Context
We have a Splunk cluster that allows only 10k characters per log. Some of our integrations send us more that this, but we can omit some of the objects and arrays from the log, as they don't contain important information.
This can benefit other people with same issue, or even to reduce the amount of money spent storing logs, etc.

## Possible Implementation
Create a method inside `JsonBodyFilters` class to replace an entire array or entire object.

## Your Environment
* Version used: 2.6.2

PascalSchumacher:
I think this issue should be closed. A JsonPathBodyFilter (see: https://github.com/zalando/logbook#jsonpath-body-filtering-experimental) was added in version `2.9.0`.

eusourafael:
That is true. Great addition to the project. Thank you very much.

PR #1022, opened 2021-05-04
Add JsonPath body filters

vdsirotkin:
Added some JsonPath body filters

## Description
JsonPath makes filtering json messages much easier, comparing to regex approach. Also, some of the methods i've implemented are too complicated to be processed by regexes, i suppose that regex implementation of such features is close to writing your our json processor library. 

## Motivation and Context
So, i've implemented:
* `deleteJsonObject` - deletes key from json object. This one is inspired by issue #1003 
* `replaceJsonStringArray` - replaces all strings in array. It's too complicated to be written using regex.
* `replaceJsonStringDynamically` - replaces string using regex. It's often needed to replace data in json partially. You can check usage in test. This one _could_ be written using regex, but i think JsonPath way is much more clear.
* `replaceJsonString` - just replaces property. For example, this one could be used in case when you don't need to replace all matches of field name in document. To be honest, i wrote this method just to run JMH test :)

## JMH results

Obviously, JsonPath is slower than regex. Here are JMH results.

My machine:
```
Processor Name: Quad-Core Intel Core i7
Processor Speed: 2,5 GHz
Number of Processors: 1
Total Number of Cores: 4
L2 Cache (per Core): 256 KB
L3 Cache: 6 MB
Hyper-Threading Technology: Enabled
Memory: 16 GB
```

JMH Options
```
# JMH version: 1.29
# VM version: JDK 11.0.2, OpenJDK 64-Bit Server VM, 11.0.2+9
# Blackhole mode: full + dont-inline hint
# Warmup: 2 iterations, 3 s each
# Measurement: 5 iterations, 3 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Throughput, ops/time
```

Results:
```
Benchmark                                                       Mode  Cnt       Score        Error  Units
JsonPathBodyFilterBenchmark.replaceStringDynamicallyBenchmark  thrpt    5  273453,753 ±  90998,584  ops/s
JsonPathBodyFilterBenchmark.replaceStringJsonPathBenchmark     thrpt    5  412863,444 ± 117605,330  ops/s
JsonPathBodyFilterBenchmark.replaceStringPrimitiveBenchmark    thrpt    5  786072,679 ±  23076,179  ops/s
```
## Types of changes

## Checklist:


whiskeysierra:
@vdsirotkin That's a great contribution, thank you for proposing it!

I haven't worked with Jayway's JsonPath API directly before.
Let me know if the following suggestions don't work.
After reviewing the API that you crafted, I thought that maybe there is a leaner way to enable the same functionality plus more.

From a user's perspective I'd like to have something like this:

```java
Logbook.builder()
    .bodyFilter(jsonPath("$.password").delete())
    .bodyFilter(jsonPath("$.password").replace("XXX"))
    .bodyFilter(jsonPath("$.pan").replace(compile("(\\d{6})\\d+(\\d{4})"), "$1********$2"))
    .bodyFilter(jsonPath("$.addresses[*].street").replace("XXX"))
    .build();
```

The idea would be split up the problem in two:

1. Select
2. Modify

You can select a set of nodes within the JSON tree/document and apply an operation to each match.
The node could be a single, direct property or an array.
Modifications can be one of:

1. Delete the whole node
2. Replace (static value)
3. Replace (regex, requires primitive values to be either cast to strings or use their string representation?)

vdsirotkin:
@whiskeysierra Your API suggestion looks really nice. I just tried to keep it as close to your's `JsonBodyFilters` as possible.

If you wish, i can keep `JsonPathBodyFilter`, and also write such fluent builder API like you've suggested.

vdsirotkin:
@whiskeysierra plz check latest update. I've decided to remove first API variant, and did just like you suggested. Looks much better even under the hood IMHO :)
You can check usage in tests, looks beautiful.

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
```diff
@@ -0,0 +1,107 @@
...
+
+        public BodyFilter forceReplace(Number replacement) {
+            return replace(replacement, true, Number.class);
+        }
+
+        public BodyFilter replace(String matchRegex, String replacementRegex) {
+            Pattern pattern = Pattern.compile(matchRegex);
+            return (JsonBodyFilter) documentContext -> {
+                Object obj = documentContext.read(jsonPath);
+                if (isPrimitive(obj)) {
```
whiskeysierra, line 45: Could this be simplified by enabling the `ALWAYS_RETURN_LIST` option?

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
```diff
@@ -0,0 +1,107 @@
...
+import net.minidev.json.JSONArray;
+import org.zalando.logbook.BodyFilter;
+
+import javax.annotation.Nullable;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class JsonPathBodyFilters {
```
whiskeysierra, line 15:
```suggestion
public final class JsonPathBodyFilters {
```

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
```diff
@@ -0,0 +1,107 @@
...
+
+        public BodyFilter delete() {
+            return (JsonBodyFilter) documentContext -> documentContext.delete(jsonPath).jsonString();
+        }
+
+        public BodyFilter replace(String replacement) {
+            return replace(replacement, false, String.class);
+        }
+
+        public BodyFilter forceReplace(String replacement) {
```
whiskeysierra, line 29: What does force do?

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
```diff
@@ -0,0 +1,107 @@
...
+        private Optional<String> replaceWithRegex(Object obj, Pattern pattern, String replacementRegex) {
+            String stringRepresentation = obj.toString();
+            Matcher matcher = pattern.matcher(stringRepresentation);
+            if (matcher.find()) {
+                return Optional.of(matcher.replaceAll(replacementRegex));
+            }
+            return Optional.empty();
+        }
+
+        private boolean isPrimitive(Object obj) {
```
whiskeysierra, line 89: Could this be simplified by just calling toString/toText (not sure if that exists here) to just operate on primitives as strings?

whiskeysierra:
Looks nice already. Once we're settled, then we should add a section to the README to advertise this. This change would be great for a lot of folks, judging from past issues.

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
vdsirotkin: This method doesn't actually convert, it only checks if object is primitive. We need this check to be sure, that specified jsonpath doesn't contain some `{}` object.
As you can see, i call `toString()` on that object after check.

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
vdsirotkin: It forces replace :))
For example, we have json `{"test":{}}` and we call `jsonPath("$.test").replace("XXX")`. In such case nothing will be changed, because $.test is neither json array nor primitive. But, if we call `jsonPath("$.test").forceReplace("XXX")`, we'll get `{"test":"XXX"}` as the result of filtering.

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
vdsirotkin: The problem with `ALWAYS_RETURN_LIST` is that it is usually used only if we need to parse some data, but not to put it back to document context.
Here's an example:
```
String json = "{\"test\": \"value\"}";
DocumentContext context = JsonPath.using(Configuration.builder().options(Option.ALWAYS_RETURN_LIST).build()).parse(json);
JsonPath path = JsonPath.compile("$.test");
Object read = context.read(path);
System.out.println(read.getClass());
context.set(path, read);
System.out.println(context.jsonString());
```
This code prints:
```
class net.minidev.json.JSONArray
{"test":["value"]}
```

As you can see, we replaced string value with array value. It's not the desired behavior, i suppose :)

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
vdsirotkin: fixed

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
whiskeysierra: I was expecting that the *force* behavior is the default. I wouldn't have any special semantics for strings, arrays or objects. It should always behave the same way. Just replace every matching node with the given string.

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
whiskeysierra: Does JsonPath provide you with a way to find the definitive path for each match?

whiskeysierra: https://github.com/json-path/JsonPath#path-vs-value that would be nice. You would basically use that option to always get a list of definitive paths and replace those individually.

vdsirotkin: Ok, i just didn't understand your idea :) I think all of the logics may be refactored according to it, including https://github.com/zalando/logbook/pull/1022#discussion_r627321116 this comment

vdsirotkin:
@whiskeysierra plz check the update. I've rewritten everything using `AS_PATH_LIST`. And removed all value type checks. Array replacement looks a bit confusing, but i think this is not the main use-case :)

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
```diff
@@ -0,0 +1,91 @@
...
+                });
+                return documentContext.jsonString();
+            };
+        }
+
+        private BodyFilter replaceInternal(Object replacement) {
+            return (JsonBodyFilter) documentContext -> {
+                List<String> obj = documentContext.read(jsonPath);
+                obj.forEach(path -> documentContext.set(path, replacement));
+                documentContext.set(jsonPath, replacement);
```
whiskeysierra, line 62: When replacing with a static replacement, I guess you don't need to read first, you should be able to call `set` directly:

```suggestion
                documentContext.set(jsonPath, replacement);
```

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
```diff
@@ -0,0 +1,91 @@
...
+        private Optional<String> replaceWithRegex(Object obj, Pattern pattern, String replacementRegex) {
+            String stringRepresentation = obj.toString();
+            Matcher matcher = pattern.matcher(stringRepresentation);
+            if (matcher.find()) {
+                return Optional.of(matcher.replaceAll(replacementRegex));
+            }
+            return Optional.empty();
+        }
+
+    }
```
whiskeysierra, line 76: I believe this conditional/optional handling is not really necessary and could be simplified:

```suggestion
        private String replace(Object node, Pattern pattern, String replacement) {
            Matcher matcher = pattern.matcher(node.toString());
            return matcher.replaceAll(replacement));
        }

    }
```

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
```diff
@@ -0,0 +1,91 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class JsonPathBodyFilters {
+
+    private static final Configuration conf = Configuration.builder().options(Option.AS_PATH_LIST).build();
+
+    @RequiredArgsConstructor
+    public static class JsonPathBodyFilterBuilder {
```
whiskeysierra, line 25:
```suggestion
    @RequiredArgsConstructor(access = PRIVATE)
    public static final class JsonPathBodyFilterBuilder {
```

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
```diff
@@ -0,0 +1,91 @@
...
+                List<String> paths = documentContext.read(jsonPath);
+                Object jsonObject = documentContext.json();
+                paths.forEach(path -> {
+                    Object obj = JsonPath.compile(path).read(jsonObject);
+                    replaceWithRegex(obj, pattern, replacementRegex)
+                            .ifPresent(s -> documentContext.set(path, s));
+                });
+                return documentContext.jsonString();
+            };
+        }
```
whiskeysierra, line 56: Based on the suggestion below:

```suggestion
        public BodyFilter replace(final String matchRegex, final String replacement) {
            final Pattern pattern = Pattern.compile(matchRegex);
            return (JsonBodyFilter) context -> {
                final List<String> paths = context.read(jsonPath);
                final Object document = context.json();
                paths.forEach(path -> {
                    final Object node = JsonPath.compile(path).read(document);
                    context.set(path, replace(node, pattern, replacement));
                });
                return context.jsonString();
            };
        }
```

logbook-json/src/test/java/org/zalando/logbook/json/JsonPathBodyFiltersTest.java
```diff
@@ -0,0 +1,87 @@
...
+        assertThat(result).isEqualTo("{\"test\":{},\"test2\":\"value\"}");
+    }
+
+    @Test
+    public void replaceStringTest() {
+        BodyFilter filter = jsonPath("$.test").replace("XXX");
+        String result = filter.filter(CONTENT_TYPE, "{\"test\": [\"test123\", \"test321\", 1.0], \"test123\": \"testing\"}");
+        assertThat(result).isEqualTo("{\"test\":\"XXX\",\"test123\":\"testing\"}");
+
+        result = filter.filter(CONTENT_TYPE, "{\"test\": 1, \"test123\": \"testing\"}");
```
whiskeysierra, line 36: Tests become more readable and manageable if there is, as a rule of thumb, one assertion per test. I'd extract every filter invocation and the corresponding assertion into its own test method.

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
```diff
@@ -0,0 +1,91 @@
...
+
+import javax.annotation.Nullable;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+public final class JsonPathBodyFilters {
```
whiskeysierra, line 20:
```suggestion
@API(status = EXPERIMENTAL)
@NoArgsConstructor(access = AccessLevel.PRIVATE)
public final class JsonPathBodyFilters {
```

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
vdsirotkin: If we do so, we'll lose all the benefits from `AS_PATH_LIST`...

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
vdsirotkin: That logics were implemented to make replacement a bit more fail-safe. For example, if we have some number, which didn't match regex, it will be converted into String, when using your variant. If we have object/array - it will behave in the same way. 
If it doesn't matter - i can update this logics :)

whiskeysierra: Ah, I totally missed that. Good point. Better leave it like it is then. Is that covered by a test? I guess so.

vdsirotkin:
@whiskeysierra plz check updates. I'd like you to pay a bit of attention to `replaceObjectDynamicallyTest` test, especially to replacement result. I think it looks a bit confusing, but in my opinion it's a rare case.

logbook-json/src/test/java/org/zalando/logbook/json/JsonPathBodyFiltersTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+    @Test
+    public void replaceStringDynamicallyTest() {
+        BodyFilter filter = jsonPath("$.test").replace("(\\d{6})\\d+(\\d{4})", "$1******$2");
+        String result = filter.filter(CONTENT_TYPE, "{\"test\": \"5213486633218931\", \"test123\": \"5213486633218931\"}");
+        assertThat(result).isEqualTo("{\"test\":\"521348******8931\",\"test123\":\"5213486633218931\"}");
+    }
+
+    @Test
+    public void replaceArrayDynamicallyTest() {
```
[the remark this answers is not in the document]
vdsirotkin: Everything may happen...) 
I always assume that user will do everything wrong :D
I've added new test for wildcard behavior :)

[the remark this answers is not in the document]
whiskeysierra: >  I always assume that user will do everything wrong :D

You're absolutely right. Given that the feature is intended to prevent sensitive information from leaking into log files, defaulting to *"stringify JSON array/object and replace values"* is a good and safe default. 

We should add docs for the most common use cases.

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
vdsirotkin: done

vdsirotkin:
@whiskeysierra hi. Check readme updates please. I'm not quite experienced in writing documentation for such projects, especially in English, so plz rewrite it if you find it weird)

whiskeysierra:
> hi. Check readme updates please. I'm not quite experienced in writing documentation for such projects, especially in English, so plz rewrite it if you find it weird)

@vdsirotkin README changes look good. But github claims that the whole file changed, for some reason. Any idea why? Different encoding?

whiskeysierra:
👍

vdsirotkin:
@whiskeysierra i dunno... maybe some "smart" indentions from intellij :)
Edited readme with github online, now it's OK.

whiskeysierra:
@vdsirotkin Thank you for your contribution! 📣 

I just released your changes (+ some modifications of my own) as 2.9.0 to central just now.

PR #1025, opened 2021-05-06
create support for absent header condition

Mikhail-Polivakha:
## Description
Create additional condition in order to support condition - "header is absent in http request" 

## Motivation and Context
Encounter in our project. Requirement was to disable logging for requests (and corresponding responses) without certain headers 

## Types of changes
New feature (and corresponding tests for it also goes along)

## Checklist:
Required tests have been already created. Consider to doc this change somewhere. If you will provide me with the link I can supply with one :)

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/Conditions.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/Conditions.java
@@ -78,4 +78,8 @@ public final class Conditions {
                                 e.getValue().stream().anyMatch(v -> predicate.test(e.getKey(), v)));
     }
 
+    public static <T extends HttpMessage> Predicate<T> withoutHeader(String headerName) {
+        return message -> !message.getHeaders().containsKey(headerName);
+    }
+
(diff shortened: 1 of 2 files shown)


logbook-core/src/main/java/org/zalando/logbook/Conditions.java
```diff
@@ -78,4 +78,8 @@ private Conditions() {
                                 e.getValue().stream().anyMatch(v -> predicate.test(e.getKey(), v)));
     }
 
+    public static <T extends HttpMessage> Predicate<T> withoutHeader(String headerName) {
```
whiskeysierra, line 81:
```suggestion
    public static <T extends HttpMessage> Predicate<T> withoutHeader(final String headerName) {
```

whiskeysierra:
👍

whiskeysierra:
Thank you for your contribution! 📣 

I released your change as 2.8.0 to central.

PR #1036, opened 2021-05-13
Logbook-json: Remove json-smart dependency.

PascalSchumacher:
This avoids the `json-smart` and `accessors-smart` dependencies.

The `JsonHttpLogFormatterTest#shouldNotEmbedReplacedJsonRequestBody` failures seem to expose a problem (not sure if the problem is with the implementation or the test), because the generated JSON is not valid 
`{"origin":"remote","type":"request","correlation":"","protocol":"HTTP/1.1","remote":"127.0.0.1","method":"GET","uri":"http://localhost/","host":"localhost","path":"/","scheme":"http","port":"80","body":<skipped>}` (The result of the second parameterized execution also has a duplicated member body:  
`{"origin":"remote","type":"request","correlation":"","protocol":"HTTP/1.1","remote":"127.0.0.1","method":"GET","uri":"http://localhost/","body":<skipped>,"body":<skipped>}`).

The `beforeAll` methods are there to prevent `json-path-assert` from using `json-smart` and failing. They are unnecessary if `json-smart` is added as a test dependency instead (this would also make the test failure mentioned above go away). Not sure which is the better option.

By the way: Thank you very much for providing logbook.

### DIFF
--- a/logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
@@ -9,6 +9,8 @@ import com.jayway.jsonpath.DocumentContext;
 import com.jayway.jsonpath.JsonPath;
 import com.jayway.jsonpath.ParseContext;
 import com.jayway.jsonpath.spi.json.JacksonJsonNodeJsonProvider;
+import com.jayway.jsonpath.spi.mapper.JacksonMappingProvider;
+
 import lombok.AllArgsConstructor;
 import lombok.NoArgsConstructor;
 import lombok.RequiredArgsConstructor;
@@ -81,6 +83,7 @@ public final class JsonPathBodyFilters {
         private static final ParseContext CONTEXT = JsonPath.using(
                 Configuration.builder()
                         .jsonProvider(new JacksonJsonNodeJsonProvider())
+                        .mappingProvider(new JacksonMappingProvider())
                         .build());
 
         private final Operation operation;
--- a/logbook-json/src/test/java/org/zalando/logbook/json/JsonHttpLogFormatterTest.java
+++ b/logbook-json/src/test/java/org/zalando/logbook/json/JsonHttpLogFormatterTest.java
@@ -2,6 +2,8 @@ package org.zalando.logbook.json;
 
 import lombok.AllArgsConstructor;
 import lombok.Getter;
+
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.MethodSource;
 import org.zalando.logbook.Correlation;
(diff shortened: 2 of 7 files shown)


logbook-json/pom.xml
```diff
@@ -34,6 +34,13 @@
         <dependency>
             <groupId>com.jayway.jsonpath</groupId>
             <artifactId>json-path</artifactId>
+            <exclusions>
```
whiskeysierra, line 37: Wouldn't it be better if we have this exclusion in the parent pom's dependency management, so that it applies to all modules and scopes?

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
```diff
@@ -81,6 +83,7 @@ private static JsonPathBodyFilter filter(final Operation operation) {
         private static final ParseContext CONTEXT = JsonPath.using(
                 Configuration.builder()
                         .jsonProvider(new JacksonJsonNodeJsonProvider())
+                        .mappingProvider(new JacksonMappingProvider())
```
whiskeysierra, line 86: What does a MappingProvider do exactly? I tried this but couldn't observe any difference and the tests still worked.

logbook-json/src/test/java/org/zalando/logbook/json/JsonHttpLogFormatterTest.java
```diff
@@ -34,6 +46,27 @@
 
 final class JsonHttpLogFormatterTest {
 
+    @BeforeAll
+    static void beforeAll() {
+        Configuration.setDefaults(new Defaults() {
```
whiskeysierra, line 51: Instead of modifying a global static variable, couldn't we pass a concrete configuration to the `with` calls down below?

logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
PascalSchumacher: I do not know. I know the default MappingProvider use json-smart, so override it to be save.

logbook-json/src/test/java/org/zalando/logbook/json/JsonHttpLogFormatterTest.java
PascalSchumacher: I do think that is possible. But maybe I am missing something?

PascalSchumacher:
> The `JsonHttpLogFormatterTest#shouldNotEmbedReplacedJsonRequestBody` failures seem to expose a problem (not sure if the problem is with the implementation or the test), because the generated JSON is not valid
> `{"origin":"remote","type":"request","correlation":"","protocol":"HTTP/1.1","remote":"127.0.0.1","method":"GET","uri":"http://localhost/","host":"localhost","path":"/","scheme":"http","port":"80","body":<skipped>}` (The result of the second parameterized execution also has a duplicated member body:
> `{"origin":"remote","type":"request","correlation":"","protocol":"HTTP/1.1","remote":"127.0.0.1","method":"GET","uri":"http://localhost/","body":<skipped>,"body":<skipped>}`).
> 
> The `beforeAll` methods are there to prevent `json-path-assert` from using `json-smart` and failing. They are unnecessary if `json-smart` is added as a test dependency instead (this would also make the test failure mentioned above go away). Not sure which is the better option.

@whiskeysierra Any input on this? Should I just make json-smart a test dependency.

whiskeysierra:
> @whiskeysierra Any input on this? Should I just make json-smart a test dependency?

I'd like to use Jackson exclusively and figure out any issues that we're having without json-smart in our tests.

logbook-json/src/test/java/org/zalando/logbook/json/JsonHttpLogFormatterTest.java
PascalSchumacher: I took another look, it is not possible to pass configuration to with calls. As far as I can tell static configuration is the only way.

PascalSchumacher:
O.k. I have fixed all tests.

whiskeysierra:
👍

PR #1041, opened 2021-05-19
wrap client http response body in buffered input stream to support mark/reset

noffke:
## Description
When using `LogbookClientHttpRequestInterceptor` on spring resttemplate and the the underlying HTTP library (e.g. apache httpclient) returns an input stream that doesn't support mark/reset the response can't be logged by logbook.

## Motivation and Context
Fixes the second exception mentioned in #963

## Types of changes

## Checklist:

### DIFF
--- a/logbook-spring/src/main/java/org/zalando/logbook/spring/RemoteResponse.java
+++ b/logbook-spring/src/main/java/org/zalando/logbook/spring/RemoteResponse.java
@@ -62,6 +62,7 @@ final class RemoteResponse implements HttpResponse {
         @Override
         public State buffer(final ClientHttpResponse response) throws IOException {
             InputStream responseBodyStream = response.getBody();
+            responseBodyStream.mark(Integer.MAX_VALUE);
             byte[] data = ByteStreams.toByteArray(responseBodyStream);
             responseBodyStream.reset();
             return new Buffering(data);
--- a/logbook-spring/src/test/java/org/zalando/logbook/spring/BufferingClientHttpResponseWrapperTest.java
+++ b/logbook-spring/src/test/java/org/zalando/logbook/spring/BufferingClientHttpResponseWrapperTest.java
@@ -1,0 +1,98 @@
+package org.zalando.logbook.spring;
+
+import java.io.BufferedInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.client.ClientHttpResponse;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+class BufferingClientHttpResponseWrapperTest {
+
+    @Mock
+    private ClientHttpResponse delegate;
+
+    @Mock
+    private InputStream inputStream;
+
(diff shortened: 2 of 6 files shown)


logbook-spring/src/main/java/org/zalando/logbook/spring/RemoteResponse.java
```diff
@@ -62,6 +62,7 @@ public State without() {
         @Override
         public State buffer(final ClientHttpResponse response) throws IOException {
             InputStream responseBodyStream = response.getBody();
+            responseBodyStream.mark(Integer.MAX_VALUE);
```
whiskeysierra, line 65: Why is that needed?

logbook-spring/src/test/java/org/zalando/logbook/spring/BufferingClientHttpResponseWrapperTest.java
```diff
@@ -0,0 +1,82 @@
+package org.zalando.logbook.spring;
```
whiskeysierra, line 1: It would be nice to have a more realistic test which verifies that a stream can be read multiple times.

logbook-spring/src/main/java/org/zalando/logbook/spring/RemoteResponse.java
noffke: Doesn't a stream need to be marked before it can be reset?

logbook-spring/src/test/java/org/zalando/logbook/spring/BufferingClientHttpResponseWrapperTest.java
noffke: Do you have a suggestion for a more realistic test? The Spring Mock Server always returns a byte array input stream which doesn't exhibit the original problem.

logbook-spring/src/main/java/org/zalando/logbook/spring/RemoteResponse.java
whiskeysierra: You're probably right. Interesting that this never caused any issue before, since we had `reset` in there for a while.

logbook-spring/src/test/java/org/zalando/logbook/spring/BufferingClientHttpResponseWrapperTest.java
noffke: Ok, I've adapted the test to do real network requests instead of mock requests. This way, if the `BufferingClientHttpResponseWrapper` is not used, the original "mark/reset not supported" exception will be thrown.

whiskeysierra: > This way, if the `BufferingClientHttpResponseWrapper` is not used, the original "mark/reset not supported" exception will be thrown.

Perfect! Exactly the kind of test I like for this kind problem.

whiskeysierra:
Mmh, the spring4 tests are failing because of lack of coverage. Any clue why that is?

noffke:
I have unfortunately no clue why the spring4 tests fail. I originally also wasn't able to run them locally. When commenting out the `logbook-servlet` module in the root pom, I'm able to progress the build to the `logbook-spring` module. I was then not able to make it work as a spring boot test so I've changed the test to use wiremock instead. After that, I was able to successfully run "spring4" profile locally.

noffke:
@whiskeysierra did I miss anything here for the PR to be approved?

whiskeysierra:
Sorry for the delay, I was on vacation.

Coming back to this I finally remembered why we added reset to begin with. It's meant to be a more efficient alternative to buffering. I feel that it should be changed to the following:

- check if mark/reset is supported (using `markSupported()`)
- if it's supported, then mark it in the beginning, then reset afterwards
- if it's not suppoirted, wrap the stream to enable buffering

noffke:
Thanks for the clarification. I'll see what I can do but I'm rather busy at the moment so it might take a moment to follow up, sorry.

DaniloVeljovic:
@noffke Any follow ups here? We would be interested in merging this PR, if you would resolve @whiskeysierra remarks

noffke:
@DaniloVeljovic sorry, completely forgot about this, too much work :-) Does the PR look acceptable now?

DaniloVeljovic:
LGTM 👍🏼

PR #1066, opened 2021-06-16
Do not fail on missing JSON Path (fixes #1049, fixes #1056)

### LINKED ISSUE #1049: Error using bodyFilter with jsonPath
I'm using logbook version 2.9.0 and I'd like to mask a specific attribute (cpf) in the request and response body. The bean that I created is below:

```
@Bean 
public Logbook logbook() {
   return Logbook.builder()
        .bodyFilter(jsonPath("$.cpf").replace(compile("(\\d{3})(\\d{3})(\\d{3})-?(\\d{2})"), "$1.XXX.$3-XX"))
        .bodyFilter(jsonPath("$.results.client.cpf").replace(compile("(\\d{3})(\\d{3})(\\d{3})-?(\\d{2})"), "$1.XXX.$3-XX"))
        .build();
}
```
When I make an request with body {"cpf": "21343575698"}, an error occur: "Missing property in path $['results']","trace":"com.jayway.jsonpath.PathNotFoundException: Missing property in path $['results']"

is there something wrong with my configuration?

Thanks a lot.

### LINKED ISSUE #1056: Need help with bodyFilter
Hi ! Although it is not an issue, just wondering if can be helped. I am trying to mask one of the field in my request with the name as Value, which works fine with below.
Properties is my set with that field's name.
However I have another field with same name in my Json request and it is masking that too. Is it possible if I can specify or mark somehow that which field needs to be masked?

  Logbook.builder().sink(new DefaultSink(new SplunkHttpLogFormatter(), new DefaultHttpLogWriter()))
        .bodyFilter(JsonBodyFilters.replaceJsonStringProperty(properties, "<masked>")

PascalSchumacher:


whiskeysierra:
:+1:

whiskeysierra:
> I solved this by catching the com.jayway.jsonpath.PathNotFoundException directly where it was needed.

Sounds like the better way.

PascalSchumacher:
@SpiReCZ O.k thanks, feel free to submit a pull request with a better fix, then we can just close this one. :)

SpiReCZ:
> 
> 
> @SpiReCZ O.k thanks, feel free to submit a pull request with a better fix, then we can just close this one. :)

I mean, from a framework that should do a side job apart from the primary business logic. I guess we do not want the logbook to throw any unneeded exceptions and fail without compromising the request processing. 

By that logic, your solution is the best there could be. I was just not sure if that was wanted or by-product of the Option.SUPPRESS_EXCEPTIONS.

I will surely try to create a pull request for my solution if it is any better...

whiskeysierra:
>  I guess we do not want the logbook to throw any unneeded exceptions and fail without compromising the request processing.
> 
> By that logic, your solution is the best there could be.

Technically correct. But... especially JSON path filtering is primarily intended to prevent sensitive data from ending up in logs. It could very well be a trade-off that users would make differently. Essentially it boils down to:

1. Never fail request processing, even if it means sensitive data might end up in logs
2. Never log sensitive data, even if it means requests might be rejected

whiskeysierra:
@PascalSchumacher Can you resolve the conflict? I believe your version is a good first fix. We can always make it better later.

PascalSchumacher:
I have resolved the merge conflict.

whiskeysierra:
👍

PR #1079, opened 2021-06-24
Update README.md replaceBody filters example

SpiReCZ:
The original example provided was not compatible with the current code.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,8 +229,8 @@ import static org.zalando.logbook.QueryFilters.accessToken;
 import static org.zalando.logbook.QueryFilters.replaceQuery;
 
 Logbook logbook = Logbook.builder()
-    .requestFilter(RequestFilters.replaceBody(contentType("audio/*"), "mmh mmh mmh mmh"))
-    .responseFilter(ResponseFilters.replaceBody(contentType("*/*-stream"), "It just keeps going and going..."))
+    .requestFilter(RequestFilters.replaceBody(message -> contentType("audio/*").test(message) ? "mmh mmh mmh mmh" : null))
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -229,8 +229,8 @@ import static org.zalando.logbook.QueryFilters.accessToken;
 import static org.zalando.logbook.QueryFilters.replaceQuery;
 
 Logbook logbook = Logbook.builder()
-    .requestFilter(RequestFilters.replaceBody(contentType("audio/*"), "mmh mmh mmh mmh"))
-    .responseFilter(ResponseFilters.replaceBody(contentType("*/*-stream"), "It just keeps going and going..."))
+    .requestFilter(RequestFilters.replaceBody(message -> contentType("audio/*").test(message) ? "mmh mmh mmh mmh" : null))
```
whiskeysierra, line 232: Technically, the correct version (what I failed to do properly the last time I updated the README) would be this:

```java
.requestFilter(RequestFilters.replaceBody(BodyReplacers.replaceBody(contentType("audio/*"), "mmh mmh mmh mmh")))
```

Writing this for the first time myself (I seem to have never used it like this) made me realize, that it's not the best API that we could possibly offer here.

- It uses the method name twice. Since they are both in different classes, one can't statically import both of them.
- There should be an easier way to transform a body replacer into a request/response filter.

Maybe this is a good opportunity to change `BodyReplacer` and make it extend `RequestFilter` and `ResponseFilter`. The default implementation could just call the static methods that we already have. (That would be a bit tricky and probably require some kind of SPI trick, since the interfaces are in the api module, but the static factories in core.)

The usage would then change to:

```java
.requestFilter(BodyReplacers.replaceBody(contentType("audio/*"), "mmh mmh mmh mmh"))
```

whiskeysierra:
👍

Issue #1071, opened 2021-06-17
For remote request  write response body before request body. Relevant for Netty 


straestrae:
## Description
If use spring webflux with netty then request body will be written together with response. Like:
{
  "timestamp": "2021-06-17T17:41:15.122+02:00",
  "level": "TRACE",
  "thread": "reactor-http-nio-2",
  "context": "default",
  "message": "{\"origin\":\"remote\",\"type\":\"request\",\"correlation\":\"afcc716b4a81cc2d\",\"protocol\":\"HTTP/1.1\",\"remote\":\"/0:0:0:0:0:0:0:1:63141\",\"method\":\"POST\",\"uri\":\"http://localhost:8080/sample\",\"Content-Type\":[\"application/json\"],\"Host\":[\"localhost:8080\"]},\"body\":Some response body{\r\n    \"RequestBody\" : \"RequestBodyValue\"\r\n}}",
  "source": "org.zalando.logbook.Logbook",
  "application_name": "demo"
}
In log message above   **"body\":Some response body{\r\n    \"RequestBody\" : \"RequestBodyValue\"\r\n}}"**

## Expected Behavior
Should log only request body 
 **"body\":{\r\n    \"RequestBody\" : \"RequestBodyValue\"\r\n}}"**

## Actual Behavior
Body contains response then request.

## Possible Fix
Remove line
org/zalando/logbook-netty/2.10.0/logbook-netty-2.10.0-sources.jar!/org/zalando/logbook/netty/LogbookServerHandler.java:69

## Steps to Reproduce

1. Add logbook to Spring webflux application
2. Create test controller:
```
@RestController
public class ControllerForTest {

    @RequestMapping(value = "/sample",
            method = RequestMethod.POST)
    public Mono<String> test() {
        return Mono.fromCallable(() -> "Some response body");
    }
}
```
3. Call controller

## Your Environment

* Version used:
* Link to your project:

whiskeysierra:
That's so weird!

Are you able to reproduce that in an isolated unit test?

straestrae:
> That's so weird!
> 
> Are you able to reproduce that in an isolated unit test?

Below simple project to illustrate. Inside logs after test bug will present 
[demo.zip](https://github.com/zalando/logbook/files/6672825/demo.zip)

sokomishalov:
@straestrae current workaround is to add dummy `@RequestBody` inside your controller

sokomishalov:
Well, I've investigated a bit on this bug. 
[...]

PR #1100, opened 2021-07-25
Possible fix for #1071

sokomishalov:
Possible fix for #1071


whiskeysierra:
Good catch! Looks like a typical copy-&-paste error... 🤦‍♂️

whiskeysierra:
👍

whiskeysierra:
👍

sokomishalov:
> Good catch! Looks like a typical copy-&-paste error... 🤦‍♂️

I supposed the same but was worried a bit if there is some explanation for this line

PR #1105, opened 2021-08-06
logbook-ktor modules

sokomishalov:
## Description

Client and server [ktor](https://ktor.io/) modules

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ Logbook is ready to use out of the box for most common setups. Even for uncommon
 - OkHttp 2.x **or 3.x** (optional)
 - Spring 4.x **or 5.x** (optional)
 - Spring Boot 1.x **or 2.x** (optional)
+- Ktor (optional)
 - logstash-logback-encoder 5.x (optional)
 
 ## Installation
@@ -70,7 +71,8 @@ Alternatively, you can import our *bill of materials*...
 </dependencyManagement>
 ```
 
-... which allows you to omit versions:
+<details>
+  <summary>... which allows you to omit versions:</summary>
 
 ```xml
 <dependency>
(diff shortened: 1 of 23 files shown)


whiskeysierra:
@sokomishalov 📣 Thank you for your contribution! I'm really looking forward to this!

If you can take another look at this failing test case, we can get this merged and released soon.

sokomishalov:
I've finally fixed those failing tests and made some API cleanup.
My main concern is - should we split this module on client and server side? Each of them requires its own non-tiny dep (ktor-client-core-jvm ~630kb, ktor-server-core ~780kb)

whiskeysierra:
>  My main concern is - should we split this module on client and server side? Each of them requires its own non-tiny dep (ktor-client-core-jvm ~630kb, ktor-server-core ~780kb).

Mmh, good question. Which use case do we anticipate the most, client-only, server-only or both within a single codebase?

sokomishalov:
I would expect a single codebase, but we are able to provide it via two separate modules (`logbook-ktor-client` and `logbook-ktor-server`) and one "starter" module (`logbook-ktor`) with both client and server deps.

whiskeysierra:
> I would expect a single codebase, but we are able to provide it via two separate modules (`logbook-ktor-client` and `logbook-ktor-server`) and one "starter" module (`logbook-ktor`) with both client and server deps.

So we would have four modules?

1. `logbook-ktor-common`, common code base, optionally depends on client + server deps
2. `logbook-ktor-client`, depends on common + client deps
3. `logbook-ktor-server`, depends on common + server deps
4. `logbook-ktor`, depends on client + server

sokomishalov:
> 1. `logbook-ktor-common`, common code base, optionally depends on client + server deps

@whiskeysierra I don't think we need a common module - the only common code for now is `@ExperimentalLogbookKtorApi`.
I suppose to introduce common global annotation for optional kotlin cases - `@ExperimentalLogbookApi` to require temporary opt-in for these features. Maybe it would useful somehow later in other kotlin-first modules (such as okhttp).
How do you think?

sokomishalov:
@whiskeysierra Well, It's almost done, just one weird failing test left yet, while locally it's all just fine - I suppose it's some race condition somewhere

whiskeysierra:
👍

whiskeysierra:
Looks great! 📣 Thanks for your contribution! Much appreciated! 🎉 

I'll try to create a new release later today.

sokomishalov:
@whiskeysierra Please take a look at [a bug-fix and test-fix PR for ktor](https://github.com/zalando/logbook/pull/1116) before release.

whiskeysierra:
I'll create a follow-up, bugfix.

PR #1145, opened 2021-09-09
Fix NullPointerException when the element value is null and PathNotFoundException when body is unwrapped Array

ismail2ov:
We are using JsonPath dynamic value replacement, but when any JSON element has a null value, we get a  NullPointerException. 
This RP fixes this issue.

### DIFF
--- a/logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
@@ -97,7 +97,8 @@ public final class JsonPathBodyFilters {
                 @Nullable final String contentType, final String body) {
 
             if (JSON.test(contentType)) {
-                final DocumentContext original = CONTEXT.parse(body);
+                String normalizedBody = body.trim().startsWith("[") ? "{\"content\":" + body + "}" : body;
+                final DocumentContext original = CONTEXT.parse(normalizedBody);
                 return operation.filter(original).jsonString();
             }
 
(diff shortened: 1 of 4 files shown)


fatroom:
👍

whiskeysierra:
Does the NPE come from our use of `toString()` there? Instead of not calling the replacement function, what about allowing the replacement function to handle null itself? That would be a breaking change though.

ismail2ov:
Yes, the NPE came from the use of toString (). I try to handle the exception in the  replacement function, but the NPE came before the execution of the function.

whiskeysierra:
> I try to handle the exception in the replacement function, but the NPE came before the execution of the function.

Yes, correct.

I see two options so far:

1. Detect null, translate to a `NullNode`, don't call the replacement function:
   * Pros:
     * Fixes NPE
   * Cons:  
     * Does't allow replacing of absent or `null` values.
2. Detect null, translate it to `null` (don't call `toString` on it), pass it to the replacement function:
   * Pros:
     * Fixes NPE
     * Does't allow replacing of absent or `null` values.
   * Cons:  
     * Technically breaks existing clients who have non-nullsafe replacement functions.
       **Although, due to this bug right now, they are broken anyway.**

ismail2ov:
What would be the reason for substituting null or absent values?  We use it to obfuscate sensitive data like phone, card number...
Existing customers who have non-null safe replacement features already have a problem.

ismail2ov:
I have also added the PathNotFoundException correction when the body is unwrapped Array. This happens because JsonProvider is defined as com.jayway.jsonpath.spi.json.JacksonJsonNodeJsonProvider

ismail2ov:
@whiskeysierra Can you check this PR?

atrujillofalcon:
@ismail2ov +1

PR #1165, opened 2021-09-21
Dynamic path parameters replacing of PathFilters

### LINKED ISSUE #1164: Dynamic path parameters replacing of PathFilters
t would be quite useful and consistent add a functional replacing interface to PathFilters, in the same mode that JsonPath filters. (Related to [#1127](https://github.com/zalando/logbook/issues/1127) )

I will create a PR to add this feature. 

## Context and Possible Implementation
We need to salt with SHA512 the path param value instead of fixed replacing.

```
List<PathFilter> pathFilters = new ArrayList<>();

pathFilters.add(DynamicPathFilter.replace("/profiles/{organization}/email/{email}/user", DigestUtils::sha512Hex));

Logbook logbook = Logbook.builder()
        .pathFilters(pathFilters)
        .build();
```

ismail2ov:
## Description
Added a new method to PathFilters with a String UnaryOperator to allow the dynamic customization of replacing target value.

## Motivation and Context
Resolves #1164 

## Types of changes

## Checklist:

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/DynamicPathFilter.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/DynamicPathFilter.java
@@ -1,0 +1,66 @@
+package org.zalando.logbook;
+
+import java.util.Objects;
+import java.util.function.UnaryOperator;
+
+final class DynamicPathFilter implements PathFilter {
+
+    private final UnaryOperator<String> replacementFunction;
+    private final String[] parts;
+
+    DynamicPathFilter(final UnaryOperator<String> replacementFunction, final String pathExpression) {
+        this.replacementFunction = replacementFunction;
+
+        final String[] parts = pathExpression.split("/");
+
+        for (int i = 0; i < parts.length; i++) {
+            final String part = parts[i];
+            if (part.startsWith("{") && part.endsWith("}")) {
+                parts[i] = null;
+            }
+        }
+
+        this.parts = parts;
+    }
+
+
+    @Override
+    public String filter(String path) {
+        final String[] pathParts = path.split("/");
+
+        if (parts.length > pathParts.length) {
+            return path;
+        }
+
+        final StringBuilder builder = new StringBuilder();
+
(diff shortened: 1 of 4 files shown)


logbook-core/src/main/java/org/zalando/logbook/DynamicPathFilter.java
```diff
@@ -0,0 +1,66 @@
...
+                parts[i] = null;
+            }
+        }
+
+        this.parts = parts;
+    }
+
+
+    @Override
+    public String filter(String path) {
```
whiskeysierra, line 28: Is this logic similar to the one from the default path filter and if yes, can it be shared somehow?

whiskeysierra:
That's a really nice feature! @ismail2ov Thanks for contributing it! 📣

whiskeysierra:
@ismail2ov Can you take a look at the code coverage?

logbook-core/src/main/java/org/zalando/logbook/DynamicPathFilter.java
ismail2ov: Yes, the logic is similar, but not quite. I prefer to overwrite the method instead of doing a generic one with many IFs

whiskeysierra: Yeah, good call!

ismail2ov:
> @ismail2ov Can you take a look at the code coverage?

@whiskeysierra Can you check where the problem is? 
JaCoCo gives me 100% coverage for DynamicPathFilter

whiskeysierra:
Line coverage yes, but there is an uncovered branch. If you run `mvn clean verify` locally and check `target/site/jacoco/index.html` afterwards you'll find the line.

ismail2ov:
Done

whiskeysierra:
👍

whiskeysierra:
🚀 Great, thank you for your contribution! 📣

PR #1201, opened 2021-10-21
logbook-webflux fixes

### LINKED ISSUE #1200: Missing response log for GET request for Webflux integration
https://github.com/zalando/logbook/blob/01c8a8112867ab014faccbfefae245fd675d7e9e/logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/LogbookWebFilter.java#L41

Hi, we really appreciate that Logbook has Webflux support and doing the integration. During the integration, we found out that the response log might not be written down. Here are our finding:

For the GET request, the state transitions in `LogbookWebFilter.java` might not be complete. So the last write() might not be called in the `triggerStageChange`.
```
private void triggerStageChange(AtomicReference<Object> stage, ServerRequest request, ServerResponse response) {
    stage.updateAndGet(throwingUnaryOperator(s -> {
      if (s instanceof Logbook) return ((Logbook) s).process(request);
      if (s instanceof Logbook.RequestWritingStage) return ((Logbook.RequestWritingStage) s).write().process(response);
      if (s instanceof Logbook.ResponseWritingStage) ((Logbook.ResponseWritingStage) s).write();
      return null;
    }));
  }
```

A further look in the four calls to `triggerStageChange` shows that the `getBody()` in `BufferingServerHttpRequest` is not being triggered during the entire processing for the GET request, so we are missing one of the `triggerStageChange` calls by using the buffer hook. 

We are using undertow server, here is how we get the Logbook and register the web filter:
```
  @Bean
  public Logbook getLogbook() {
    return Logbook.builder().build();
  }

  @Bean
  public WebFilter logbookServerFilter(final Logbook logbook) {
    return new LogbookWebFilter(logbook);
  }
```
We are not sure if we are missing other initial steps thus causing the missing `getBody()`. @sokomishalov , could you help to confirm what we saw here?

A simple trick seems to fix the issue by adding an additional `triggerStageChange` after the last `doOnSuccess`:
```
  return Mono
    .just(exchange)
    .doOnNext((e) -> triggerStageChange(stage, serverRequest, serverResponse))
    .map(e -> exchange
            .mutate()
            .request(new BufferingServerHttpRequest(e.getRequest(), serverRequest, bytes -> triggerStageChange(stage, serverRequest, serverResponse)))
            .response(new BufferingServerHttpResponse(e.getResponse(), serverResponse, bytes -> triggerStageChange(stage, serverRequest, serverResponse)))
            .build()
    )
    .flatMap(chain::filter)
    .doOnSuccess(it -> triggerStageChange(stage, serverRequest, serverResponse)) // ensure response has been written
    .doOnSuccess(it -> triggerStageChange(stage, serverRequest, serverResponse)) // Force to trigger stage change again
    .then();
```

Looks like it does no harm to other requests like PUT/POST. But we are not sure if it is the right way to do it.

sokomishalov:
Fixes #1200, #1219

## Types of changes

## Checklist:

### DIFF
--- a/logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/LogbookWebFilter.java
+++ b/logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/LogbookWebFilter.java
@@ -11,7 +11,7 @@ import reactor.core.publisher.Mono;
 import java.util.concurrent.atomic.AtomicReference;
 
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
-import static org.zalando.fauxpas.FauxPas.*;
+import static org.zalando.fauxpas.FauxPas.throwingUnaryOperator;
 
 
 @RequiredArgsConstructor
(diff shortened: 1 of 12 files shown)


PascalSchumacher:
@whiskeysierra It would be great if this was reviewed, merged and released!

logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/LogbookWebFilter.java
```diff
@@ -30,24 +30,27 @@
...
     }
 
-    private void triggerStageChange(AtomicReference<Object> stage, ServerRequest request, ServerResponse response) {
-        stage.updateAndGet(throwingUnaryOperator(s -> {
-            if (s instanceof Logbook) return ((Logbook) s).process(request);
-            if (s instanceof Logbook.RequestWritingStage) return ((Logbook.RequestWritingStage) s).write().process(response);
-            if (s instanceof Logbook.ResponseWritingStage) ((Logbook.ResponseWritingStage) s).write();
-            return null;
-        }));
+    private void triggerStageChange(AtomicReference<Object> stage, ServerRequest request, ServerResponse response, Boolean forceFinish) {
```
whiskeysierra, line 45: I can't remember reviewing the original version of this method. There has to be a cleaner way to do this...

sokomishalov: Do you mean this stage transition or do-while? If it's regarding the cycle, I've removed it.

whiskeysierra:
👍

PascalSchumacher:
@whiskeysierra A logbook release with this (and the other fixes) would be great!

envious:
has this been released yet?

mfvanek:
Hi there! Could you please release a new version with this fix?

Ellstrom:
Is there any planned release-date for this fix?

kidhack83:
Any news about new release? The last release (2.14.0) is from Oct 12, 2021 (more than one year ago)

sergiotudela:
@whiskeysierra any update to release this fix?

malshoff:
This is a problem for our org. Any plans on a release with this fix?

PR #1218, opened 2021-11-25
Fix NPE in json replace filter

sokomishalov:
## Types of changes

## Checklist:

### DIFF
--- a/logbook-json/src/test/java/org/zalando/logbook/json/JsonPathBodyFiltersTest.java
+++ b/logbook-json/src/test/java/org/zalando/logbook/json/JsonPathBodyFiltersTest.java
@@ -1,6 +1,5 @@
 package org.zalando.logbook.json;
 
-import com.fasterxml.jackson.databind.node.DoubleNode;
 import com.google.common.io.Resources;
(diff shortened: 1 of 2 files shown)


logbook-json/src/test/java/org/zalando/logbook/json/JsonPathBodyFiltersTest.java
```diff
@@ -28,13 +28,14 @@
 import static org.zalando.logbook.json.JsonBodyFilters.accessToken;
 import static org.zalando.logbook.json.JsonPathBodyFilters.jsonPath;
 
+@SuppressWarnings("UnstableApiUsage")
 class JsonPathBodyFiltersTest {
 
-    private final String type = "application/json";
```
whiskeysierra, line 33: Those changes are not really necessary for the fix itself, are they?

sokomishalov: Yes, they are not necessary at all. I'll revert it

whiskeysierra:
👍

Issue #1219, opened 2021-12-05
logbook-spring-webflux: LogbookExchangeFilterFunction does not log bodies of chunked responses


PascalSchumacher:
Using `WebClient` (Netty) we tried to switch from `LogbookClientHandler` to `LogbookExchangeFilterFunction` (from `logbook-spring-webflux`, version `2.14.0`), but when using `LogbookExchangeFilterFunction` the bodies of chunked PATCH responses are no longer logged.

LogbookClientHandler:
```
Incoming Response: c1b96b4bc972c9cd
Duration: 8 ms
HTTP/1.1 200 OK
Connection: close
Content-Type: application/json
Date: Fri, 03 Dec 2021 09:54:12 GMT
Transfer-Encoding: chunked

json body (ommited)
```

LogbookExchangeFilterFunction:

```
Incoming Response: 8f9f882da93f2ae4
Duration: 14 ms
HTTP/1.1 200 OK
Connection: close
Content-Type: application/json
Date: Fri, 03 Dec 2021 08:51:31 GMT
Transfer-Encoding: chunked
```

I guess it is not specific to PATCH. I tried to replicate this in a unit test but failed. In general PATCH bodies are logged, that why I think it may be related to `Transfer-Encoding: chunked`.

Edit:
I did some further testing and is not PATCH specific, the body is also missing for GET responses with `Transfer-Encoding: chunked`.

I was able to replicate this behavior in a unit test, see: https://github.com/zalando/logbook/pull/1221

PascalSchumacher:
Was fixed with: https://github.com/zalando/logbook/pull/1201 Thanks!

PR #1221, opened 2021-12-05
logbook-spring-webflux: LogbookExchangeFilterFunction does not log bo…

PascalSchumacher:
…dies of chunked responses (issue #1219)

No idea if this is the correct way to fix it.


whiskeysierra:
@PascalSchumacher There are conflicts now, unfortunately.

sokomishalov:
@PascalSchumacher Fixed in #1201

PascalSchumacher:
@sokomishalov Thanks!

PR #1304, opened 2022-03-23
Add support for Extended Log Format

greg65236592:
Add support for Extended Log Format(ELF)

## Description

Adding support for Extended Log Format(ELF).
From Wikipedia: ELF is a standardised text file format, like [Common Log Format](https://en.wikipedia.org/wiki/Common_Log_Format) (CLF), that is used by web servers when generating [log files](https://en.wikipedia.org/wiki/Server_log), but ELF files provide more information and flexibility.
ELF is also described in [W3C](https://www.w3.org/TR/WD-logfile.html).

## Motivation and Context

Logbook usually separates the logging of request and response like [described](https://github.com/zalando/logbook#request), except for [Common Log Format](https://github.com/zalando/logbook#common-log-format).

In some cases, we might want use tools to query or analyze server logs, and it could be much easier to get it done if we could print the request and response log in one row.

`CommonsLogFormatSink` is the only one that can do this among all sinks. If we like to print other fields like request time taken, or some special header values, CLF is unable to provide such information.

Thus, I would like to bring in Extended Log Format(ELF) for extending the CLF, which allows users to customize log fields, and outputs the request and response information together in the same row.

## Types of changes

## Checklist:


logbook-core/src/main/java/org/zalando/logbook/ExtendedLogFormatSink.java
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+    }
+
+    private List<String> getSupportedFields() {
+        return Arrays.stream(Field.values()).map(field -> field.value)
+                .collect(Collectors.toList());
+    }
+
+    // validate fields, ignore unknown fields
```
DaniloVeljovic, line 220: Can we remove this comment, please?

logbook-core/src/main/java/org/zalando/logbook/ExtendedLogFormatSink.java
```diff
@@ -0,0 +1,245 @@
...
+        fieldValMap.put(Field.RESP_STATUS.value, status);
+        fieldValMap.put(Field.RESP_COMMENT.value, comment);
+        fieldValMap.put(Field.REQ_METHOD.value, method);
+        fieldValMap.put(Field.REQ_URI.value, uri);
+        fieldValMap.put(Field.REQ_URI_STEM.value, uriStem);
+        fieldValMap.put(Field.REQ_URI_QUERY.value, uriQuery);
+
+        final List<String> outputFields = new ArrayList<>(fields);
+        final String output = outputFields.stream().map(outputField -> {
+                    if (supportedFields.contains(outputField)) {
```
DaniloVeljovic, line 144: Would it be possible to create a private function and invoke it here, instead of this giant block? It is not as easy to read this way

logbook-core/src/main/java/org/zalando/logbook/ExtendedLogFormatSink.java
```diff
@@ -0,0 +1,245 @@
...
+                    }
+                    return OMITTED_FIELD;
+                })
+                .reduce((f1, f2) -> String.join(DELIMITER, f1, f2))
+                .orElse("");
+
+        writer.write(correlation, output);
+    }
+
+    private String getCsHeader(HttpRequest request, String key) {
```
DaniloVeljovic, line 165: Can we make all method parameters (in all of the functions `final`) ?

logbook-core/src/main/java/org/zalando/logbook/ExtendedLogFormatSink.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    @Override
+    public boolean isActive() {
+        return writer.isActive();
+    }
+
+    @Override
+    public void write(final Precorrelation precorrelation, final HttpRequest request) {
+        // nothing to do...
```
DaniloVeljovic, line 67: Out of curiosity, why do we do nothing here?

logbook-spring/src/test/java/org/zalando/logbook/spring/RemoteResponseTest.java
```diff
@@ -7,19 +7,9 @@
...
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
 
 class RemoteResponseTest {
 
-    @Test
-    void statusCanThrow() throws IOException {
```
DaniloVeljovic, line 17: Any particular reason why this test was removed?

logbook-core/src/main/java/org/zalando/logbook/ExtendedLogFormatSink.java
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    @Override
+    public void write(final Correlation correlation, final HttpRequest request,
+                      final HttpResponse response) throws IOException {
+
+        final ZonedDateTime startTime = correlation.getStart().atZone(timeZone);
+
+        // date
```
DaniloVeljovic, line 76: Do we have any reason why we choose to put these comments on top of the fields? If they are not needed, maybe we can remove them?

logbook-core/src/main/java/org/zalando/logbook/ExtendedLogFormatSink.java
whiskeysierra: Because this method is called upon request, but the format may include response fields which means all logging needs to be delayed until the response is received. Which is what the other `write` method is for.

logbook-core/src/main/java/org/zalando/logbook/ExtendedLogFormatSink.java
```diff
@@ -0,0 +1,245 @@
...
+                                .orElse(OMITTED_FIELD))
+                .orElse(OMITTED_FIELD);
+    }
+
+    /**
+     * Common supported fields
+     * @see <a href="https://docs.oracle.com/cd/A97329_03/bi.902/a90500/admin-05.htm#634823">Oracle analytical tool log
+     * parsing description</a>
+     */
+    private enum Field {
```
whiskeysierra, line 190: It would be nicer if this enum not only holds the field name, but also the logic (as a lambda) how to extract the information. That way the field name and its implementation would be very close to each other. It would also remove the need for the lcoal variables with the comments above.

logbook-spring/src/test/java/org/zalando/logbook/spring/RemoteResponseTest.java
greg65236592: It was removed due to the dependency upgrade. I have reverted this commit at https://github.com/zalando/logbook/pull/1304/commits/2a167380774f04f47b6cbded8dbbb934d360d1a6 to see if it can pass the build.

logbook-core/src/main/java/org/zalando/logbook/ExtendedLogFormatSink.java
greg65236592: Nice suggestion. Since the logic of getting field values requires data from several places, such as request and response, I have to encapsulate those objects into `FieldParameter` for the lambda usage, please see if that fits. https://github.com/zalando/logbook/pull/1304/commits/45435897f614a8d324e2bf0b1370fb79023f0571

greg65236592:
Thanks for reviewing! All comments should be addressed now.

DaniloVeljovic:
Looks good to me 👍🏼  Let's just fix the build, and we should be able to merge it now

greg65236592:
Looks like the issue: https://github.com/jeremylong/DependencyCheck/issues/5502 is not resolved yet, add a suppression to pass the build.

DaniloVeljovic:
👍🏼

PR #1320, opened 2022-04-29
Preserve case of reason phrase from the actual server response

andersjaensson:
## Description
Some servers respond with an upper case reason phrase. 

logbook-httpclient5 uses fixed reason phrase strings from [`org.zalando.logbook.HttpResponse`](https://github.com/zalando/logbook/blob/16ced8bfa91c59bd4bb758dd5a8adc10b5d5d545/logbook-api/src/main/java/org/zalando/logbook/HttpResponse.java#L19). 
This PR makes it take the reason phrase from the actual server response instead.

#### Example:
A request to https://httpbin.org/status/400
```
$ curl --head httpbin.org/status/400
HTTP/1.1 400 BAD REQUEST
```
will currently be logged like this by logbook:
```
HTTP/1.1 400 Bad Request
```
This change will make it log like this instead:
```
HTTP/1.1 400 BAD REQUEST
```

I guess this can possibly be considered a breaking change.

## Motivation and Context
I want the log to be as close to the actual server response as possible.

## Types of changes

## Checklist:


whiskeysierra (approved):

whiskeysierra:
@andersjaensson Can you rebase onto master to trigger a fresh build? The pipeline didn't run for some reason...

whiskeysierra:
👍

PR #1342, opened 2022-07-19
response.body() is nullable

coltonenglish:
## Description

OpenFeign's response object is nullable.  All that is required here is to propagate the null where appropriate.

## Motivation and Context

Notably a 404 with an empty response body trigger an NPE causing the library to blow up in ways that are unexpected.

## Types of changes

## Checklist:


hgabor83:
Lol, I've done a similar fix at #1360

kasmarian:
The fix is merged with https://github.com/zalando/logbook/pull/1362

Issue #1360, opened 2022-09-29
NPE at FeignLogbookLogger.java method logAndRebufferResponse if response body is empty


hgabor83:
## Description
If I got a response from feign client server, like 401 unauthorized with empty body, the logAndRebufferResponse throws an NPE here:
response.body().asInputStream()

## Expected Behavior
It should handle if there is empty body and skip the body checking part

## Actual Behavior
It always want to process body even if it is null

## Possible Fix
Check first if body is null, something like this:
byte[] body = response.body()!=null? ByteStreams.toByteArray(response.body().asInputStream()) : null;

## Steps to Reproduce
Have a response with empty body

## Context
N/A

## Your Environment
Spring Boot project

kasmarian:
This is released in 2.15.0

PR #1362, opened 2022-10-06
1360-NPE-at-FeignLogbookLogger

hgabor83:
1360-NPE-at-FeignLogbookLogger: if response body is empty then don't throw NPE, tests are fixed.

## Description
#1360 

## Motivation and Context
So if a response from a server comes without body then FeignLogbookLogger throws an NPE, instead of logging the correct response, like in my example 401 Unauthorized. With my fix we avoid NPE in such cases. The empty body related test was wrong I fixed it as well to test this new bugfix.

## Types of changes

## Checklist:


hgabor83:
@whiskeysierra or @lukasniemeier-zalando or @AlexanderYastrebov  could you please approve it? It's just a small modification, but would be very useful to fix this issue.

hgabor83:
Hi, could you please review this PR? We would need this essential fix as soon as possible. Thank you.

kasmarian:
Hi @hgabor83, sorry for the silence. Could you please merge main branch again to have a green build?

kasmarian:
:+1:

Issue #1112, opened 2021-08-16
Response decompression in Apache Http Client 5


aleksey-petrov-github:
There is known issue (#135) : The Logbook HTTP Client integration is handling gzip-compressed response entities incorrectly if the interceptor runs before a decompressing interceptor. Since logging compressed contents is not really helpful it's advised to register the logbook interceptor as the last interceptor in the chain.

This advice is not suitable for Apache Http Client 5, because [ResponseContentEncoding interceptor](https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/client/protocol/ResponseContentEncoding.java) now is implemented as [ContentCompressionExec handler](https://github.com/apache/httpcomponents-client/blob/master/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/ContentCompressionExec.java) which is called after interceptors.
So the logged content is compressed and also body filters are broken because they test compressed content.

For my project workaround is to implement strategy, that wraps response with decompressing logic:

```java
package a.b.c;

import org.apache.hc.client5.http.entity.DeflateInputStreamFactory;
import org.apache.hc.client5.http.entity.GZIPInputStreamFactory;
import org.apache.hc.client5.http.entity.InputStreamFactory;
import org.zalando.logbook.HttpRequest;
import org.zalando.logbook.HttpResponse;
import org.zalando.logbook.Strategy;

import java.io.IOException;
import java.util.Map;

public class ResponseDecompressingStrategy implements Strategy {

    private static final Map<String, InputStreamFactory> DECODER_REGISTRY = Map.of(
            "gzip", GZIPInputStreamFactory.getInstance(),
            "x-gzip", GZIPInputStreamFactory.getInstance(),
            "deflate", DeflateInputStreamFactory.getInstance());

    @Override
    public HttpResponse process(HttpRequest request, HttpResponse response) throws IOException {
        return new DecompressingHttpResponse(response.withBody(), DECODER_REGISTRY);
    }
}
```

```java
package a.b.c;

import lombok.AllArgsConstructor;
import lombok.NonNull;
import org.apache.hc.client5.http.entity.InputStreamFactory;
import org.apache.hc.core5.http.HttpHeaders;
import org.zalando.logbook.ForwardingHttpResponse;
import org.zalando.logbook.HttpResponse;

import java.io.ByteArrayInputStream;
import java.io.IOException;
import java.util.Collections;
import java.util.Map;
import java.util.Optional;

[...]

PR #1373, opened 2022-11-10
Fix issue 1112: apache http-client5 decompression

SimplicialCycle:
If the response is compressed the Interceptor will log at the wrong time, thus use an ExecHandler.

## Motivation and Context
This fixes Issue #1112 

## Types of changes

## Checklist:

### DIFF
--- a/logbook-httpclient5/src/main/java/org/zalando/logbook/httpclient5/LogbookHttpExecHandler.java
+++ b/logbook-httpclient5/src/main/java/org/zalando/logbook/httpclient5/LogbookHttpExecHandler.java
@@ -18,11 +18,11 @@ public class LogbookHttpExecHandler implements ExecChainHandler {
     }
 
     @Override
-    public ClassicHttpResponse execute(ClassicHttpRequest request, ExecChain.Scope scope, ExecChain execChain) throws IOException, HttpException {
-        LocalRequest localRequest = new LocalRequest(request, request.getEntity());
-        Logbook.ResponseProcessingStage stage = logbook.process(localRequest).write();
+    public ClassicHttpResponse execute(final ClassicHttpRequest request, final ExecChain.Scope scope, final ExecChain execChain) throws IOException, HttpException {
+        final LocalRequest localRequest = new LocalRequest(request, request.getEntity());
+        final Logbook.ResponseProcessingStage stage = logbook.process(localRequest).write();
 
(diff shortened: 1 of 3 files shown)


logbook-httpclient5/src/main/java/org/zalando/logbook/httpclient5/LogbookHttpExecHandler.java
```diff
@@ -0,0 +1,31 @@
...
+
+    private final Logbook logbook;
+
+    public LogbookHttpExecHandler(Logbook logbook) {
+        this.logbook = logbook;
+    }
+
+    @Override
+    public ClassicHttpResponse execute(ClassicHttpRequest request, ExecChain.Scope scope, ExecChain execChain) throws IOException, HttpException {
+        LocalRequest localRequest = new LocalRequest(request, request.getEntity());
```
DaniloVeljovic, line 22: Would it be possible to make the local variables and function parameters final?

logbook-httpclient5/src/test/java/org/zalando/logbook/httpclient5/LogbookHttpExecHandlerTest.java
```diff
@@ -0,0 +1,46 @@
...
+import java.io.IOException;
+import java.util.concurrent.ExecutionException;
+
+import static com.github.restdriver.clientdriver.RestClientDriver.giveResponse;
+import static com.github.restdriver.clientdriver.RestClientDriver.onRequestTo;
+import static org.apache.hc.core5.http.ContentType.TEXT_PLAIN;
+import static org.apache.hc.core5.http.HttpHeaders.CONTENT_TYPE;
+import static org.junit.jupiter.api.Assertions.*;
+
+class LogbookHttpExecHandlerTest extends AbstractHttpTest {
```
DaniloVeljovic, line 22: The description of the PR is `If the response is compressed the Interceptor will log at the wrong time, thus use an ExecHandler`, but I don't see any test case that would cover this use-case. Can you maybe add a few test cases to demonstrate this behaviour?

SimplicialCycle: I added a testcase for to see the logging of an compressed response after decompression

DaniloVeljovic:
👍🏼

PR #1400, opened 2023-01-31
Updating the dependencies and making the build pass

DaniloVeljovic:
## Description
The PR should update the dependencies with vulnerabilites and make the build pass

## Motivation and Context

The change will deblock the pipeline and update all the dependencies with vulnerabilites and make the build pass

## Types of changes

## Checklist:

### DIFF
--- a/logbook-parent/pom.xml
+++ b/logbook-parent/pom.xml
@@ -43,11 +43,11 @@
         <spring-boot1.version>1.5.22.RELEASE</spring-boot1.version>
         <spring-security4.version>4.2.20.RELEASE</spring-security4.version>
 
-        <spring5.version>5.3.16</spring5.version>
+        <spring5.version>5.3.25</spring5.version>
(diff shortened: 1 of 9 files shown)


logbook-parent/pom.xml
```diff
@@ -44,7 +44,6 @@
         <spring-security4.version>4.2.20.RELEASE</spring-security4.version>
 
         <spring5.version>5.3.25</spring5.version>
-        <spring-boot2.version>2.6.4</spring-boot2.version>
```
whiskeysierra, line 47: If this is meant to stop supporting Spring Boot 2, wouldn't it make sense to also drop support for Spring Boot 1 at this point (see above).

kasmarian: I'm about to bring the property back as it's still used by spring4 profile (that was a miss). But I'd suggest dropping the support of Spring Boot 1 overall.

kasmarian (approved):

PR #1407, opened 2023-02-09
End of support Spring Boot 1 and Spring Famework 4 

kasmarian:
## Description
As spring-boot 1.5 has reached it's commercial support on [2020-11-06](https://spring.io/projects/spring-boot#support), with these changes I propose to stop supporting it in Logbook as well.

## Motivation and Context
Support of older versions of Spring Framework and Spring Boot is becoming more problematic, it's a source of vulnerabilities, that can't be addressed, and number of potential users should be low at this point. 

Clients will still be able to use Logbook versions that support older versions of the Spring Framework.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        profile: ['', 'spring4']
+        profile: ['']
     steps:
     - name: Checkout
       uses: actions/checkout@v1
--- a/README.md
+++ b/README.md
@@ -36,8 +36,8 @@ Logbook is ready to use out of the box for most common setups. Even for uncommon
 - JAX-RS 2.x Client and Server (optional)
 - Netty 4.x (optional)
 - OkHttp 2.x **or 3.x** (optional)
-- Spring 4.x **or 5.x** (optional)
-- Spring Boot 1.x **or 2.x** (optional)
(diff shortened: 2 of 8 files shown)


kasmarian:
@whiskeysierra I've opened this PR to kick off the discussion if we really need to support old spring versions. 
After a short interaction with the library, I learned that spring4 related builds are more fragile ([latest example](https://github.com/zalando/logbook/pull/1298)), but I fail to understand if the time spent on fixing them is well spent, considering that end of support happened long time ago for Spring Boot 1. 
I'd appreciate your opinion here.

P.s. how do I get rid of `build (spring4)` status?

whiskeysierra:
> I'd appreciate your opinion here.

Yeah, support for Spring 4 and Spring Boot 1 should be discontinued for Logbook, imo.

> P.s. how do I get rid of `build (spring4)` status?

You need to modify the github action and then the build requirement (in the settings of this repo).

msdousti:
It will be great if we merge this sooner.
Regarding @whiskeysierra's comment below:

> and then the build requirement (in the settings of this repo).

I guess only authorized people can access `settings`:
image

kasmarian:
@lukasniemeier-zalando can you please assist with removing the `spring4` step from the build requirement?

kasmarian:
:+1:

otrosien:
👍

PR #1417, opened 2023-02-16
Migrate to Spring 6 / Spring Boot 3

### LINKED ISSUE #1344: Spring Framework 6 / Spring Boot 3 Support?

msdousti:
Bumping version of Spring to 6, Spring Boot to 3, and using `jakarta.*` instead of `javax.*` in most places. This is a breaking change, as it drops support for previous versions. See `Description` for further info.

## Description
Migrate Logbook to exclusively use Spring 6 / Spring Boot 3. That is, we drop support for Spring 5.x or Spring Boot 2.x.
It is yet to be decided whether we want to drop this support, so this PR is provided on a just-in-case basis.

## Motivation and Context
Fixes #1344

## Types of changes

## Checklist:
    - **Some changes to the documentation are required. They will be applied once we make sure that we want to proceed with this approach.**

### DIFF
--- a/logbook-spring-boot-autoconfigure/pom.xml
+++ b/logbook-spring-boot-autoconfigure/pom.xml
@@ -48,8 +48,8 @@
             <artifactId>logbook-servlet</artifactId>
         </dependency>
         <dependency>
-            <groupId>javax.servlet</groupId>
-            <artifactId>javax.servlet-api</artifactId>
+            <groupId>jakarta.servlet</groupId>
+            <artifactId>jakarta.servlet-api</artifactId>
         </dependency>
         <dependency>
             <groupId>org.zalando</groupId>
--- a/logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/ServerResponse.java
+++ b/logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/ServerResponse.java
@@ -1,7 +1,7 @@
 package org.zalando.logbook.spring.webflux;
 
 import lombok.RequiredArgsConstructor;
-import org.springframework.http.HttpStatusCode;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.server.reactive.ServerHttpResponse;
 import org.springframework.util.MimeType;
 import org.zalando.logbook.HttpHeaders;
@@ -34,8 +34,9 @@ final class ServerResponse implements HttpResponse {
 
     @Override
     public int getStatus() {
-        HttpStatusCode httpStatus = response.getStatusCode();
-        return httpStatus != null ? httpStatus.value() : 200;
+        return Optional.ofNullable(response.getStatusCode())
+                .orElse(HttpStatus.OK)
+                .value();
     }
 
     @Nullable
--- a/logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/ServerResponse.java
+++ b/logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/ServerResponse.java
@@ -1,6 +1,7 @@
 package org.zalando.logbook.spring.webflux;
 
 import lombok.RequiredArgsConstructor;
+import org.springframework.http.HttpStatusCode;
 import org.springframework.http.server.reactive.ServerHttpResponse;
 import org.springframework.util.MimeType;
 import org.zalando.logbook.HttpHeaders;
@@ -8,7 +9,6 @@ import org.zalando.logbook.HttpResponse;
 import org.zalando.logbook.Origin;
 
 import javax.annotation.Nullable;
-import java.io.IOException;
 import java.nio.charset.Charset;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
@@ -34,7 +34,8 @@ final class ServerResponse implements HttpResponse {
 
     @Override
     public int getStatus() {
-        return Optional.ofNullable(response.getRawStatusCode()).orElse(200);
+        HttpStatusCode httpStatus = response.getStatusCode();
+        return httpStatus != null ? httpStatus.value() : 200;
     }
 
     @Nullable
@@ -53,7 +54,7 @@ final class ServerResponse implements HttpResponse {
     }
 
     @Override
-    public HttpResponse withBody() throws IOException {
+    public HttpResponse withBody() {
         state.updateAndGet(State::with);
         return this;
     }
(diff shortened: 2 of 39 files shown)


logbook-spring/src/main/java/org/zalando/logbook/spring/RemoteResponse.java
```diff
@@ -105,7 +105,8 @@ private static final class Passing implements State {
     @Override
     public int getStatus() {
         try {
-            return response.getRawStatusCode();
+            // This may throw NPE
```
kasmarian, line 108: Should we fallback to the same status on every implementation? I wouldn't say I like that in some modules it falls back to 200, but could a common practice. Or we could fallback to `-1` as a way to say "no data" provided.

I can't say I'm a fan of the idea that a logging library can fail requests because one of the attributes (even the one that's probably always there) is missing.

logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/ServerResponse.java
```diff
@@ -34,7 +34,9 @@ public HttpHeaders getHeaders() {
 
     @Override
     public int getStatus() {
-        return Optional.ofNullable(response.getRawStatusCode()).orElse(200);
+        return Optional.ofNullable(response.getStatusCode())
```
kasmarian, line 37: I think you missed one implementation in [org.zalando.logbook.spring.webflux.ClientResponse](https://github.com/zalando/logbook/blob/6b1168bbd922f02058344ab436c2d2230a0fd97d/logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/ClientResponse.java#L27)

kasmarian:
Checked on a simple webflux app and works like a charm :partying_face:

logbook-spring-boot-autoconfigure/pom.xml
```diff
@@ -48,8 +48,8 @@
             <artifactId>logbook-servlet</artifactId>
         </dependency>
         <dependency>
-            <groupId>javax.servlet</groupId>
-            <artifactId>javax.servlet-api</artifactId>
+            <groupId>jakarta.servlet</groupId>
+            <artifactId>jakarta.servlet-api</artifactId>
```
kasmarian, line 52: One thing I missed though, I think you need to add the `org.springframework.boot.autoconfigure.AutoConfiguration.imports` file with `org.zalando.logbook.autoconfigure.LogbookAutoConfiguration` in it for autoconfiguration to work. As per [this comment in another PR](https://github.com/zalando/logbook/pull/1375#discussion_r1027269042)

logbook-spring/src/main/java/org/zalando/logbook/spring/RemoteResponse.java
msdousti: We should definitely discuss it. I also thought about changing the interface to return `Integer` instead of `int`, but didn't consider its implications.

logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/ServerResponse.java
msdousti: Thanks, fixed. I added a `FIX-ME` comment to keep track of locations where we should discuss how to avoid NPE.

logbook-spring-boot-autoconfigure/pom.xml
msdousti: Thanks, fixed: https://github.com/zalando/logbook/pull/1417/commits/a39e77c964f75f0224801d1b19e23ac2e3e76e75

logbook-spring/src/main/java/org/zalando/logbook/spring/RemoteResponse.java
msdousti: FYI: Tried my approach (`Integer` instead of `int`) and failed: One of the classes implements a Spring interface which has `public int getStatus()`, and extends a class which would then have `public Integer getStatus()`, and bam! clash.

I don't like default `200` either (because it can be misleading), and would like to see some fallback like `-1` as you suggested.

darrenskywalker:
Any ETA or update on when this PR might be merged in? I'm really looking forward to being able to use this library in SpringBoot 3 / Spring 6

msdousti:
@darrenskywalker

> Any ETA or update on when this PR might be merged in? I'm really looking forward to being able to use this library in SpringBoot 3 / Spring 6

Thanks a lot for the interest in Logbook! We are considering several approaches to add this support. While two working PRs exist, the decision in which way to go is yet to be made.

We need some time to discuss this and make the best decision to support our users.

Please have a look at the discussion I opened here:
https://github.com/zalando/logbook/discussions/1426

darrenskywalker:
> @darrenskywalker
> 
> > Any ETA or update on when this PR might be merged in? I'm really looking forward to being able to use this library in SpringBoot 3 / Spring 6
> 
> Thanks a lot for the interest in Logbook! We are considering several approaches to add this support. While two working PRs exist, the decision in which way to go is yet to be made.
> 
> We need some time to discuss this and make the best decision to support our users.
> 
> Please have a look at the discussion I opened here: #1426

Thanks @msdousti , that makes sense and thanks for the explanation and update.

owebb98:
> Checked on a simple webflux app and works like a charm 🥳

Hi! How did you do the build to test this? I've cloned the repo and checked out this branch but unit tests + build are failing via IntelliJ Ultimate. I can skip the tests but the build fails with "package org.zalando.logbook.httpclient5 does not exist" @ logbook-spring-boot-autoconfigure\src\main\java\org\zalando\logbook\autoconfigure\LogbookAutoConfiguration.java:336:44

Please let me know if you came across this during your testing. 

Cheers!

hemju:
@msdousti could you please give us an ETA when this will be fixed? I do understand that there needs to be decisions made, but the discussions started months ago and Spring Boot 3 was in November (with years of knowledge that javax will be removed).

msdousti:
I'm closing this branch as the team decided against it: Logbook will keep support for older version of Spring, and add support for the new version as well.

As for an ETA, I unfortuantely can't make an estimation. I know it's frustrating, but I ask for your understanding.

PR #1440, opened 2023-02-25
Upgrade to Ktor 2.x

msdousti:
## Types of changes

## Checklist:

### DIFF
--- a/cve-suppressions.xml
+++ b/cve-suppressions.xml
@@ -1,15 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
-        <suppress>
-            <cve>CVE-2021-37533</cve>
-            <!-- Suppressing until the migration to Spring Framework 6-->
-            <cve>CVE-2016-1000027</cve>
-            <!-- False positive -->
-            <cve>CVE-2020-5408</cve>
-            <!-- Okhttp2 module uses the latest version of okhttp. The vulnerability is in OkHostnameVerifier.java, which is not used in this library -->
-            <cve>CVE-2021-0341</cve>
-            <!-- ktor requires a major upgrade. Suppressing until then -->
-            <cve>CVE-2021-4277</cve>
-        </suppress>
-
+    <suppress>
+        <cve>CVE-2021-37533</cve>
+        <!-- Suppressing until the migration to Spring Framework 6-->
+        <cve>CVE-2016-1000027</cve>
+        <!-- False positive -->
+        <cve>CVE-2020-5408</cve>
+        <!-- Okhttp2 module uses the latest version of okhttp. The vulnerability is in OkHostnameVerifier.java, which is not used in this library -->
+        <cve>CVE-2021-0341</cve>
+        <!-- ktor requires a major upgrade. Suppressing until then -->
+        <cve>CVE-2021-4277</cve>
+        <!-- jackson-core-2.14.2 and json-path-2.7.0; waiting for new versions -->
+        <cve>CVE-2022-45688</cve>
+    </suppress>
 </suppressions>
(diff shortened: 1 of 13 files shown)


logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/LogbookServer.kt
```diff
@@ -27,22 +28,24 @@ class LogbookServer(
...
             pipeline.receivePipeline.intercept(ApplicationReceivePipeline.Before) {
                 val request = ServerRequest(call.request)
-                val requestWritingStage = feature.logbook.process(request)
+                val requestWritingStage = plugin.logbook.process(request)
                 val proceedWith = when {
                     request.shouldBuffer() && !call.request.receiveChannel().isClosedForRead -> {
                         val content = call.request.receiveChannel().readBytes()
                         request.buffer(content)
-                        ApplicationReceiveRequest(it.typeInfo, ByteReadChannel(content))
+                        ByteReadChannel(content)
```
msdousti, line 46: This line needs revision.
I just put something here so that the class compiles and the tests pass.

logbook-ktor-client/src/main/kotlin/org/zalando/logbook/client/LogbookClient.kt
```diff
@@ -29,40 +29,43 @@ class LogbookClient(
...
+                val responseWritingStage = responseProcessingStage.process(clientResponse)
+                if (clientResponse.shouldBuffer() && !loggingContent.isClosedForRead) {
                     val content = loggingContent.readBytes()
-                    response.buffer(content)
+                    clientResponse.buffer(content)
                 }
                 responseWritingStage.write()
 
-                val proceedWith = context.wrapWithContent(responseContent).response
+                val proceedWith = httpResponse.call.wrapWithContent(responseContent).response
```
msdousti, line 68: This line needs revision.
I just put something here so that the class compiles and the tests pass.

kasmarian: I'm also not an expert in Ktor, but this change seems right :thinking: . If you compare where `context.wrapWithContent` was used in 1.6, now this construct is used by the library itself. e.g. in `io.ktor.client.features.observer.ResponseObserver`

kasmarian: I'm also not an expert in Ktor, but this change seems right :thinking: . If you compare where `context.wrapWithContent` was used in 1.6, now this construct is used by the library itself. e.g. in `io.ktor.client.features.observer.ResponseObserver`

logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/LogbookServer.kt
ChristianLohmann: plugins work different these days. I created a new branch from the current main. I also try to adapt the tests to the new API

kasmarian: This also seems right according to the migration docs - https://ktor.io/docs/migrating-2.html#serialization-custom-converter

kasmarian: Or I misunderstood it 🤔

cve-suppressions.xml
```diff
@@ -1,15 +1,16 @@
...
+    <suppress>
+        <cve>CVE-2021-37533</cve>
+        <!-- Suppressing until the migration to Spring Framework 6-->
+        <cve>CVE-2016-1000027</cve>
+        <!-- False positive -->
+        <cve>CVE-2020-5408</cve>
+        <!-- Okhttp2 module uses the latest version of okhttp. The vulnerability is in OkHostnameVerifier.java, which is not used in this library -->
+        <cve>CVE-2021-0341</cve>
+        <!-- ktor requires a major upgrade. Suppressing until then -->
+        <cve>CVE-2021-4277</cve>
```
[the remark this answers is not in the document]
msdousti: This is now greatly simplified: Only one CVE remaining 🙂

kasmarian:
:+1:

PR #1441, opened 2023-02-25
Add Coverage Profile To Reactor Pom

ChristianLohmann:
Using a coverage profile in the build action to ensure coveralls plugin runs with provided jaxb-api


kasmarian:
:+1:

Issue #1199, opened 2021-10-21
Logbook will change response header  during pure servlet project


wyzssw:
LogbookFilter flush buffer will make  tomcat add `Transfer-Encoding: chunked` header

## Description
In pure servlet project , no spring mvc , no spring boot
`LogbookFilter`
```

  private void write(RemoteRequest request, LocalResponse response, ResponseWritingStage writing) throws IOException {
        final AtomicBoolean attribute = (AtomicBoolean) request.getAttribute(responseWritingStageSynchronizationName);
        if (!attribute.getAndSet(true)) {
            response.flushBuffer(); // <---- will triger tomcat  org.apache.coyote.http11.AbstractHttp11Processor#prepareResponse which will add header Transfer-Encoding: chunked but not content-length
            writing.write();
        }
    }

```

## Expected Behavior
```
curl -i 'http://127.0.0.1:8089/servlettest/helloworld'
HTTP/1.1 200
Content-Length: 10
Date: Thu, 21 Oct 2021 08:29:25 GMT

helloworld
```

## Actual Behavior

```
curl -i 'http://127.0.0.1:8089/servlettest/helloworld'
HTTP/1.1 200
Transfer-Encoding: chunked
Date: Thu, 21 Oct 2021 08:38:23 GMT

helloworld
```
It is compatible to  use standard http client ，but there will be problems when use raw http parser which only support content-length 
 
## Possible Fix
Refer to Springboot
`org.springframework.http.converter.AbstractHttpMessageConverter#addDefaultHeaders`
add default headers include `Content-Length`

## Steps to Reproduce

1.  create a pure servlet project
2. import logbook dependency, add logbookfilter to web.xml 
3. add a servlet class
```
public class HelloworldServlet extends HttpServlet {

    protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException {
        resp.getWriter().write("helloworld");
    }
}
```
4. mvn clean package to project.war ,add copy to tomcat/webapps/
5. bin/catalina.sh start 

## Context
My team has a old  pure servlet project , which not depend on springmvc。Some raw c++ http client cannot recognize the `Transfer-Encoding` header ，can not parse response body without `Content-Length`

## Your Environment

* Version used: Logbook 2.14.0 ,  Tomcat 8.0
* Link to your project: not upload

nhmarujo:
[...]

PR #1476, opened 2023-04-01
Fix: The underlying HTTP client completed without emitting a response

kasmarian:
## Description
LogbookExchangeFilterFunction was not handing case of empty response with `Transfer-Encoding: chunked`. This should not happen, but it happens. This resulted in WebClient throwing an error as ExchangeFunction returned an empty Mono.

The issue was not reproducible without Wiremock as Spring would always return `Content-Length: 0` with empty response. To add tests for the issue, I rewrote LogbookExchangeFilterFunctionTest to use WireMock.

Note: I added wiremock version 3.0.0-beta-7, which supports jakarta servlet api. Versions lower than 3 were resulting in application startup errors.

## Motivation and Context

Addresses the issue mentioned in [this comment](https://github.com/zalando/logbook/issues/1199#issuecomment-1466977403) in issue 1199, but not the initial issue. 

## Types of changes

## Checklist:

### DIFF
--- a/logbook-parent/pom.xml
+++ b/logbook-parent/pom.xml
@@ -263,6 +263,12 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+            <dependency>
+                <groupId>com.github.tomakehurst</groupId>
+                <artifactId>wiremock</artifactId>
+                <version>3.0.0-beta-7</version>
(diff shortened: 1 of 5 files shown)


kumarprabhashanand:
👍

PR #1477, opened 2023-04-03
Introduce JDK9 modules

### LINKED ISSUE #625: Enable JDK9+ module-info
Add module info OR default module name in manifest, so that it is possible to use this project as a dependency in a project using modules (JDK9+).

## Detailed Description
It should be possible to use this project as a dependency in a JDK 9+ project, in other words, at a minimum the module package name must be specified. For the module info case, creating a multi-release jar using the 'usual suspects' of maven plugins is currently quite hard / painful. So staying on JDK 8 and supporting 9+ via some plugins seems like the best use of time.

## Possible Implementation
Use moditect-maven-plugin with 

```
<plugin>
    <groupId>org.moditect</groupId>
    <artifactId>moditect-maven-plugin</artifactId>
    <version>1.0.0.Beta2</version>
    <executions>
        <execution>
            <id>add-module-infos</id>
            <phase>package</phase>
            <goals>
                <goal>add-module-info</goal>
            </goals>
            <configuration>
                <jvmVersion>${java.module.version}</jvmVersion>
                <overwriteExistingFiles>true</overwriteExistingFiles>
                <module>
                    <moduleInfoFile>src/main/moditech/module-info.java</moduleInfoFile>
                </module>
            </configuration>
        </execution>
    </executions>
</plugin>
```

or maven-jar-plugin with

```
<plugin>
	<groupId>org.apache.maven.plugins</groupId>
	<artifactId>maven-jar-plugin</artifactId>
	<version>3.1.2</version>
	<configuration>
		<archive>
			<manifestEntries>
			    <Automatic-Module-Name>x.y.z</Automatic-Module-Name>
			</manifestEntries>
		</archive>
	</configuration>
</plugin>
```

## Your Environment
Java 8.

msdousti:
To support JDK9 modules:
- [x] Each Maven module should have its unique package name.
- [ ] A Maven plugin can be used to add package info to generated JAR files.

## Motivation and Context
Closes #625

## Types of changes

## Checklist:


README.md
```diff
@@ -254,19 +254,19 @@ respectively (in conjunction with `ForwardingHttpRequest`/`ForwardingHttpRespons
...
-    .queryFilter(replaceQuery("password", "<secret>"))
-    .headerFilter(authorization()) 
-    .headerFilter(eachHeader("X-Secret"::equalsIgnoreCase, "<secret>"))
-    .build();
+import static org.zalando.logbook.core.HeaderFilters.authorization;
+import static org.zalando.logbook.core.HeaderFilters.eachHeader;
+import static org.zalando.logbook.core.QueryFilters.accessToken;
+import static org.zalando.logbook.core.QueryFilters.replaceQuery;
+
+Logbook logbook=Logbook.builder()
```
ChristianLohmann, line 262: some weird formatting happened here

ChristianLohmann: fixed it already

.gitignore
```diff
@@ -16,3 +16,4 @@ target/
 
 ### various ###
 jmh-result.json
+dependency-reduced-pom.xml
```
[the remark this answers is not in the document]
kasmarian: I'm also getting this file and it annoys me :). The `maven-shade-plugin` adds it during the build. See [maven shade docs](https://maven.apache.org/plugins/maven-shade-plugin/shade-mojo.html#createDependencyReducedPom)

logbook-api/src/main/java/org/zalando/logbook/api/RequestURI.java
```diff
@@ -7,35 +7,35 @@
...
-import static org.zalando.logbook.RequestURI.Component.QUERY;
-import static org.zalando.logbook.RequestURI.Component.SCHEME;
+import static org.zalando.logbook.api.RequestURI.Component.AUTHORITY;
+import static org.zalando.logbook.api.RequestURI.Component.PATH;
+import static org.zalando.logbook.api.RequestURI.Component.QUERY;
+import static org.zalando.logbook.api.RequestURI.Component.SCHEME;
 
 @UtilityClass
-final class RequestURI {
+public final class RequestURI {
```
kasmarian, line 16: Looks like this class and `LogbookFactory` are becoming public. Not sure if we should be concerned about it. I wonder if it would be possible to hide internal classes like this one behind the java 9 modules. Would it be compatible with java 8 somehow? Does it make sense to do it?

msdousti:
I temporarily enabled Java 17 as source & target, to have a glimpse of the future and see modules in practice.
It will be set back to 1.8 later on.

logbook-api/src/main/java/org/zalando/logbook/api/RequestURI.java
msdousti: We can use Java 9 modules to expose classes of one package (e.g., `org.zalando.logbook.internal`) to specific modules only (e.g., `core` module). This still requires the classes to be **public**. Also, it won't work with Java 8 as it does not recognize modules.

More explanation from ChatGPT 🙂

image

kasmarian:
:+1:

PR #1498, opened 2023-05-15
Make servlet-api optional 

ChristianLohmann:
set the servlet-api to optional in order to be still able to compile


msdousti:
👍

ChristianLohmann:
👍

Issue #1497, opened 2023-05-11
UnsupportedClassVersionError when running on Java 8


hiji:
## Description

In a Java 8 environment, registering a Logbook with the JAX-RS client (implemented in Jersey) and sending a request generated an UnsupportedClassVersionError.

In the README, following is written as Dependencies.
> Java 8 (for Spring 6 / Spring Boot 3 Java 17 is required)

After investigating the cause, it seems that `Paguro:3.10.3`, on which `logbook-api` depends, is compiled with Java 11.

Is Java 11 required to use Logbook?

```
java.lang.UnsupportedClassVersionError: org/organicdesign/fp/collections/ImSortedMap has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:756)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:473)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:74)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:369)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
	at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClassFromSelf(ClassRealm.java:425)
	at org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy.loadClass(SelfFirstStrategy.java:42)
	at org.codehaus.plexus.classworlds.realm.ClassRealm.unsynchronizedLoadClass(ClassRealm.java:271)
	at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass(ClassRealm.java:247)
	at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass(ClassRealm.java:239)
	at org.zalando.logbook.HttpHeaders.empty(HttpHeaders.java:79)
	at org.zalando.logbook.HttpHeaders.of(HttpHeaders.java:87)
	at org.zalando.logbook.jaxrs.LocalRequest.getHeaders(LocalRequest.java:145)
	at org.zalando.logbook.Cache.lambda$get$0(Cache.java:16)
[...]

PR #1499, opened 2023-05-15
Remove Paguro dependency from the project

msdousti:
closes https://github.com/zalando/logbook/issues/1497

### DIFF
--- a/logbook-api/src/main/java/org/zalando/logbook/DefaultHttpHeaders.java
+++ b/logbook-api/src/main/java/org/zalando/logbook/DefaultHttpHeaders.java
@@ -4,7 +4,6 @@ import lombok.AllArgsConstructor;
 import lombok.With;
 import lombok.experimental.Delegate;
 
-import java.util.AbstractList;
 import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -56,10 +55,7 @@ final class DefaultHttpHeaders
     }
 
     public static <T> List<T> immutableCopy(final Collection<T> values) {
-        if (values instanceof ImListWithToString) {
-            return (List<T>) values;
-        }
-        return new ImListWithToString<>(Collections.unmodifiableList(new ArrayList<>(values)));
+        return Collections.unmodifiableList(new ArrayList<>(values));
     }
 
     private static <K, V> TreeMap<K, V> associate(final TreeMap<K, V> original, final K key, final V value) {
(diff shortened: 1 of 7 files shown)


logbook-api/src/main/java/org/zalando/logbook/DefaultHttpHeaders.java
```diff
@@ -43,34 +44,38 @@ public HttpHeaders update(
...
+        Comparator<? super K> cmp = original.comparator();
+        Stream<Map.Entry<K, V>> tmpStream = Stream.of(new SimpleEntry<>(key, value));
+
+        return Stream.concat(original.entrySet().stream(), tmpStream)
+                .collect(Collectors.toMap(
+                        Map.Entry::getKey,
+                        entry -> cmp.compare(key, entry.getKey()) == 0 ? value : entry.getValue(),
+                        (v1, v2) -> v2,
+                        () -> new TreeMap<>(cmp)
+                ));
```
kasmarian, line 71: Would this be the same as this? 

```suggestion
        TreeMap<K, V> tmpMap = new TreeMap<>(original);
        tmpMap.put(key, value);
        return tmpMap;
```

whiskeysierra: I believe the first version is a complicated way of saying `tmpMap.putIfAbsent(key, value)` instead of just `put`.

whiskeysierra: Ah, nevermind. I misread the old code. I think you're right.

msdousti: @kasmarian Yes, you are right. I'll simplify it and push again once I have time today :)

msdousti: @kasmarian The fix is applied now and I also simplified a bit here and there to make it more efficient. Thanks a lot for your suggestion. Would you mind reviewing again? 🙏

logbook-api/src/main/java/org/zalando/logbook/DefaultHttpHeaders.java
```diff
@@ -43,34 +43,43 @@ public HttpHeaders update(
...
+            final TreeMap<String, List<String>> original,
+            final Collection<String> keys) {
+
+        if (Collections.disjoint(original.keySet(), keys))
+            return original;
 
+        Comparator<String> cmp = String.CASE_INSENSITIVE_ORDER;
+        TreeMap<String, List<String>> modifiedMap = new TreeMap<>(original);
+        modifiedMap.keySet().removeIf(k -> keys.stream().anyMatch(key -> cmp.compare(k, key) == 0));
+        return modifiedMap;
```
kasmarian, line 82: Sorry, maybe it's just me, but wouldn't it be simpler to have something like this? The whole direct comparator usage seems unnecessary to my taste.
```suggestion
        TreeMap<String, List<String>> modifiedMap = new TreeMap<>(original);
        keys.forEach(modifiedMap::remove);
        return modifiedMap;
```

msdousti: Thanks, this is a great suggestion - The simpler, the better!
I applied it.

kasmarian:
:+1:

Issue #1500, opened 2023-05-17
JsonPath body filter does not obfuscate due to unexisting field in request/response


paswd:
## Description
In my project I use JsonPathBodyFilters to obfuscate json properties in several responses. These properties are unique for each response. So, at least one BodyFilter throws ClassCastException as in #1369.

P.S. I catch this issue in 2.15.0, but, as I see in issues and source code, it has't been fixed yet.

## Actual Behavior
Result of this bug is totally not obfuscated response.

## Expected Behavior
I think that correct behavior is to ignore unexisting field and obfuscate another correct fields. Or at least make it optional.

## Possible Fix
I cannot expand or redefine existing **JsonPathBodyFilters** due to most of it's classes and methods are private static. But I made for me a working copy of it.

In class **CompositeOperation** we have method _filter_.

It's current implementation:
```
@Override
public DocumentContext filter(final DocumentContext context) {
    DocumentContext result = context;

    for (final Operation operation : operations) {
        result = operation.filter(result);
    }

    return result;
}
```

My implementation:
```
@Override
public DocumentContext filter(final DocumentContext context) {
    DocumentContext result = context;

    for (final Operation operation : operations) {
        try {
            result = operation.filter(result);
        } catch (Exception ex) {
            // maybe logging or smth else
        }
    }

    return result;
}
```

So, single operation exception does not break another filters anymore.

whiskeysierra:
> Result of this bug is totally not obfuscated response.

If that is the current behavior, then there is already a place in the code somewhere that catches and ignores exceptions in body filters. Have you found it?

paswd:
> > Result of this bug is totally not obfuscated response.
> 
> If that is the current behavior, then there is already a place in the code somewhere that catches and ignores exceptions in body filters. Have you found it?

Yes, I found it. This place is in **JsonPathBodyFilter.filter**:
```
@Override
public String filter(
        @Nullable final String contentType, final String body) {

    if (body.isEmpty() || !JSON.test(contentType)) {
        return body;
    }

    try {
        final DocumentContext original = CONTEXT.parse(body);
        // BodyFilter exception throws here
        return operation.filter(original).jsonString();
    } catch (Exception e) {
        log.trace("The body could not be filtered, the following exception {} has been thrown", e.getClass());
        // The place where returns not obfuscated response
        return body;
    }
}
```

kasmarian:
@paswd Thank you for pinpointing to the concrete bug location, the fix should be available in the next release.

Issue #1369, opened 2022-10-20
JacksonJsonNodeJsonProvider throws class cast exceptions


ahaverdings-deloitte:
https://github.com/zalando/logbook/blob/bc4afe1b12215f32197449b54ae8f96e6c44be55/logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java#L92

This JacksonJsonNodeJsonProvider throws class cast exceptions:
class com.fasterxml.jackson.databind.node.ArrayNode cannot be cast to class java.util.List (com.fasterxml.jackson.databind.node.ArrayNode is in unnamed module of loader 'app'; java.util.List is in module java.base of loader 'bootstrap')

Resulting that the body is never transformed when there is one rule applied that handels arrays. Once the exception is throw the original body is returned (and this line also hides the exception pretty well...)
https://github.com/zalando/logbook/blob/786791737dfcf8e83daa7152e381836eeda254cd/logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java#L112-L113

Issue is that the expected type of (T) would be List<String> but JacksonJsonNodeJsonProvider on createArray() is returning an ArrayNode which is not castable to List.

A solution would be to use JacksonJsonProvider instead of JacksonJsonNodeJsonProvider

whiskeysierra:
Where is this cast to a List happening?

ahaverdings-deloitte:
The cast itself is implicit and part of `JsonPath` (and maybe not much to do about it):
https://github.com/json-path/JsonPath/blob/3f4900d6e7361cd21f0327c828dabf83fd188227/json-path/src/main/java/com/jayway/jsonpath/internal/JsonContext.java#L131
> then:
https://github.com/json-path/JsonPath/blob/3f4900d6e7361cd21f0327c828dabf83fd188227/json-path/src/main/java/com/jayway/jsonpath/JsonPath.java#L225
> then:
https://github.com/json-path/JsonPath/blob/3f4900d6e7361cd21f0327c828dabf83fd188227/json-path/src/main/java/com/jayway/jsonpath/JsonPath.java#L752


the line it starts is here:
https://github.com/zalando/logbook/blob/786791737dfcf8e83daa7152e381836eeda254cd/logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java#L58

whiskeysierra:
[...]

PR #1514, opened 2023-05-29
Do not fail all composite body filters if one fails

kasmarian:
## Description
If one out of several body filters fails (for example because of the issue mentioned in https://github.com/zalando/logbook/issues/1369), it should not discard the results of other filters.

## Motivation and Context
addresses https://github.com/zalando/logbook/issues/1500

## Types of changes

## Checklist:

### DIFF
--- a/logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
@@ -154,7 +154,11 @@ public final class JsonPathBodyFilters {
             DocumentContext result = context;
 
             for (final Operation operation : operations) {
-                result = operation.filter(result);
+                try {
+                    result = operation.filter(result);
+                } catch (Exception e) {
+                    log.trace("Composite filter operation could not complete, the following exception {} has been thrown", e.getClass());
+                }
             }
 
             return result;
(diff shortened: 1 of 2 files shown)


logbook-json/src/main/java/org/zalando/logbook/json/JsonPathBodyFilters.java
```diff
@@ -154,7 +154,11 @@ public DocumentContext filter(final DocumentContext context) {
             DocumentContext result = context;
 
             for (final Operation operation : operations) {
-                result = operation.filter(result);
+                try {
+                    result = operation.filter(result);
+                } catch (Exception e) {
+                    log.trace("Composite filter operation could not complete, the following exception {} has been thrown", e.getClass());
+                }
```
msdousti, line 161: This is nice in that it handles the issue. I have two suggestions to improve it:

1. Adding all exceptions to a list, and logging them all at once (if the list is non-empty). This is useful in that it won't clutter the logs, should there be multiple exceptions.
2. If the user is expected to understand what "composite filter" means, it's fine. Otherwise, something about `JsonPath Body Filter` can be used instead.

kasmarian: Thank you for the suggestion, added this in https://github.com/zalando/logbook/pull/1514/commits/4fb418f4d4e1dca5829fd239a4270f7da4c264d4

msdousti:
👍

kasmarian:
:+1:

Issue #870, opened 2020-10-24
The log contains Chinese, which is not displayed correctly.


tian-jeff:
## Description

When print Chinese in Spring Boot projects, then looks like not right, for example:
```json
{"code":45,"name":"å¹¿è¥¿å£®æèªæ²»åº"}
```
## Expected Behavior

Print Chinese normal, like this:
```json
{"code":45,"name":"广西壮族自治区"}
```
## Actual Behavior

```json
{"code":45,"name":"å¹¿è¥¿å£®æèªæ²»åº"}
```
## Possible Fix

Must force set utf-8 encoding
```yaml
server:
  servlet:
    encoding:
      enabled: true
      charset: UTF-8
      force: true
```
## The root cause of this problem
![image](https://user-images.githubusercontent.com/1702836/97088366-dae55200-1662-11eb-8047-a4ed0d7ad3bc.png)
> The priority order for specifying the response body is:
> 1. explicitly per request using setCharacterEncoding and setContentType
> 2. implicitly per request using setLocale
> 3. per web application via the deployment descriptor or ServletContext.setRequestCharacterEncoding(String)
> 4. container default via vendor specific configuration
> 5. ISO-8859-1

If any possible, get encoding form somewhere else.
![image](https://user-images.githubusercontent.com/1702836/97088562-45e35880-1664-11eb-8115-1e6c16b56849.png)

## Your Environment

* Version used: ```org.zalando:logbook-spring-boot-starter:2.3.0``` & Spring Boot Ver: 2.3.4.RELEASE

whiskeysierra:
> ISO-8859-1

Is that the fallback from Tomcat or defined in the Servlet Spec?

tian-jeff:
Thx for you reply, I think this is fallback from Tomcat.

whiskeysierra:
> I think this is fallback from Tomcat.

Ok, if that is in fact the case (could you verify that somehow?), then I see the following options:

1. We leave it as it is. Works for UTF-8, but breaks for everything else. Bad user experience for some users and no possible workaround.
2. We leave UTF-8 as our default fallback, but we make it configurable. Either with a system property or something alike. It will work for most users out of the box and those that need something different have something to work around the issue.
3. We could try to detect in which servlet container we're actually running (inspecting class names maybe?) and change our fallback charset accordingly.

[...]

PR #1521, opened 2023-05-30
Use UTF-8 as a fallback for JSON content type

kasmarian:
## Motivation and Context
According to RFC 8259 JSON uses UTF-8 for encoding messages. 

In this changeset, I'm using a common fallback to UTF-8 when determining a Charset for the HttpMessage, if no explicit Charset is defined via Content Type header (is this the right approach for responses as well?)

This change also makes `ContentType` utility class public and moves `JsonMediaType.java` from `logbook-json` module to `logbook-api` module. 

Was inspired by discussions in https://github.com/zalando/logbook/issues/870

## Types of changes

## Checklist:


logbook-api/src/main/java/org/zalando/logbook/internal/JsonMediaType.java
```diff
@@ -1,4 +1,4 @@
-package org.zalando.logbook.json;
+package org.zalando.logbook.internal;
```
msdousti, line 1: `org.zalando.logbook.internal` package is already defined in the `logbook-api` module. I wish I could move everything in that package to `org.zalando.logbook.api`, but it would have been a very severe breaking change.

In order to be Java 9 module compatible, let's keep packages separate. Here I suggest using the following package instead:
```suggestion
package org.zalando.logbook.json.internal;
```

logbook-api/src/test/java/org/zalando/logbook/internal/JsonMediaTypeTest.java
```diff
@@ -1,20 +1,14 @@
-package org.zalando.logbook.json;
+package org.zalando.logbook.internal;
```
msdousti, line 1:
```suggestion
package org.zalando.logbook.json.internal;
```

logbook-api/src/test/java/org/zalando/logbook/ContentTypeTest.java
```diff
@@ -15,11 +15,15 @@ void unitTest() {
...
         assertThat(ContentType.parseCharset("charset=;")).isNull();
         assertThat(ContentType.parseCharset("charset=foo\"")).isNull();
         assertThat(ContentType.parseCharset("charset=\"bar")).isNull();
         assertThat(ContentType.parseCharset("application/json; charset=\"us-ascii\"")).isEqualTo(StandardCharsets.US_ASCII);
     }
+
+    @Test
+    void fallbackToUTF8WhenJson() {
+        assertThat(ContentType.parseCharset("application/json; charset=")).isEqualTo(StandardCharsets.UTF_8);
+        assertThat(ContentType.parseCharset("application/json; charset=unknown-charset")).isEqualTo(StandardCharsets.UTF_8);
```
msdousti, line 27: - Is the same logic applied if the content type is, e.g., `application/problem+json`? If so, may I suggest adding a test for it as well?
- Also, can we add a test to make sure content-type matching is **not** case sensitive? (both for header name `Content-Type:`/`content-type:` and its value `Application/JSON` / `application/json`)

logbook-jmh/src/main/java/org/zalando/logbook/benchmark/json/JsonMediaTypeBenchmark.java
```diff
@@ -10,7 +10,7 @@
 import org.openjdk.jmh.runner.RunnerException;
 import org.openjdk.jmh.runner.options.Options;
 import org.openjdk.jmh.runner.options.OptionsBuilder;
-import org.zalando.logbook.json.JsonMediaType;
+import org.zalando.logbook.internal.JsonMediaType;
```
msdousti, line 13:
```suggestion
import org.zalando.logbook.json.internal.JsonMediaType;
```

logbook-json/src/main/java/org/zalando/logbook/json/CompactingJsonBodyFilter.java
```diff
@@ -4,6 +4,7 @@
 import lombok.extern.slf4j.Slf4j;
 import org.apiguardian.api.API;
 import org.zalando.logbook.BodyFilter;
+import org.zalando.logbook.internal.JsonMediaType;
```
msdousti, line 7:
```suggestion
import org.zalando.logbook.json.internal.JsonMediaType;
```

logbook-json/src/main/java/org/zalando/logbook/json/FastCompactingJsonBodyFilter.java
```diff
@@ -3,6 +3,7 @@
 import lombok.extern.slf4j.Slf4j;
 import org.apiguardian.api.API;
 import org.zalando.logbook.BodyFilter;
+import org.zalando.logbook.internal.JsonMediaType;
```
msdousti, line 6:
```suggestion
import org.zalando.logbook.json.internal.JsonMediaType;
```

logbook-json/src/main/java/org/zalando/logbook/json/FastJsonHttpLogFormatter.java
```diff
@@ -11,6 +11,7 @@
 import org.zalando.logbook.HttpRequest;
 import org.zalando.logbook.HttpResponse;
 import org.zalando.logbook.Precorrelation;
+import org.zalando.logbook.internal.JsonMediaType;
```
msdousti, line 14: (Same suggestion for other parts; I just looked ahead and found there are several more, so I stopped here 🙂 )

```suggestion
import org.zalando.logbook.json.internal.JsonMediaType;
```

logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
```diff
@@ -176,8 +178,11 @@ public HttpHeaders getHeaders() {
 
     @Override
     public Charset getCharset() {
-        return Optional.ofNullable(getCharacterEncoding())
-                .map(Charset::forName)
+        String contentTypeHeader = getHeaders().getFirst(CONTENT_TYPE_HEADER);
```
msdousti, line 181: I'm unsure whether this should be `getFirst` or `getLast`. I tested sending multiple content-type headers to a server using `curl`, and it seemed that the server used the last one.

logbook-api/src/main/java/org/zalando/logbook/internal/JsonMediaType.java
kasmarian: The problem here is that I actually moved the file to the api module so that I can use it in `org.zalando.logbook.ContentType` class. That's one of the changes I'm concerned about, but as we use default methods in the api interfaces, it seems to be a necessary evil. If you have good ideas how to avoid this without major breaking changes, please share.

logbook-api/src/test/java/org/zalando/logbook/ContentTypeTest.java
kasmarian: That's a very good catch! Indeed, there was a bug in recognizing Content-Type headers and recognizing JSON Content-Type headers. But according to the RFC 9110 they should be case-insensitive. I added more test cases and fixed it in https://github.com/zalando/logbook/pull/1521/commits/0d7cda1162867131d40003f952b9b5bd796cbef3

logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
kasmarian: Good question. I couldn't find any recommendations in this regard in the RFC, I think multiple header values overall should not be expected for Content-Type header. Here I followed the same approach we had in other classes (everywhere `getFirst()` is used). Bur if you find any leads about recommendations here, please share. I was mostly going through https://www.rfc-editor.org/rfc/rfc9110.html#fields

msdousti: I also couldn't find a reference on this. Yesterday, I read that only comma-separated headers (arrays for that matter) can be repeated multiple times. To test this non-standard behavior, I created a simple Python application (Flask) that echoed the request headers in response body, and sent the following curl command:

```bash
curl -v http://127.0.0.1:8080/headers -H 'content-type: application/json' -H 'content-type: text'
```

It returned the **last** `content-type`. Today, I tested it with the following Spring Boot Web controller:

```java
@RestController
@RequestMapping("/headers")
public class HeadersController {

    @GetMapping
    public String getRequestHeaders(@RequestHeader Map<String, String> headers) {
        StringBuilder sb = new StringBuilder();
        headers.forEach((key, value) -> sb.append(key).append(": ").append(value).append("\n"));
        return sb.toString();
    }
}
```

It returned the **first**  `content-type`. We can conclude that there is no standard behavior, and therefore the code is OK 👌 .

logbook-api/src/test/java/org/zalando/logbook/ContentTypeTest.java
msdousti: Thanks!
One more suggestion: Could something like this simplify the nested-if logic?

```java
public static String getSubstringBeforeSemicolon(String s) {
    final int index = s.indexOf(';');  // not sure if this should be: s.indexOf(';', 12)
    if (index != -1) return s.substring(0, index);
    return s;
}
```

logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
kasmarian: Thank you for verifying! I'll resolve the thread then.

msdousti:
👍

kasmarian:
:+1:

Issue #1216, opened 2021-11-18
Logbook only logs once with Netty module


jhoweaa:
I'm trying to use Logbook with Micronaut to configure my log messages. I'm using the logbook-netty module. I've noticed that I only get logging on the first request made to my server.

## Description
I have a simple Micronaut (3.1.4) web server and I want to filter log messages. I've configured logging per instructions found in the Micronaut documentation. When testing my application, I've noticed that I only get logbook messages the first time a request is made to the server. More specifically, I only get logbook messages the first time a request is handled by a particular thread in my server. Once a request is handled by thread 1, any other request handled by thread 1 will no longer emit any logbook messages.

## Expected Behavior

When I make a request to my server, I expect every request that isn't specifically filtered to appear in the logs

## Actual Behavior

What I'm observing is that I only get log messages the first time a particular thread handles a request.

## Possible Fix

I think the source of the bug is with the way Sequence is implemented. When a LogbookServerHandler is
created, a Sequence object is created with room for two tasks. The tasks are used to display the
request and response. The code in Sequence uses a variable called 'next' which controls whether a
task is executed.

```java
    synchronized void set(final int index, final Runnable task) {
        tasks.set(index, task);

        if (index == next) {
            runEagerly();
        }
    }
```
If a call is made to set with an index value ~= next, the task will not run. The 'runEagerly' method in
Sequence looks like this:

```java
    private void runEagerly() {
        final int end = tasks.size();

        for (@Nullable final Runnable task : tasks.subList(next, end)) {
            if (task == null) {
                return;
            }

            task.run();
            tasks.set(next, null);
            next++;
        }
    }
```

When a request and response are processed, both tasks will be processed. After the response 
task is completed, the value of 'next' will be 2. The value will remain at 2. When additional 
requests are made on the same thread, the value of 'next' is always 2 so the 'runEagerly' method 
is never called again.

I don't necessarily understand all the nuances involved, but I think the following change in
the 'runEagerly' method would solve the problem:

```java
    private void runEagerly() {
        final int end = tasks.size();

        for (@Nullable final Runnable task : tasks.subList(next, end)) {
            if (task == null) {
                return;
            }

            task.run();
            tasks.set(next, null);
            next++;
        }

        // Reset next to 0, we've exhausted the tasks and this instance will live on in the thread
        // that it was created in, so we need to reset next back to 0 if we want to see any more output

        next = 0;
    }
```

If my assumption is correct, the code will never normally exit the for loop without the value 
[...]

PR #1538, opened 2023-06-06
fix netty sequence ability to log subsequent reads and writes

arowe-adaptive:
## Description

This fixes the bug with the Netty module where it doesn't log subsequent requests.  The issue is that the Sequence class that manages sequential processing of reads, then writes has a bug.  The method runEagerly() keeps track of a next pointer that tells it which task to run.  The problem, also described in the linked issue, is the next pointer is never reset back to the first position in the task array.  Applying the suggested changes have fixed the issue.  I tested this in a Micronaut app and it is now logging all requests from the server and client reliably.  I also wrote a new unit tests showing the Sequence can be reused to process new tasks/requests/responses.  Also I wrote the unit test first to ensure it fails without the change.

## Motivation and Context
Fixes [1216](https://github.com/zalando/logbook/issues/1216)

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ x] I have added tests to cover my changes.

### DIFF
--- a/logbook-netty/src/main/java/org/zalando/logbook/netty/Sequence.java
+++ b/logbook-netty/src/main/java/org/zalando/logbook/netty/Sequence.java
@@ -35,6 +35,11 @@ final class Sequence {
             tasks.set(next, null);
             next++;
         }
+
+        // Reset next to 0, we've exhausted the tasks and this instance will live on in the thread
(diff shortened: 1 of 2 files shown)


whiskeysierra:
👍

PR #1576, opened 2023-07-11
LogbookFilter flush buffer will make tomcat add Transfer-Encoding: chunked header

dartilla:
LogbookFilter flush buffer will make tomcat add Transfer-Encoding: chunked header

## Description
HttpServletResponseWrapper.flushBuffer() changes output header in Spring-Boot:  (v2.5.4) with Tomcat
It adds Transfer-Encoding: chunked, and not allow Tomcat to insert Content-Length=0 with empty body responses.

## Motivation and Context
As I see there is no need in executing HttpServletResponseWrapper.flushBuffer() in LogbookFilter (actually any HttpFilter)
Logbook logging works without it.

## Types of changes

### DIFF
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LogbookFilter.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LogbookFilter.java
@@ -85,7 +85,7 @@ public final class LogbookFilter implements HttpFilter {
     private void write(RemoteRequest request, LocalResponse response, ResponseWritingStage writing) throws IOException {
         final AtomicBoolean attribute = (AtomicBoolean) request.getAttribute(responseWritingStageSynchronizationName);
         if (!attribute.getAndSet(true)) {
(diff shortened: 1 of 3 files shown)


whiskeysierra:
This would reintroduce a regression fixed by https://github.com/zalando/logbook/pull/617

whiskeysierra:
See also https://github.com/zalando/logbook/issues/603#issuecomment-540724918

dartilla:
I suspected that there are cases when this flush is mandatory.
But there are cases where it is not.
Maybe it should be property to flush or not to flush for those who struggle from tomcat?

kasmarian:
@dartilla would you volunteer to add such a property? 

It could probably sit under `logbook.write...`, e.g. `logbook.write.response.flush-before-write`, or could even make sense to use `servlet` in the name, as it's specific for servlets. E.g. `logbook.servlet.flush-response-before-write`. With default = true.

PR #1578, opened 2023-07-13
Use getCharacterEncoding when no Content-Type header is set

kasmarian:
## Description
Use `getCharacterEncoding()` if `Conent-Type` header is missing in the response, before falling back to ` ISO-8859-1` when determining charset.

## Motivation and Context

Changes in https://github.com/zalando/logbook/commit/ca3e0327972bfaa518c238a3a808ff2b4c9797ba started using Content-Type header of response object to determine the Charset. This was done as most server implementations would fallback to ` ISO-8859-1` encoding if the encoding wasn't explicitly set as part of Conent-Type header or in other ways. This cause problems with applications that work with JSON responses, as according to RFC 8259 JSON uses UTF-8 for encoding messages. 

[It was reported](https://github.com/zalando/logbook/issues/870#issuecomment-1625338454) that Thymeleaf projects are affected by the change as at the time of the evaluation, Content-Type header is not yet set to the response object, and Logbook falls back to ISO-8859-1. At the same time, `getCharacterEncoding()` does contain the correct encoding value for the response object. 

## Types of changes

## Checklist:

### DIFF
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LocalResponse.java
@@ -183,13 +183,16 @@ final class LocalResponse extends HttpServletResponseWrapper implements HttpResp
         return Optional
                 .ofNullable(contentTypeHeader)
                 .map(ContentType::parseCharset)
-                /*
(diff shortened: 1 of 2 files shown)


msdousti:
👍

kasmarian:
:+1:

Issue #1587, opened 2023-07-21
Webflux: No logs when request is filtered


MatthiasDrews:
Webflux: No logs when request is filtered out by a WebFilter

## Description
In a spring cloud gateway application it is possible (and maybe common) to have a set of global WebFilters which are applied to all requests. It turns out, that having a WebFilter which prevents further request processing (e.g. by returning a 401), prevents logging of the whole request/response by Logbook.

## Expected Behavior
All requests/responses should be logged, even if further request processing is cancelled by a WebFilter.

## Actual Behavior
Currently there are no logs for affected requests/responses.

## Steps to Reproduce
Add a filter to your application, e.g.:
```
@Bean
public WebFilter securityFilter() {
    return ((exchange, chain) -> {
        if (exchange.getRequest().getPath().value().equals("/secured")) {
            exchange.getResponse().setStatusCode(HttpStatus.UNAUTHORIZED);
            return exchange.getResponse().setComplete();
        } else {
            return chain.filter(exchange);
        }
    });
}
```

## Your Environment

* SpringBoot 3.1.2
* Logbook 3.2.0

kasmarian:
The fix will be part of 3.3.0

PR #1588, opened 2023-07-21
log request/response for rejected requests

MatthiasDrews:
## Description
Fix for #1587  

## Types of changes

## Checklist:

### DIFF
--- a/logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/BufferingServerHttpRequest.java
+++ b/logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/BufferingServerHttpRequest.java
@@ -30,16 +30,9 @@ final class BufferingServerHttpRequest extends ServerHttpRequestDecorator {
         super(delegate);
         this.serverRequest = serverRequest;
         this.writeHook = writeHook;
-    }
-
-    @Override
-    public HttpMethod getMethod() {
-        HttpMethod method = super.getMethod();
-        // the only way so far to check if request has body or not
-        if (METHODS_WITHOUT_BODY.contains(method)) {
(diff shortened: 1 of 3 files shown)


logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/BufferingServerHttpRequest.java
```diff
@@ -30,16 +30,9 @@ final class BufferingServerHttpRequest extends ServerHttpRequestDecorator {
...
         this.serverRequest = serverRequest;
         this.writeHook = writeHook;
-    }
-
-    @Override
-    public HttpMethod getMethod() {
-        HttpMethod method = super.getMethod();
-        // the only way so far to check if request has body or not
-        if (METHODS_WITHOUT_BODY.contains(method)) {
+        if (METHODS_WITHOUT_BODY.contains(super.getMethod()) || !serverRequest.shouldBuffer()) {
```
kasmarian, line 33: Interesting, so now writeHook will be executed immediately, probably it's fine in case of requests...

Why's `!serverRequest.shouldBuffer()`required?  Removing it doesn't fail any tests.

MatthiasDrews: The idea is to log requests as early as possible. Depending on the application, request processing might take some time. Without the `!serverRequest.shouldBuffer()`, POST requests in a `without-body` strategy are logged when `getBody()` is invoked, and this can be delayed

kasmarian:
:+1:

logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/BufferingServerHttpRequest.java
kasmarian: Understood, thank you. I merged the PR

PR #1589, opened 2023-07-22
Extract subject claim from JWT token

msdousti:
Extract subject claim from JWT token.
[This repository](https://github.com/msdousti/LogbookJwtTester) can be used to show how it works.

## Description

Original spec, as per #381:
image

Important changes made:
- [x] Added a new class, `HttpAttributes`  to keep interesting attributes as key-value pairs
- [x] Added this class to `HttpRequest` with default empty. Implemented it for ` CachingHttpRequest`, as it is the important class used in `DefaultLogbook` right before filtering the request.
- [x] Implemented `NoOpRequestAttributesExtractor`  and `JwtClaimExtractor`. The NoOp extractor is the default as per Willi's [secure by default idea](https://github.com/zalando/logbook/issues/381#issuecomment-561156707). The `JwtClaimExtractor` implements the actual logic in the above spec.
- [x] Added the extraction logic to the `Strategy` class. This way, it can be mixed with existing strategies. Implemented `ForwardingStrategy` to facilitate this behavior.

Remaining:
- [x] Check the idea with the team
- [ ] Should we move `JwtClaimExtractor` to another module, say `logbook-json`, so that `logbook-api` is kept free of `ObjectMapper`?
- [x] Add `HttpAttributes`  to `HttpResponse`, and adapt (or create) `Strategy`, ` CachingHttpResponse` and `DefaultLogbook`.
- [x] Logging the attributes is currently added to `StructuredHttpLogFormatter` (and therefore its children like `JsonHttpLogFormatter`). Support the attributes in other formatters.
- [x] Add settings to Spring Boot module.
- [ ] Adapt documentation.

## Motivation and Context
Addresses #381.

## Types of changes

## Checklist:


logbook-api/src/main/java/org/zalando/logbook/StructuredHttpLogFormatter.java
```diff
@@ -65,6 +65,9 @@ default Map<String, Object> prepare(final Precorrelation precorrelation, final H
         content.put("scheme", request.getScheme());
         content.put("port", preparePort(request));
 
+        if (!request.getAttributes().isEmpty())
```
kasmarian, line 68: suggestion: should the attributes also be added to the `DefaultHttpLogFormatter` ?

logbook-api/src/main/java/org/zalando/logbook/ForwardingStrategy.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.apiguardian.api.API;
+import org.zalando.logbook.attributes.RequestAttributesExtractor;
+
+import java.io.IOException;
+
+import static org.apiguardian.api.API.Status.STABLE;
+
+@API(status = STABLE)
+public interface ForwardingStrategy extends Strategy {
```
kasmarian, line 11: What's the idea for this strategy? Is it related to the JWT?

.mvn/extensions.xml
```diff
@@ -0,0 +1,10 @@
+<extensions xmlns="http://maven.apache.org/EXTENSIONS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+            xsi:schemaLocation="http://maven.apache.org/EXTENSIONS/1.0.0 http://maven.apache.org/xsd/core-extensions-1.0.0.xsd">
+
+    <extension>
+        <groupId>org.apache.maven.extensions</groupId>
+        <artifactId>maven-build-cache-extension</artifactId>
```
kasmarian, line 6: suggestion: put this in a separate PR, so that the change is recorded in the changelog

logbook-api/src/main/java/org/zalando/logbook/attributes/RequestAttributesExtractor.java
```diff
@@ -0,0 +1,17 @@
...
+import org.apiguardian.api.API;
+import org.zalando.logbook.HttpRequest;
+
+import javax.annotation.Nonnull;
+
+import static org.apiguardian.api.API.Status.STABLE;
+
+@API(status = STABLE)
+@FunctionalInterface
+public interface RequestAttributesExtractor {
```
kasmarian, line 12: Are you planning to add this concept to the `logbook-spring-boot-autoconfigure` module? So that extractors can be registered as beans and configured via properties, e.g. claim names could be controlled via properties.

logbook-api/src/main/java/org/zalando/logbook/Strategy.java
```diff
@@ -119,4 +121,8 @@ default void write(final Correlation correlation, final HttpRequest request, fin
         sink.write(correlation, request, response);
     }
 
+    default RequestAttributesExtractor getRequestAttributesExtractor() {
```
kasmarian, line 124: Why attributes extractors should be coupled with strategies and not be independent concept? If I want to use `StatusAtLeastStrategy` I should be able to use JWT attributes at the same time. 

Suggestion: add AttributesExtractor as a separate property in the `LogbookCreator`.

logbook-api/src/main/java/org/zalando/logbook/attributes/RequestAttributesExtractor.java
kasmarian, line 12: This change deserves a README update. As a customization feature.

msdousti:
@kasmarian 

Thanks for the insightful comments - I'll apply them soon.

As a high-level note: There are several remaining points that I documented in the PR:
image

logbook-api/src/main/java/org/zalando/logbook/ForwardingStrategy.java
msdousti: I'll use it later on when I implement it in Spring Boot module.
Also, our clients can use like this:

https://github.com/msdousti/LogbookJwtTester/blob/45d2b385fb937e44f13ed660fcc8648fd2faa48b/src/main/java/org/example/Main.java#L66-L77

logbook-api/src/main/java/org/zalando/logbook/attributes/RequestAttributesExtractor.java
msdousti: Definitely, it's on my ToDo list for this PR.

msdousti: Definitely, it's on my ToDo list for this PR.

logbook-api/src/main/java/org/zalando/logbook/Strategy.java
msdousti: They are not coupled; please see [this example](https://github.com/msdousti/LogbookJwtTester/blob/45d2b385fb937e44f13ed660fcc8648fd2faa48b/src/main/java/org/example/Main.java#L66-L77).

It is also a possibility to add it to `LogbookCreator` and I considered it; but having a single Strategy made more sense to me.

Of course I'm open for changing it - Let's discuss tomorrow.

whiskeysierra: > They are not coupled in the sense that you can easily mix

It may be easy (actually debatable, but let's say it is), but I wouldn't call it simple in the sense of Rich Hickey's [Simple made Easy](https://www.youtube.com/watch?v=LKtk3HCgTa8) talk.

> but having a single Strategy made more sense to me.

I'm with @kasmarian on this. `Strategy` and `AttributesExtractor` are are two unrelated concepts, they shouldn't depend or coupled together like this.

logbook-core/src/main/java/org/zalando/logbook/core/DefaultLogbook.java
```diff
@@ -44,7 +45,8 @@ public RequestWritingStage process(final HttpRequest originalRequest, final Stra
             final Precorrelation precorrelation = newPrecorrelation(originalRequest);
             final HttpRequest processedRequest = strategy.process(originalRequest);
 
-            final HttpRequest request = new CachingHttpRequest(processedRequest);
+            final HttpAttributes httpAttributes = strategy.getRequestAttributesExtractor().extractOrEmpty(processedRequest);
```
whiskeysierra, line 48: This is the only place that actually needs the `orEmpty` semantic. I would move it here.

logbook-api/src/main/java/org/zalando/logbook/attributes/HttpAttributes.java
```diff
@@ -0,0 +1,77 @@
...
+    public int hashCode() {
+        return map.hashCode();
+    }
+
+    @Override
+    public String toString() {
+        return map.toString();
+    }
+
+    public HttpAttributes fluentPut(String key, String value) {
```
whiskeysierra, line 47: Are those used anywhere? I'd ditch them in favor of making this class immutable.

logbook-api/src/main/java/org/zalando/logbook/attributes/RequestAttributesExtractor.java
```diff
@@ -0,0 +1,24 @@
...
+
+import org.apiguardian.api.API;
+import org.zalando.logbook.HttpRequest;
+
+import javax.annotation.Nonnull;
+
+import static org.apiguardian.api.API.Status.STABLE;
+
+@API(status = STABLE)
+public abstract class RequestAttributesExtractor {
```
whiskeysierra, line 11: This should be an interface with just the `extract` method. The `extractOrEmpty` is only needed internally and shouldn't be exposed (or imposed) on extractor implementations.

logbook-api/src/main/java/org/zalando/logbook/Strategy.java
msdousti: Hi Willi,

Thanks a lot for the review! It's really encouraging, and I'll apply them as soon as possible 👍 
I talked with Karen yesterday about how to proceed with the rest of this PR, and to add the HttpAttributes to the response as well. We have several choices:

- Have separate strategy interfaces for extracting them from request and from response, or a single interface having two methods.
- The method to extract HttpAttributes from the response can only take response as parameter, or both request and response.
- We can allow an array/list of such strategies to be registered during the creation of Logbook, or only a single strategy that combines them all.

Suggestions are very welcome!

whiskeysierra: > * Have separate strategy interfaces for extracting them from request and from response, or a single interface having two methods.

I'd go with a single interface and two methods both with a default implementation that returns an empty list of attributes maybe?

> * The method to extract HttpAttributes from the response can only take response as parameter, or both request and response.

I'd add both, giving implementations more information, e.g. only extract response attributes from those responses where the request matches a certain criterion.

> * We can allow an array/list of such strategies to be registered during the creation of Logbook, or only a single strategy that combines them all.

I'd go with a `CompositeAttributesExtractor` implementation that holds a list of extractors and is dealing with the loop internally. It should be an implementation detail though and during Logbook creation the user should be able to just call the corresponding builder method multiple times. I think we do the same approach with filters.

msdousti: Thanks a lot Willi, I'll follow your excellent suggestions. 👏

logbook-api/src/main/java/org/zalando/logbook/ForwardingStrategy.java
msdousti: Removed `ForwardingStrategy`.

logbook-api/src/main/java/org/zalando/logbook/Strategy.java
msdousti: Applied all suggestions.

logbook-api/src/main/java/org/zalando/logbook/attributes/HttpAttributes.java
msdousti: Removed all mutating methods; the class is now Immutable.

logbook-api/src/main/java/org/zalando/logbook/attributes/RequestAttributesExtractor.java
msdousti: Made it an interface and removed the extra method.

logbook-core/src/main/java/org/zalando/logbook/core/DefaultLogbook.java
msdousti: Moved it here as per suggestion.

msdousti:
@kasmarian @whiskeysierra 
Thank you again for the great review. I applied all your suggestions.
A remaining task is to add documentation, but I wanted to first make sure the code is finalized, and then add documentation to avoid rework.

msdousti:
Suggestions from JF:
- Simplify the logging of exceptions if possible
- Make APIs experimental
- Allow for a map with Object values

logbook-api/src/main/java/org/zalando/logbook/attributes/CompositeAttributeExtractor.java
```diff
@@ -0,0 +1,61 @@
...
+import org.zalando.logbook.HttpResponse;
+
+import javax.annotation.Nonnull;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@AllArgsConstructor
+@EqualsAndHashCode
+public final class CompositeAttributeExtractor implements AttributeExtractor {
```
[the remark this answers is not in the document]
msdousti: Yes, totally forgot that 👍 
Fixed in 9413622

logbook-api/src/main/java/org/zalando/logbook/attributes/JwtBaseExtractor.java
```diff
@@ -0,0 +1,85 @@
...
+                    getLogMarker(),
+                    "Encountered error while extracting attributes: `{}`",
+                    (Optional.ofNullable(exception.getCause()).orElse(exception)).getMessage()
+            );
+    }
+
+    @Nonnull
+    protected abstract Marker getLogMarker();
+
+    protected Map<?,?> extractClaims(final HttpRequest request) throws JsonProcessingException {
```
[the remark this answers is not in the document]
msdousti: Good idea. Fixed in 43236f5.

logbook-api/src/main/java/org/zalando/logbook/attributes/AttributeExtractor.java
```diff
@@ -0,0 +1,26 @@
...
+import org.zalando.logbook.HttpResponse;
+
+import javax.annotation.Nonnull;
+
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+
+@API(status = EXPERIMENTAL)
+public interface AttributeExtractor {
+
+    @SuppressWarnings("RedundantThrows")
```
kasmarian, line 14: Sorry, one more question: do we really want to have a checked exception in the interface? I understand that JWT related extractors need to deal with checked exceptions, but incorporating Exception in an interface signature seems to be unnecessary. Otherwise, most Logbook interfaces would have it in their method signatures.

msdousti: Your comments are really useful and always welcome, please don't apologize 😊
I'll fix it ASAP.

msdousti: Done: 3ebef9f.

kasmarian:
:+1:

msdousti:
@kasmarian Thanks for the approval.
I'll wait for your PRs (#1592, #1606) to be merged. They will result in some conflicts with this one, which I'll handle + update documentation.
Then we'll merge this one.

kasmarian:
@msdousti please don't wait for 1606, I'll continue the work on it after 2 weeks only. 1592 is merged!

msdousti:
@kasmarian Thanks for the heads up. Then I'll contine with this.

msdousti:
@kasmarian It's now ready for final review 🙏

rishiraj88:
A lot to learn out of this request. Thanks to all involved.

logbook-api/src/test/java/org/zalando/logbook/attributes/HttpAttributesTest.java
```diff
@@ -0,0 +1,75 @@
...
+        assertThat(attributes.toString()).isEqualTo("{}");
+
+        //noinspection DataFlowIssue
+        assertThatThrownBy(() -> attributes.put("key", "val"))
+                .isInstanceOf(UnsupportedOperationException.class);
+    }
+
+    @Test
+    void singletonHttpAttributesShouldBeImmutable() {
+        final Map<String, Object> map1Clone = new HashMap<>(mapWithOneKey);
```
whiskeysierra, line 44: Why do you make clones everywhere?

msdousti: Just wanted to make sure that the equality is based on the `map.equals(o)` and not object reference comparison `this == o`:

https://github.com/zalando/logbook/blob/b9a093dac25f6ddde152c00d018d99cb33218f4d/logbook-api/src/main/java/org/zalando/logbook/attributes/HttpAttributes.java#L30-L35

README.md
```diff
@@ -204,6 +204,61 @@ Logbook comes with some built-in strategies:
...
+### Attribute Extractor
+Starting with version 3.4.0, Logbook is equipped with a feature called *Attribute Extractor*. Attributes are basically a
+list of key/value pairs that can be extracted from request and/or response, and logged with them. The idea was sprouted
+from [issue 381](https://github.com/zalando/logbook/issues/381), where a feature was requested to extract the subject
+claim from JWT tokens in the authorization header.
+
+The `AttributeExtractor` interface has two `extract` methods: One that can extract attributes from the request only, and
+one that has both request and response at its avail. The both return an instance of the `HttpAttributes` class, which is
+basically a fancy `Map<String, Object>`. Notice that since the map values are of type `Object`, they should have a 
+proper `toString()` method in order for them to appear in the logs in a meaningful way.
```
whiskeysierra, line 216: If we rely on `toString()` anyway, why don't we change the signature to `Map<String, String>` then to be very precise in what we support?

logbook-api/src/main/java/org/zalando/logbook/StructuredHttpLogFormatter.java
```diff
@@ -91,14 +94,17 @@ default Map<String, Object> prepare(final Correlation correlation, final HttpRes
         content.put("protocol", response.getProtocolVersion());
         content.put("status", response.getStatus());
 
+        if (!response.getAttributes().isEmpty())
+            content.put("attributes", response.getAttributes());
```
whiskeysierra, line 98:
```suggestion
        if (!response.getAttributes().isEmpty()) {
            content.put("attributes", response.getAttributes());
        }
```

logbook-api/src/test/java/org/zalando/logbook/attributes/HttpAttributesTest.java
whiskeysierra: I've used https://jqno.nl/equalsverifier/ before to check the full equals/hashCode contract. Maybe it's applicable here as well.

README.md
msdousti: Good point.
Would it make sense to rely on `toString()` for some log formatters (e.g., `DefaultHttpLogFormatter`), but some others like `JsonHttpLogFormatter` can another mechanism (say ObjectMapper) to convert objects to string representation?

This gives the implementers of AttributeExtractor the flexibility of not having to deal with Strings if they know that the log formatter they are going to use takes care of it.

logbook-core/src/main/java/org/zalando/logbook/core/DefaultLogbook.java
```diff
@@ -71,18 +80,46 @@ public ResponseWritingStage process(final HttpResponse originalResponse) throws
         }
     }
 
+    private HttpAttributes extractAttributesOrEmpty(final HttpRequest request) {
+        // CompositeAttributeExtractor handles exceptions internally
+        if (attributeExtractor instanceof CompositeAttributeExtractor)
+            return ((CompositeAttributeExtractor) attributeExtractor).extract(request);
```
whiskeysierra, line 86: That's a violation of the https://en.wikipedia.org/wiki/Liskov_substitution_principle. Why can't we just use the extractor as-is?

logbook-core/src/main/java/org/zalando/logbook/core/attributes/CompositeAttributeExtractor.java
```diff
@@ -0,0 +1,78 @@
...
+        return new HttpAttributes(map);
+    }
+
+    @Nonnull
+    private HttpAttributes safeRequestExtractor(final AttributeExtractor attributeExtractor,
+                                                final HttpRequest request) {
+        try {
+            return attributeExtractor.extract(request);
+        } catch (Exception e) {
+            log.trace(
```
whiskeysierra, line 54: Maybe instead of catching, logging and falling back to a default value here, what about collecting all exceptions in a single exception (maybe using suppressed expections?) and throw that one instead?

whiskeysierra: Do we actually want to continue in case an extractor fails?

whiskeysierra: Are we doing the same with a sink, formatter or writer right now?

logbook-core/src/main/java/org/zalando/logbook/core/attributes/JwtBaseExtractor.java
```diff
@@ -0,0 +1,71 @@
...
+ * Extracts a single claim from the JWT bearer token in the request Authorization header.
+ * By default, the subject claim "sub" is extracted, but you can pass an (ordered) list of <code>claimNames</code>
+ * to be scanned. The first claim in <code>claimNames</code> is then returned, or an empty attribute if no matching
+ * claim is found.
+ */
+@API(status = EXPERIMENTAL)
+@Slf4j
+@AllArgsConstructor
+@EqualsAndHashCode
+public abstract class JwtBaseExtractor implements AttributeExtractor {
```
whiskeysierra, line 34: Personally I'm not a fan of implementation inheritance and I pretty much avoid it 100% of the time.
You could solve the need to share the claims extraction here by simply having a dedicated `ClaimsExtractor` class that provides that functionality. Any JWT extractor could then just use that as an instance field (composition) instead of a shared base class (inheritance).

logbook-core/src/main/java/org/zalando/logbook/core/DefaultHttpLogFormatter.java
```diff
@@ -53,6 +53,13 @@ public String format(final Precorrelation precorrelation, final HttpRequest requ
         result.append('\n');
 
         writeHeaders(request.getHeaders(), result);
+
+        if (!request.getAttributes().isEmpty()) {
+            result.append("Attributes: ");
+            result.append(request.getAttributes());
```
whiskeysierra, line 59: This relies on both the `toString` of the underlying map as well as the `toString` of the values.
Why not e.g. add each attribute on a separate line, maybe prefixed by "Attribute ", so that it reads:

```
Attribute Subject: me
Attribute Source: over there
```

whiskeysierra: (Maybe you come up with another format that is easier on the eyes than `Map.toString`?

logbook-api/src/main/java/org/zalando/logbook/StructuredHttpLogFormatter.java
whiskeysierra, line 98: This actually introduces a behavior that differs from the README and isn't specified in any test.
The JSONHttpLogFormatter will try to serialize the attributes as their original types (because we pass them as Object.
So the `toString` wouldn't be used, as the README claims it.

This is an interesting aspect, because it might be nice to get fully typed JSON values for non-primitive attributes, but it's very subtle right now, because there is no documentation nor test as far as I can tell, that verifies this.

logbook-core/src/main/java/org/zalando/logbook/core/attributes/CompositeAttributeExtractor.java
msdousti: My line of thinking about how to handle extractor failure was as follows: Assume we have multiple extractors, one of them being a JWT subject extractor, and an HTTP request has a malformed bearer token. While the extractor cannot extract the JWT subject claim (and thus throws an exception), we may be interested in extracting other attributes via the other extractors.

I'll look at `CompositeSink` to check whether it follows a similar logic.

logbook-api/src/test/java/org/zalando/logbook/attributes/HttpAttributesTest.java
msdousti: Addressed in 22d7160. I still needed clones to assert immutability in some cases, but testing eaulas and hashCode is much better.

logbook-core/src/main/java/org/zalando/logbook/core/DefaultLogbook.java
msdousti: Fixed in 4da59d4.

logbook-core/src/main/java/org/zalando/logbook/core/attributes/JwtBaseExtractor.java
msdousti: Fixed in cd098b2.

logbook-api/src/main/java/org/zalando/logbook/StructuredHttpLogFormatter.java
msdousti: Fixed in ee6ef41.

logbook-core/src/main/java/org/zalando/logbook/core/attributes/CompositeAttributeExtractor.java
msdousti: > Do we actually want to continue in case an extractor fails?

Yes, maybe some extractors succeed. I'd like to make it as robust as possible, and allow Logbook to extract as much information as it can. This way, extractors may fail independently of each other, allowing other successful ones to continue.

> Maybe instead of catching, logging and falling back to a default value here, what about collecting all exceptions in a single exception (maybe using suppressed expections?) and throw that one instead?

Fixed in 6fbff7a, with a little modification to the proposal: Instead of catching and throwing, we catch them and log all in one statement. This allows the non-failing ones to extract.

> Are we doing the same with a sink, formatter or writer right now?

`CompositeSink`, for example, uses fauxpas's `ThrowingConsumer`:

https://github.com/zalando/logbook/blob/8ad19a3e067b1d265cbf219714e21850f415f697/logbook-core/src/main/java/org/zalando/logbook/core/CompositeSink.java#L39-L41

logbook-core/src/main/java/org/zalando/logbook/core/DefaultHttpLogFormatter.java
msdousti: Fixed in e8c4248.

logbook-api/src/main/java/org/zalando/logbook/StructuredHttpLogFormatter.java
msdousti: Fixed in 0725b94: It's now bothe documented and tested.

README.md
msdousti: Please also see 0725b94.

README.md
```diff
@@ -933,27 +991,28 @@ MyClient(RestTemplateBuilder builder, LogbookClientHttpRequestInterceptor interc
...
-| `logbook.filters.body.default-enabled`   | Enables/disables default body filters that are collected by java.util.ServiceLoader                                                                                                 | `true`            |
+The following tables show the available configuration (sorted alphabetically):
+
+| Configuration                            | Description                                                                                                                                                                                                  | Default            |
+|------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------|
+| `logbook.attribute-extractors`           | List of [AttributeExtractor](#attribute-extractor)s, including configurations such as `type` (currently `JwtFirstMatchingClaimExtractor` or `JwtAllMatchingClaimsExtractor`), `claim-names` and `claim-key`. | `[]`               |
```
kasmarian, line 998: Would it make sense to also add it to `Example configuration` below?

msdousti: Done in ad5b01e.

kasmarian:
:+1:

msdousti:
👍

Issue #1506, opened 2023-05-27
Springboot configuration file prompt: Cannot resolve configuration property 'logbook.include'


343585776:
## Description
After upgrading the package version 3.0.0, the Springboot configuration file prompts: Cannot resolve configuration property 'logbook.include'.

## Expected Behavior
Support prompts in yml

## Actual Behavior
<img width="571" alt="image" src="https://github.com/zalando/logbook/assets/15976826/02b5b49d-f61d-47c0-9440-624bbd700b7d">

kasmarian:
@343585776 care to create a pull request?

kasmarian:
This should be covered by https://github.com/zalando/logbook/pull/1592 and will be in the next release. Please re-open the issue if it's not fully addressed

Issue #1266, opened 2022-01-05
Add a property do define fields in body to be obfuscated


nhmarujo:
## Detailed Description
I would like to have a property on which we could define the fields to obfuscate on body  (something like `logbook.obfuscate.body-fields`), just like there are similar properties for headers or query parameters. This would be very useful to avoid having to define a `BodyFilter` bean. This way, simple use cases could be completely handled by configuration.

kasmarian:
A comment to shoo away the bot, as I think the idea is valid and should be straight-forward to add. For `JsonPathBodyFilters` only though.

kasmarian:
@nhmarujo this should be covered by https://github.com/zalando/logbook/pull/1592 and will be in the next release. Please re-open the issue if it's not fully addressed

nhmarujo:
Thank you so much @kasmarian . Will take a look when the new release is out.

rishiraj88:
thank you, @kasmarian

Issue #1202, opened 2021-10-21
"logbook.write.max-body-size" not working when creating own BodyFilter that includes defaultValue()


kosi2801:
## Description
With using Spring autoconfiguration when a custom BodyFilter is created and the BodyFilters.defaultValue() is used in it, the configuration property "logbook.write.max-body-size" is not working anymore. 
This is somewhat unexpected, as there is no indication that the BodyFilters.truncate() is *not* a default filter. 
By reading the documentation it would be assumed that the truncation is always possible (ie. part of the defaultValue() filters).

## Expected Behavior
When a custom `@Bean BodyFilter bodyFilter()` is created and the BodyFilters.defaultValue() is part of the filters, the truncation functionality is applied.

## Actual Behavior
No truncation of the body content configured by the "logbook.write.max-body-size" property is applied.

## Possible Fix
No exact fix, but as a workaround I had to use:
```
@Autowired
LogbookAutoConfiguration autoConfiguration;

@Bean
public BodyFilter bodyFilter() {
    return merge(
            autoConfiguration.bodyFilter(), 
            replaceJsonStringProperty(singleton("secret"), "XXX"));
}
```

## Steps to Reproduce

1. Create SpringBoot app, create dummy endpoint
2. Create custom @Bean BodyFilter bodyFilter like in the documentation example 
```
@Bean
public BodyFilter bodyFilter() {
    return merge(
            defaultValue(), 
            replaceJsonStringProperty(singleton("secret"), "XXX"));
}
```
3. configure "logbook.write.max-body-size = 10"
4. send request to dummy endpoint with larger body content

## Context
I was trying to add custom filtering to the existing filters, suddenly the logfiles were swamped by large body logs.

## Your Environment

* Version used: 2.14.0
* SpringBoot Version: 2.0.3.RELEASE
* Java Version: 1.8

kosi2801:
This issue is still present. If a better handling is not possible or feasible, maybe a mention in the documentation could indicate that this is an unexpected behavior? eg. at the documentation for "logbook.write.max-body-size" or the @Bean Bodyfilter.

kasmarian:
This should be covered by https://github.com/zalando/logbook/pull/1592 and will be in the next release. Please re-open the issue if it's not fully addressed

PR #1592, opened 2023-07-24
Define json body obfuscation fields via configuration

kasmarian:
## Description

Make JSON body fields obfuscation configuration in Spring Boot applications via properties.
`JacksonJsonFieldBodyFilter` is initialized when `logbook.obfuscate.json-body-fields` property is defined. 

This change set also includes:
- separate default body filter from truncating so that it's possible to use truncating body filter if they define custom body filters.
- add configuration descriptions to spring's configuration-metadata.json file to show hints in IDEs when configuring logbook
- Introduced a property to disable the default logbook filter via configuration
- Introduced a configuration property to define obfuscation replacement string

## Motivation and Context
Relates to: 
- #1266
- #1506
- #1202

## Types of changes

## Checklist:

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/core/TruncatingBodyFilter.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/core/TruncatingBodyFilter.java
@@ -1,0 +1,23 @@
+package org.zalando.logbook.core;
+
+import lombok.experimental.Delegate;
+import org.apiguardian.api.API;
+import org.zalando.logbook.BodyFilter;
+
+import static org.apiguardian.api.API.Status.INTERNAL;
+import static org.zalando.logbook.core.BodyFilters.truncate;
+
+@API(status = INTERNAL)
+public final class TruncatingBodyFilter implements BodyFilter {
+
+    @Delegate
+    private final BodyFilter delegate;
+
+    public TruncatingBodyFilter(final int maxBodySize) {
+        if (maxBodySize < 0) {
+            this.delegate = (contentType, body) -> body;
+        } else {
+            this.delegate = truncate(maxBodySize);
+        }
+    }
+}
--- a/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
+++ b/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
@@ -212,7 +212,6 @@ public class LogbookAutoConfiguration {
     @API(status = INTERNAL)
     @Bean
     @ConditionalOnMissingBean(JacksonJsonFieldBodyFilter.class)
-    @ConditionalOnProperty("logbook.obfuscate.json-body-fields")
     public BodyFilter jsonBodyFieldsFilter() {
         final LogbookProperties.Obfuscate obfuscate = properties.getObfuscate();
         final List<String> jsonBodyFields = obfuscate.getJsonBodyFields();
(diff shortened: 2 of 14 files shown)


logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookProperties.java
```diff
@@ -27,6 +27,8 @@ public static class Obfuscate {
         private final List<String> headers = new ArrayList<>();
         private final List<String> parameters = new ArrayList<>();
         private final List<String> paths = new ArrayList<>();
+        private final List<String> jsonBodyFields = new ArrayList<>();
+        private final String replacement = "XXX";
```
msdousti, line 31: I wrote a simple test for this, and it fails to run due to `No setter found for property: replacement`.

The test looks like this, but I'm sure you can improve it:

```java
@LogbookTest(properties = "logbook.obfuscate.replacement = ZZZ")
class ObfuscateReplacementTest {

    @Autowired
    private Logbook logbook;

    @Test
    void shouldInitializeLogbook() {
        assertThat(logbook).isNotNull();
    }

}
```

logbook-core/src/main/java/org/zalando/logbook/core/TruncatingBodyFilter.java
```diff
@@ -0,0 +1,23 @@
...
+
+@API(status = INTERNAL)
+public final class TruncatingBodyFilter implements BodyFilter {
+
+    @Delegate
+    private final BodyFilter delegate;
+
+    public TruncatingBodyFilter(final int maxBodySize) {
+        if (maxBodySize < 0) {
+            this.delegate = (contentType, body) -> body;
```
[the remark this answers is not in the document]
kasmarian: `-1` is the default value. That's when we don't want any truncation at all. Another way to do it would be to have this check in the bean definition itself. I can move it there.

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookProperties.java
kasmarian: Thank you for catching this! I'll add the test.

kasmarian: added in https://github.com/zalando/logbook/pull/1592/commits/bb40c742faaaa0ff2980d1fb07da391dcee85dc8

msdousti:
👍

kasmarian:
:+1:

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -161,7 +162,7 @@ public QueryFilter queryFilter() {
         final List<String> parameters = properties.getObfuscate().getParameters();
         return parameters.isEmpty() ?
                 QueryFilters.defaultValue() :
-                replaceQuery(new HashSet<>(parameters)::contains, "XXX");
+                replaceQuery(new HashSet<>(parameters)::contains, properties.getObfuscate().getReplacement());
```
ChristianLohmann, line 165: is it a good idea to make the replacement configurable? In theory we could replace it with everything then. Or at least we should test it with  any imaginable weird replacement. Any kind of whitespace, special characters, even code snippets maybe ;)

PR #1595, opened 2023-07-28
Update Release Script

lukasniemeier-zalando:
## Description

* Update CHANGELOG as part of the "release" commit of the version being released
* Publish only once at the end of the release flow
* Add CHANGELOG diff to Github Release

## Motivation and Context

Better release experience for publishers (less risk) and better release consumer experience

## Checklist:

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -9,12 +9,16 @@ jobs:
   release:
     runs-on: ubuntu-latest
     steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+    - name: Extract Release Notes
+      uses: ffurrer2/extract-release-notes@v1.16.0
+      id: notes
     - name: Create Release
-      uses: actions/create-release@v1
+      uses: softprops/action-gh-release@v1
       env:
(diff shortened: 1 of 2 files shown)


lukasniemeier-zalando:
:+1:

.github/workflows/release.yaml
```diff
@@ -9,12 +9,16 @@ jobs:
   release:
     runs-on: ubuntu-latest
     steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+    - name: Extract Release Notes
+      uses: ffurrer2/extract-release-notes@v1.16.0
```
kasmarian, line 15: No way! Are we going to have proper release notes now? :heart_eyes:

kasmarian:
:+1:

.github/workflows/release.yaml
lukasniemeier-zalando: Haha yes that's the idea :smile: We may have also used the "auto-generated" release notes of Github here, but I like to have them look exactly like they look in `CHANGELOG.md`

Issue #1597, opened 2023-07-31
Changelog Contains Issues Closed Due to Inactivity


PascalSchumacher:
https://github.com/zalando/logbook/blob/main/CHANGELOG.md#330-2023-07-25 contains issues which were not resolved but closed by a bot due to inactivity, e.g.: https://github.com/zalando/logbook/issues/1307

Imho it is confusing to have these unresolved enhancement requests/bug reports etc. as part of the changelog.

By the way: Thank you very much for providing logbook! 👍

PascalSchumacher:
@lukasniemeier-zalando Thank you very much!

PR #1600, opened 2023-08-01
exclude stale issues from changelog

lukasniemeier-zalando:
Fixes https://github.com/zalando/logbook/issues/1597

## Description

We exclude closed issues with the `stale` label from the changelog, as they are typically not _done_ but rather just closed for backlog maintenance reasons. We use the default "issue label exclude list" plus `stale`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,9 @@
 
 [Full Changelog](https://github.com/zalando/logbook/compare/3.3.0...HEAD)
 
-**Fixed bugs:**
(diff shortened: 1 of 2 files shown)


bocytko:
Is there a plan to include the change log in the Github Releases, just like we do for [riptide](https://github.com/zalando/riptide/releases)?

lukasniemeier-zalando:
Yes, we will start with the next release (see https://github.com/zalando/logbook/pull/1595)

kasmarian:
:+1:

Issue #1556, opened 2023-06-20
Can we have filter to exclude by Http method name


sadhir1502:
it would be great if we are able to exclude logging by http method name

## Detailed Description
For example lets say i want to exclude all the GET,DELETE calls from the logging we should have configurations as we have for excluding by urls, and excluding by contenttype.

whiskeysierra:
@sadhir1502 Care to create a pull request?

sadhir1502:
Sure i will raise a pull request once changes are done.

kasmarian:
@sadhir1502 this should be available starting from the next release. See more details in this PR - https://github.com/zalando/logbook/pull/1606 and [README](https://github.com/zalando/logbook#example-configuration)

PR #1606, opened 2023-08-07
Add Method predicates

kasmarian:
## Description
- deprecate `logbook.exclude` and `logbook.include` configs (but still support them)
- move `include` & `exclude` configs under `predicate` config
- add method predicates and combine them with url based predicates

## Motivation and Context
Inspired by  #1556. But also, as in my team we use HTTP method predicate as an override, I think it could be a good addition to the default configuration.

Depends on #1592 as it's branched off it. Should be merged after.

## Types of changes

## Checklist:


README.md
```diff
@@ -937,8 +937,10 @@ The following tables show the available configuration:
...
 |------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------|
-| `logbook.include`                        | Include only certain URLs (if defined)                                                                                                                                              | `[]`               |
-| `logbook.exclude`                        | Exclude certain URLs (overrides `logbook.include`)                                                                                                                                  | `[]`               |
+| `logbook.predicate.include.urls`         | Include only certain URLs (if defined)                                                                                                                                              | `[]`               |
```
whiskeysierra, line 940: Is my understanding correct, that URL- and method-matching can't be combined for a single match, i.e. it's not possible with this configuration to e.g. say `include POST requests to /api/v1`. Is that correct?

kasmarian: On the contrary, that's how they'll work. If someone configures both `url` and `method` in `logbook.predicate.include` than only the combination of the two will be logged. See examples in [IncludeTest](https://github.com/zalando/logbook/pull/1606/files#diff-6b8212a17c483142caa6609499188dc17270b6bbd4f76a6fd01e811630272c56R66-R69)

whiskeysierra:
Good, but that means I can configure something like: only log "POST to
/api/v1" and "GET /search", right?

README.md
kasmarian: Moving the discussion back here :smile: 

> Good, but that means I can configure something like: only log "POST to
/api/v1" and "GET /search", right?


You're right here, the predicates are not grouped by routes with this implementation. All `include` predicates will be merged together (method and urls ones separately first). So if someone configures
`logbook.predicate.include.urls=/api/v1,/api/v2` and `logbook.predicate.include.methods=GET,POST`
Then it'll allow all combinations of paths and methods. 

I do like the idea of grouping the predicates per route, where per one path, several methods can be added. Then it'll work as you mentioned.

whiskeysierra: I was thinking more along the lines of this:

```yaml
logbook.predicate.include:
  - path: /api/v1
    method: POST
  - path: /search
    method: GET
```

The url/path wouldn't be special here, in fact it could even be optional (same as method):

```yaml
logbook.predicate.include:
  - method: DELETE # all paths
  - path: /api/v1 # all methods
```

kasmarian: Agreed, this would be quite flexible. I'll add it this way :+1:  (after my vacation).

kasmarian: @whiskeysierra I've updated the PR to work as we discussed here

msdousti (commented):
I tried four cases by adjusting `application-exclude.yaml` file:

1. Exclude all using predicates:
```yaml
logbook:
  predicate:
    exclude:
      - path: /**
```

2. Include all using predicates:
```yaml
logbook:
  predicate:
    include:
      - path: /**
```

3. Exclude all using deprecated exclude:
```yaml
logbook:
    exclude:
      - /**
```

4. Include all using deprecated include:
```yaml
logbook:
    include:
      - /**
```

Then I executed [shouldExcludeExpectedRequests](https://github.com/zalando/logbook/blob/109564ad90d8f410c212805a1f4766e8d9343224/logbook-spring-boot-autoconfigure/src/test/java/org/zalando/logbook/autoconfigure/ExcludeTest.java#L52-L74), while setting `shouldLog` to true/false depending on the test case.

Strangely, only one of the four test cases passed.

To double check, I also created a Spring Boot application with test case 3, and observed that Logbook still logs all the requests.

kasmarian:
@msdousti that's a good catch! Using stream() API without double-checking was a mistake. I pushed a fix for it, please take a look.

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -149,18 +149,55 @@ private Collection<BodyFilter> mergeWithTruncation(List<BodyFilter> bodyFilters)
...
                 .map(Conditions::requestTo)
                 .map(Predicate::negate)
                 .reduce(predicate, Predicate::and);
+
+        return properties.getPredicate().getExclude()
+                .stream()
+                .map(this::convertToPredicate)
+                .map(Predicate::negate)
+                .reduce(predicate, Predicate::and)
+                .and(deprecatedPredicate);
```
msdousti, line 163: I suggest simplifying this function as follows:

```suggestion
        return Stream.concat(
                        properties.getExclude() // backwards compatibility for deprecated config
                                .stream()
                                .map(Conditions::requestTo),
                        properties.getPredicate().getExclude()
                                .stream()
                                .map(this::convertToPredicate)
                )
                .map(Predicate::negate)
                .reduce(predicate, Predicate::and);
```

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -149,18 +149,55 @@ private Collection<BodyFilter> mergeWithTruncation(List<BodyFilter> bodyFilters)
...
                 .reduce(Predicate::or)
                 .map(predicate::and)
                 .orElse(predicate);
+
+        return properties.getPredicate().getInclude()
+                .stream()
+                .map(this::convertToPredicate)
+                .reduce(Predicate::or)
+                .map(deprecatedPredicate::or)
+                .orElse(deprecatedPredicate);
```
msdousti, line 179: Similarly here:

```suggestion
        return Stream.concat(
                        properties.getInclude() // backwards compatibility for deprecated config
                                .stream()
                                .map(Conditions::requestTo),
                        properties.getPredicate().getInclude()
                                .stream()
                                .map(this::convertToPredicate)
                )
                .reduce(Predicate::or)
                .map(predicate::and)
                .orElse(predicate);
```

msdousti (commented):
Thanks a lot, it works like a charm.
I have made suggestions that simplify `mergeWithExcludes` & `mergeWithIncludes`; besides this small suggestion, all looks super.

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
kasmarian: This looks much better! Applied in https://github.com/zalando/logbook/pull/1606/commits/7f73fce9ab7031c67f676b2b7b809302d1859a35

msdousti:
👍

PR #1608, opened 2023-08-09
ensure truncation body filter is always last one

kasmarian:
## Description
Adding a truncation filter as the last one in the list of BodyFilters before creating Logbook.
Also remove a separate class for truncation as it's not useless.

## Motivation and Context
This PR is a follow-up from https://github.com/zalando/logbook/pull/1592 to ensure that truncation always happens at the end.
Kudos to @ado-dao for spotting.

## Types of changes

## Checklist:


ChristianLohmann:
lgtm

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -118,14 +119,25 @@ public Logbook logbook(
...
 
+    private Collection<BodyFilter> mergeWithTruncation(List<BodyFilter> bodyFilters) {
+        final LogbookProperties.Write write = properties.getWrite();
+        final int maxBodySize = write.getMaxBodySize();
+        if (maxBodySize > 0) {
+            BodyFilter filter = truncate(maxBodySize);
+            // To ensure that truncation will happen after all other body filters
+            bodyFilters.add(filter);
+        }
+        return bodyFilters;
```
ado-dao, line 138: 1. WDYT about using new List for case if Spring provides ImmutableList for bodyFilters
2. In the documentation maxBodySize == -1 means disabled, but 0 should truncate body to empty string?
```suggestion
        if (maxBodySize < 0) {
            return bodyFilters;
        }
        // To ensure that truncation will happen after all other body filters
        List<BodyFilter> filters = new ArrayList(bodyFilters);
        filters.add(truncate(maxBodySize));
        return filters;
```

kasmarian: Good suggestion! Even though tests would fail as soon as we update the version to the one with breaking change, no need to wait for it. Added in https://github.com/zalando/logbook/pull/1608/commits/bff2e1ea40919dccabca6e05ef38d618bb84e07d

msdousti:
👍

ado-dao:
👍

kasmarian:
:+1:

PR #1665, opened 2023-11-07
Call close() on closeable ClientHttpResponse within BufferingClientHttpResponseWrapper.java

bobross-pg:
Clear resources

## Description
On exception and close operation, only 1 of 2 close-able resources are closed.

## Motivation and Context
It looks like it could lead to memory leak.

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/logbook-spring/src/main/java/org/zalando/logbook/spring/BufferingClientHttpResponseWrapper.java
+++ b/logbook-spring/src/main/java/org/zalando/logbook/spring/BufferingClientHttpResponseWrapper.java
@@ -48,7 +48,9 @@ public class BufferingClientHttpResponseWrapper implements ClientHttpResponse {
         } catch (IOException e){
             throw new RuntimeException(e);
         }
-        delegate.close();
+        finally {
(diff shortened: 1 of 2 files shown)


logbook-spring/src/main/java/org/zalando/logbook/spring/BufferingClientHttpResponseWrapper.java
```diff
@@ -48,7 +48,9 @@ public void close() {
         } catch (IOException e){
             throw new RuntimeException(e);
         }
-        delegate.close();
+        finally {
+            delegate.close();
```
kasmarian, line 52: That's a really nice catch, @bobross-pg ! Would you also add a test for this scenario? So that we don't hit regression in future.

bobross-pg:
I will use a different github account, redo PR with with signed commits, how can I refuse/close this PR ?

bobross-pg:
Found the button to close :)

PR #1670, opened 2023-11-12
Spring framework 4 compatibility

luc-leclerc:
## Description
- Added a compatibility interceptor that should work with spring framework 4.X, 5.X and 6.X.
- I only altered the current classes to allow extending them enough to patch the status code, and discard the use of HttpStatusCode.

## Motivation and Context

- I have to maintain an old spring framework 4 project, which can't be migrated. I can't use latest logbook. due to spring HttpStatusCode class missing, Intellij doesn't like it and prevent proper loading of the project.

## Types of changes

## Checklist:

## Alternative solutions (ideas to review)
- Instead of this PR, I could remove spring dependency from logbook-parent, and add new sub-modules for each client version of spring (similar to httpclient/httpclient5). I can create extra sub-module for `logbook-spring4` and `logbook-spring5`, that way the tests would be on real framework without reflection shenanigans.
-- Optionally, for consistency, renaming logbook-spring to logbook-spring6?

### DIFF
--- a/logbook-spring/src/main/java/org/zalando/logbook/spring/LogbookClientHttpRequestInterceptor.java
+++ b/logbook-spring/src/main/java/org/zalando/logbook/spring/LogbookClientHttpRequestInterceptor.java
@@ -13,7 +13,7 @@ import java.io.IOException;
 
 @API(status = API.Status.EXPERIMENTAL)
 @AllArgsConstructor
-public final class LogbookClientHttpRequestInterceptor implements ClientHttpRequestInterceptor {
+public class LogbookClientHttpRequestInterceptor implements ClientHttpRequestInterceptor {
 
     private final Logbook logbook;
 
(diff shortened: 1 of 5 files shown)


luc-leclerc:
If we can decline this PR, I can look for the sub-modules approach, I opened this PR to start the conversation, to see if there was interest to support older spring 4 version. If this approach with reflection is acceptable, I got a few ideas improve this PR. 
Waiting for feedback.

msdousti:
@luc-leclerc Thanks a lot for opening this PR. I read through it and even wrote an application to check the interceptor, both for Spring 5 (which works with Logbook 3+) and 4 (which is supported by Logbook 2.x).

While appreciating the effort you did here, and after careful consideration, there are a few reasons why I'm leaned toward not going forward with the PR:

* Spring 4.x has reached its end of life long ago, and as such, its support was dropped in Logbook 3. We are planning to also drop support for Spring 5 sometime in the future (maybe 1 year, maybe several years - but the wish is there.) As the developers for a library, supporting many different versions is really hard: Every now and then, something breaks and it takes time and effort to fix it, rather than focusing on adding new features.
* The alternative exists: Use Logbook 2.16 and it works fine with Spring 4.x.
* Class inheritance (as opposed to interface inheritance) is discouraged, and to the best of my knowledge, the original developer of this library tried to avoid it. (See [his comment on one of my PRs](https://github.com/zalando/logbook/pull/1589#discussion_r1292243235)). I agree and like to honor his design.
* This PR does not add "Spring framework 4 compatibility" - rather, it only adds support for the Logbook filter for RestTemplate. I can envision that some other developer might be interested in full Spring framework 4 compatibility, and that's a pandora box I don't dare to open (👰❓🎁🔝🌋😈).

Issue #1693, opened 2023-11-29
UnsupportedOperationException in LogbookHttpRequestInterceptor with Spring Boot 3.2.0 and Apache Http Client 5


marcindabrowski:
## Description
After upgrade to Spring Boot 3.2.0 (Spring Framework 6.1.1) I'm getting this exception:

```java
Caused by: java.lang.UnsupportedOperationException: null
	at org.springframework.http.client.HttpComponentsClientHttpRequest$BodyEntity.getContent(HttpComponentsClientHttpRequest.java:146) ~[spring-web-6.1.1.jar:6.1.1]
	at org.apache.hc.client5.http.impl.classic.RequestEntityProxy.getContent(RequestEntityProxy.java:100) ~[httpclient5-5.2.1.jar:5.2.1]
	at org.apache.hc.core5.http.io.entity.EntityUtils.toByteArray(EntityUtils.java:135) ~[httpcore5-5.2.3.jar:5.2.3]
	at org.zalando.logbook.httpclient5.HttpEntities.copy(HttpEntities.java:23) ~[logbook-httpclient5-3.6.0.jar:na]
	at org.zalando.logbook.httpclient5.LocalRequest$Offering.buffer(LocalRequest.java:86) ~[logbook-httpclient5-3.6.0.jar:na]
	at org.zalando.logbook.httpclient5.LocalRequest.lambda$getBody$3(LocalRequest.java:251) ~[logbook-httpclient5-3.6.0.jar:na]
	at org.zalando.fauxpas.ThrowingFunction.apply(ThrowingFunction.java:19) ~[faux-pas-0.9.0.jar:na]
	at java.base/java.util.concurrent.atomic.AtomicReference.updateAndGet(AtomicReference.java:210) ~[na:na]
	at org.zalando.logbook.httpclient5.LocalRequest.getBody(LocalRequest.java:250) ~[logbook-httpclient5-3.6.0.jar:na]
	at org.zalando.logbook.HttpMessage.getBodyAsString(HttpMessage.java:44) ~[logbook-api-3.6.0.jar:na]
	at org.zalando.logbook.ForwardingHttpMessage.getBodyAsString(ForwardingHttpMessage.java:47) ~[logbook-api-3.6.0.jar:na]
	at org.zalando.logbook.core.FilteredHttpRequest.getBodyAsString(FilteredHttpRequest.java:92) ~[logbook-core-3.6.0.jar:na]
[...]

PR #1701, opened 2023-12-04
Use writeTo() to buffer HttpEntity in httpclient5

kasmarian:
## Description
Use `HttpEntity.writeTo()` instead of relying on `EntityUtils.toByteArray()` to copy the `OutputStream` of the HttpEntity.

## Motivation and Context

With a change in Spring Framework 6.1, HttpComponentsClientHttpRequest has an internal class [BodyEntity](https://github.com/spring-projects/spring-framework/blob/e870912fa2d4d05830ac476903eecaf6e907ca49/spring-web/src/main/java/org/springframework/http/client/HttpComponentsClientHttpRequest.java#L120), that is meant to not cache an `OutputStream`, but to provide a callback, that's used when the `OutputStream` is required. The [getContent()](https://github.com/spring-projects/spring-framework/blob/e870912fa2d4d05830ac476903eecaf6e907ca49/spring-web/src/main/java/org/springframework/http/client/HttpComponentsClientHttpRequest.java#L146) method throws `UnsupportedOperationException`.

With this change, I'm using the `writeTo()` method to copy the `OutputStream`, as for Logbook, buffering the stream is required.

Addresses https://github.com/zalando/logbook/issues/1693 

## Types of changes

## Checklist:

### DIFF
--- a/logbook-api/pom.xml
+++ b/logbook-api/pom.xml
@@ -20,7 +20,7 @@
         <dependency>
             <groupId>nl.jqno.equalsverifier</groupId>
             <artifactId>equalsverifier</artifactId>
-            <version>3.15.3</version>
+            <version>3.15.4</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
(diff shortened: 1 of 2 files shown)


msdousti:
👍

kasmarian:
:+1:

PR #1719, opened 2023-12-14
Add custom log level support in LogstashLogbackSink constructor

achhibi:
## Description
This pull request introduces the ability for users to customize the log level when creating a new instance of LogstashLogbackSink. The addition of the `logLevel` parameter in the constructor allows developers to specify their preferred logging level (info, debug, warn, error), providing greater flexibility in log management. The default remains set to trace for seamless integration.

## Motivation and Context
This change is motivated by the need to offer users more control over log levels in LogstashLogbackSink. By allowing them to define the log level during instantiation, developers can tailor logging behavior to their specific requirements.

## Types of changes

## Checklist:
-  [x] I have fixed the unit tests..


logbook-logstash/src/main/java/org/zalando/logbook/logstash/LogstashLogbackSink.java
```diff
@@ -27,19 +27,41 @@ public final class LogstashLogbackSink implements Sink {
 
     private final String baseField;
 
+    private final String level;
```
kasmarian, line 30: Why not use `org.slf4j.event.Level` instead? In that case, we'd be able to use [atLevel](https://www.slf4j.org/api/org/slf4j/Logger.html#atLevel-org.slf4j.event.Level) method and not use a custom enum

achhibi: @kasmarian Thank you so much for your valuable comments, I have considered incorporating `org.slf4j.event.Level` as you recommended, aiming to manage logging levels more effectively.

This customization doesn't alter the default behavior of LogstashLogbackSink but provides  more flexibility for users.

kasmarian: Thank you! Looks like the test coverage is lagging behind a bit, if you could fix it as well. Otherwise LGTM

achhibi: @kasmarian 

I have corrected and verified that the coverage is indeed 100%
Thank you !

kasmarian:
:+1:

kasmarian:
@achhibi thank you for both of your contributions. We'll include them in the next release

achhibi:
@kasmarian Thank you for your time and feedback.

Issue #1702, opened 2023-12-04
Make Logbook interceptors fault tolerant


marcindabrowski:
Since Logbook interceptors are not altering HTTP requests and responses they should do not throw Exceptions on they failure, because it could break HTTP connection.

## Detailed Description

[LogbookHttpRequestInterceptor.process](https://github.com/zalando/logbook/blob/main/logbook-httpclient5/src/main/java/org/zalando/logbook/httpclient5/LogbookHttpRequestInterceptor.java#L27) method according to Apache [HttpRequestInterceptor.process](
https://hc.apache.org/httpcomponents-core-5.1.x/current/httpcore5/apidocs/org/apache/hc/core5/http/HttpRequestInterceptor.html#process(org.apache.hc.core5.http.HttpRequest,%20org.apache.hc.core5.http.EntityDetails,%20org.apache.hc.core5.http.protocol.HttpContext)) is allowed to throw [HttpException](https://hc.apache.org/httpcomponents-core-5.1.x/current/httpcore5/apidocs/org/apache/hc/core5/http/HttpException.html) and [IOException](https://docs.oracle.com/javase/7/docs/api/java/io/IOException.html?is-external=true), but actually it could throw any `RuntimeExcpetion` that happens during  processing, ie.: https://github.com/zalando/logbook/issues/1693.

## Context

With this change we will be sure that even if there will be some issues, for example after some libraries version bump, it will not affect my HTTP communication. Logbook is just logging, so if it fails, it should not break my HTTP communication.

## Possible Implementation

This is example for [LogbookHttpRequestInterceptor.process](https://github.com/zalando/logbook/blob/main/logbook-httpclient5/src/main/java/org/zalando/logbook/httpclient5/LogbookHttpRequestInterceptor.java)
```java
try {
    LocalRequest request = new LocalRequest(httpRequest, entity);
    final ResponseProcessingStage stage = logbook.process(request).write();
    context.setAttribute(Attributes.STAGE, stage);
  } catch (HttpException | IOException exception) {
    throw exception;
  } catch (Exception ignored) {
  }
}
```

Of course it should be applied to all request and response interceptors.

## Your Environment

* Version used: 3.6.0

kasmarian:
[...]

PR #1739, opened 2024-01-21
Logbook fault tolerancy

kasmarian:
## Description

This change introduces 2 different fallback mechanisms (possible overlapping): 
- in httpclient(5), okhttp(2) and spring interceptors. Where in case of exceptions, logbook will fall back to noop stage and will not fail the overall request/response processing by an application
- in `DefaultLogbook`.  Where request stage preparing method (`RequestWritingStage process(final HttpRequest originalRequest, final Strategy strategy)`), as well as request writing and response write methods of the `DefaultRequestWritingStage` (new inner class) will be intercepting all `RuntimeExceptions` and will not fail the request processing. This approach may be sufficient to cover the cases of library dependent interceptors.

## Motivation and Context

As of now, when Logbook is causing RuntimeExceptions, it affects not only the ability to write the request/response logs, but prevents clients from operating normally. This should not be the case.

Related issues: https://github.com/zalando/logbook/issues/1693, https://github.com/zalando/logbook/issues/1702

## Types of changes

## Checklist:

### DIFF
--- a/logbook-httpclient/src/main/java/org/zalando/logbook/httpclient/LogbookHttpRequestInterceptor.java
+++ b/logbook-httpclient/src/main/java/org/zalando/logbook/httpclient/LogbookHttpRequestInterceptor.java
@@ -29,7 +29,7 @@ public final class LogbookHttpRequestInterceptor implements HttpRequestIntercept
             final ResponseProcessingStage stage = logbook.process(request).write();
             context.setAttribute(Attributes.STAGE, stage);
         } catch (Exception e) {
-            log.trace("Unable to log request: {}", e.getClass());
+            log.warn("Unable to log request. Will skip the request & response logging step.", e);
         }
     }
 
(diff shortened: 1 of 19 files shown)


logbook-httpclient/src/main/java/org/zalando/logbook/httpclient/LogbookHttpRequestInterceptor.java
```diff
@@ -22,9 +24,13 @@ public LogbookHttpRequestInterceptor(final Logbook logbook) {
...
     public void process(final HttpRequest httpRequest, final HttpContext context) throws IOException {
-        final LocalRequest request = new LocalRequest(httpRequest);
-        final ResponseProcessingStage stage = logbook.process(request).write();
-        context.setAttribute(Attributes.STAGE, stage);
+        try {
+            final LocalRequest request = new LocalRequest(httpRequest);
+            final ResponseProcessingStage stage = logbook.process(request).write();
+            context.setAttribute(Attributes.STAGE, stage);
+        } catch (Exception e) {
+            log.trace("Unable to log request: {}", e.getClass());
```
lukasniemeier-zalando, line 32: See my comment above on log level and wording. Also: why log the exception class explicitly (instead of passing the exception as the last argument)?

kasmarian: > why log the exception class explicitly (instead of passing the exception as the last argument)? 

Can't remember why I did this to be honest. Changed to passing the whole exception as in other places

lukasniemeier-zalando:
:+1:

PR #1742, opened 2024-01-22
Check for Vulnerabilities via GHE Action

lukasniemeier-zalando:
Check for Vulnerabilities via GHE Action

## Description

This pull request changes the way how we check for security advisories. We move to checking pull requests for introducing new vulnerabilities due to dependency changes via https://github.com/actions/dependency-review-action. Newly discovered vulnerabilities are continued to be reported via https://github.com/zalando/logbook/security/dependabot.

## Motivation and Context

The `org.owasp.dependency-check-maven` plugin isn't working reliably anymore and an upgrade to its new version seems not to be feasible (requires API key, users still report slowness), hence we are looking for more streamlined options to make use of Github capabilities directly.

We accept the fact that local builds are not scanned anymore, given that most developers were skipping it either way (due to the slowness).


.github/workflows/build.yaml
```diff
@@ -26,6 +26,11 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v3
+    - name: Dependency Review
+      uses: actions/dependency-review-action@v4
+      with:
+        vulnerability-check: true
+        license-check: false
```
kasmarian, line 33: How about `comment-summary-in-pr`? May save some time to open the pipeline and see what actually failed.

kasmarian:
:+1:

lukasniemeier-zalando:
:+1:

PR #1744, opened 2024-01-22
Check for Vulnerabilities via GHE Action

lukasniemeier-zalando:
Follow-up of https://github.com/zalando/logbook/pull/1742

See https://github.com/zalando/logbook/actions/runs/7613871782 for example.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -29,6 +29,8 @@ jobs:
     - name: Dependency Review
       uses: actions/dependency-review-action@v4
       with:
+        base-ref: ${{ github.event.pull_request.base.sha || '0.1.0' }}
+        head-ref: ${{ github.event.pull_request.head.sha || github.ref }}
         vulnerability-check: true
         license-check: false
         comment-summary-in-pr: on-failure


lukasniemeier-zalando:
:+1:

.github/workflows/build.yaml
```diff
@@ -29,6 +29,8 @@ jobs:
     - name: Dependency Review
       uses: actions/dependency-review-action@v4
       with:
+        base-ref: ${{ github.event.pull_request.base.sha || '0.1.0' }}
```
kasmarian, line 32: In https://github.com/actions/dependency-review-action it says that `base-ref/head-ref` is used for event types other than `pull_request` and `pull_request_target`. Why do we need the `github.event.pull_request.base.sha` and `github.event.pull_request.head.sha` then?

lukasniemeier-zalando: as far as I can see their documentation is imprecise, see https://github.com/actions/dependency-review-action/blob/935098a9508277d66b6c3adc56de47383d4c1e56/src/git-refs.ts#L10-L11

kasmarian: Interesting, so if we define `base-ref` and `head-ref` they'll override PRs events as well.

lukasniemeier-zalando:
The (bit ugly) trick of running non-pull-request builds by comparing the current state with the early tag seems to work - https://github.com/zalando/logbook/actions/runs/7623503585/job/20763667755 looked for vulnerabilities across all dependencies :heavy_check_mark:

PR #1753, opened 2024-01-28
Improve InputStream and OutputStream copying

achhibi:
## Description

This merge request introduces improvements to the stream copying functionality. The existing static method has been refactored to use a dynamic buffer size and take advantage of try-with-resources for automatic stream closure and avoid using of `while(true)`. These changes enhance the clarity and maintainability of the code.

## Motivation and Context

This merge request addresses the need to improve the efficiency and clarity of stream copying operations within the codebase.

## Types of changes

## Checklist:

### DIFF
--- a/logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
@@ -24,7 +24,7 @@ import java.util.Set;
 @Slf4j
 public class JacksonJsonFieldBodyFilter implements BodyFilter {
 
-    private final static StringReplaceJsonCompactor fallbackCompactor = new StringReplaceJsonCompactor();
+    private static final StringReplaceJsonCompactor fallbackCompactor = new StringReplaceJsonCompactor();
 
     private final String replacement;
     private final Set<String> fields;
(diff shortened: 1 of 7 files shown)


logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
```diff
@@ -47,17 +47,16 @@ public String filter(@Nullable final String contentType, final String body) {
...
-                while (true) {
-                    JsonToken nextToken = parser.nextToken();
-                    if (nextToken == null) {
-                        break;
-                    }
+            try (final JsonParser parser = factory.createParser(body);
+
+                 final CharArrayWriter  writer = new CharArrayWriter(body.length() * 2); // rough estimate of final size
+
+                 final JsonGenerator generator = factory.createGenerator(writer)){
```
kasmarian, line 56: Shouldn't generator also be closed?

kasmarian (commented):
Thank you for the PR, @achhibi ! I think the refactoring is really helpful here. I just left one comment in json filter

logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
kasmarian: Ah, I just noticed that it's inside the try block. The spacing confused me at first.

logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
```diff
@@ -47,17 +47,16 @@ public String filter(@Nullable final String contentType, final String body) {
 
     public String filter(final String body) {
         try {
```
kasmarian, line 49: Could probably now combine the two try-catch blocks.

kasmarian: Another observation: the tests fail because you call `writer.toString();`, but `generator` wasn't flushed yet. It probably make sense to still keep 2 try-cache blocks, but to move the `writer` creation to the outer one, and the return statement after the inner try-catch block if closed.

achhibi: @kasmarian 

Thanks for the feedback! 
I've reordered the 'try' blocks and fixed the unit tests.

kasmarian:
:+1:

PR #1756, opened 2024-02-01
Refactor HTTP status codes with HttpStatus Enum

achhibi:
## Description

This pull request introduces a refactor by replacing the existing switch statement handling HTTP status codes with a dedicated `HttpStatus `  Enum. The enum encapsulates each status code, providing a more structured and maintainable approach to handling HTTP status codes. The refactor covers standard HTTP status codes. Notably, this design facilitates the addition of new status codes in the future, promoting extensibility and ease of maintenance.

## Motivation and Context

The motivation behind this change is to enhance code organization and readability. The existing switch statement for HTTP status codes is lengthy and can be challenging to maintain. The introduction of the `HttpStatus`  enum streamlines the code, making it more readable and modular.

A big thank you to the Zalando team for their continuous support and inspiration.

## Types of changes

- [x]  Refactoring (improvements to code structure without changing functionality)

## Checklist:


logbook-api/src/main/java/org/zalando/logbook/HttpStatus.java
```diff
@@ -0,0 +1,100 @@
...
+
+    private final int code;
+    private final String reasonPhrase;
+
+    HttpStatus(int code, String reasonPhrase) {
+        this.code = code;
+        this.reasonPhrase = reasonPhrase.intern();
+    }
+
+    public static Map<Integer, String> toMap() {
```
kasmarian, line 97: As this map is created every time to just make a single lookup, what do you think about having a method that will just return the reason phrase by code? It will look similar, but will not collect the results in a map, and will just apply the filtering directly here. 
It'll help with readability a bit, It's not obvious what the value in the map is, without looking into this file. `HttpStatus.toMap().get(getStatus())` doesn't tell me that I should get the reason phrase, but `HttpStatus.getReasonPhraseByCode(int code)` will.

logbook-api/src/main/java/org/zalando/logbook/HttpStatus.java
```diff
@@ -0,0 +1,100 @@
+package org.zalando.logbook;
+
+import lombok.Getter;
+
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Getter
+public enum HttpStatus {
```
kasmarian, line 10: I'd not make it public at the start, as we don't need this enum in other parts for now. Reverting this will be hard.

logbook-api/src/main/java/org/zalando/logbook/HttpStatus.java
achhibi: @kasmarian 

 You're absolutely right. I've made the necessary changes.

logbook-api/src/main/java/org/zalando/logbook/HttpStatus.java
achhibi: Thank you for the suggestion!

kasmarian:
👍

kasmarian:
Thank you for the PR!

PR #1759, opened 2024-02-05
feat(logbook-spring-boot-autoconfigure): header include/exclude filters

p-8-z:
Added header filtering capability to `logbook-spring-boot-autoconfigure` module.

## Description

Added header filtering capability to `logbook-spring-boot-autoconfigure` module for spring-boot autoconfiguration.
Example of usage:
```
logbook.predicate.exclude[0].path=/actuator/**
logbook.predicate.exclude[1].headers.host=localhost
```
This configuration will create predicate that will log requests that wont contain `/actuator/**` path in url or 
header with key `host` and value `localhost`.

## Motivation and Context

I'd like to filter out some request/response logs by header values.

## Types of changes

## Checklist:

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/core/Conditions.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/core/Conditions.java
@@ -100,8 +100,8 @@ public final class Conditions {
             @Nonnull final String key,
             final String value) {
 
-        if (Objects.isNull(value)){
-            return withoutHeader(key);
+        if ("*".equals(value)){
+            return header(key, $ -> true);
         }
         return header(key, value);
     }
--- a/logbook-core/src/main/java/org/zalando/logbook/core/Conditions.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/core/Conditions.java
@@ -7,8 +7,10 @@ import org.zalando.logbook.RequestURI;
 import org.zalando.logbook.common.Glob;
 import org.zalando.logbook.common.MediaTypeQuery;
 
+import javax.annotation.Nonnull;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Objects;
 import java.util.function.BiPredicate;
 import java.util.function.Function;
 import java.util.function.Predicate;
@@ -92,6 +94,16 @@ public final class Conditions {
 
     public static <T extends HttpMessage> Predicate<T> withoutHeader(final String key) {
         return message -> !message.getHeaders().containsKey(key);
+    }
+
+    public static <T extends HttpRequest> Predicate<T> conditionalHeader(
+            @Nonnull final String key,
+            final String value) {
+
+        if (Objects.isNull(value)){
+            return withoutHeader(key);
+        }
+        return header(key, value);
     }
 
 }
--- a/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
+++ b/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
@@ -70,6 +70,9 @@ import org.zalando.logbook.spring.LogbookClientHttpRequestInterceptor;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.Collections;
+import java.util.Optional;
+import java.util.Objects;
 import java.util.List;
 import java.util.Set;
 import java.util.TreeSet;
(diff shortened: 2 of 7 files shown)


logbook-core/src/main/java/org/zalando/logbook/core/Conditions.java
```diff
@@ -94,4 +96,14 @@ public static <T extends HttpMessage> Predicate<T> withoutHeader(final String ke
         return message -> !message.getHeaders().containsKey(key);
     }
 
+    public static <T extends HttpRequest> Predicate<T> conditionalHeader(
```
kasmarian, line 99: Can't say I'm convinced that this separate method is needed to be honest. You can achieve the same result with `header(final String key, final Predicate<String> predicate)` method.

kasmarian (commented):
Thank you for the PR, @p-8-z! I think overall the idea is very nice and would be a good extension to the Logbook.

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -177,24 +180,32 @@ private Predicate<HttpRequest> mergeWithIncludes(final Predicate<HttpRequest> pr
     }
 
     private Predicate<HttpRequest> convertToPredicate(final LogbookProperties.LogbookPredicate logbookPredicate) {
-        final String predicatePath = logbookPredicate.getPath();
-        final List<String> predicateMethods = logbookPredicate.getMethods();
+        Optional<LogbookProperties.LogbookPredicate> oLogbookPredicate = Optional.of(logbookPredicate);
```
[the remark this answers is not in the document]
p-8-z: Sure thing, added.
If I forgot any use-case or broke you code styles just let me know, I'll fix it.

logbook-core/src/main/java/org/zalando/logbook/core/Conditions.java
p-8-z: `conditionalHeader` method wraps `withoutHeader` and `header` logic, or it can be extracted outside, whatever you prefer just let me know.

logbook-core/src/main/java/org/zalando/logbook/core/Conditions.java
```diff
@@ -94,4 +96,14 @@ public static <T extends HttpMessage> Predicate<T> withoutHeader(final String ke
...
+            @Nonnull final String key,
+            final String value) {
+
+        if (Objects.isNull(value)){
+            return withoutHeader(key);
```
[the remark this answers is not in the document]
p-8-z: I don't think you meant Regex pattern so I just hardcoded star character that checks key and ignores value.
```java
if ("*".equals(value)){
    return header(key, $ -> true);
}
```
If this was not what you wanted just let me know.
Maybe this star use-case should be documented in `README.md` separately, what do you propose?

logbook-spring-boot-autoconfigure/src/test/java/org/zalando/logbook/autoconfigure/IncludeTest.java
```diff
@@ -21,89 +23,176 @@
...
+    @TestConfiguration
+    public static class PathConfig {
+        @Bean
+        public Predicate<HttpRequest> condition() {
+            return exclude(requestTo("/api/admin/**"));
+        }
+    }
+
+    @TestConfiguration
+    public static class HeaderConfig {
```
kasmarian, line 184: It would be also good to use test `application-exclude.yml` and `application-include.yml` files for the tests instead of defining the predicate beans. This would test that the autoconfiguration logic that you adjusted `convertToPredicate` method works as expected. It will also work as a good example how this configuration can look like together with the rest of predicates. 

The `PathConfig` was a way to test how custom predicates work in a combination with the ones from .yml ones.

kasmarian: Speaking about the test configuration in .yaml files. Would be nice to have examples of purely header predicates as well as the combination with other predicates. 
E.g. 
```
logbook:
  include:
  - /api/**
  predicate:
    include:
      - path: /another-api
      - path: /yet-another-api
        methods:
          - DELETE
      - methods:
          - PUT
      - path: /secured-api
        headers: 
           authorization: *
      - path: /only-gets-with-json
        method: GET
        headers: 
           accept: application/json
```

logbook-core/src/main/java/org/zalando/logbook/core/Conditions.java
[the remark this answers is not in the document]
kasmarian: I think that would be fine to have it hardcoded, yes. We can extend it to regex in the future, not sure if this is really needed at this point. 

> Maybe this star use-case should be documented in README.md separately, what do you propose?

I think it does deserve an description, yes. Could you adjust the [Configuration](https://github.com/zalando/logbook?tab=readme-ov-file#configuration) table to have this part explained? `logbook.predicate.include` and `logbook.predicate.exclude` properties descriptions should be updated to include the header predicates configuration.

PR #1819, opened 2024-04-29
Fix Ktor plugin to handle requests with empty body

### LINKED ISSUE #1627: logbook-ktor-server removes Content-Type header from response
When using `logbook-ktor-server` with Ktor [Content Negotation](https://ktor.io/docs/serialization.html) server plugin, the `Content-Type` header is not returned in the response.

## Description
[Content Negotation](https://ktor.io/docs/serialization.html) feature for Ktor server automatically detects the `Content-Type` based on the data returned.
For instance, configuring it with Jackon as a deserializer allows Ktor server to add the `Content-Type: application/json` header in the response.
However, also installing `logbook-ktor-server` result in the header being removed in the response.

## Expected Behavior
The `Content-Type: application/json` header should be returned in the response

## Actual Behavior
The `Content-Type` header is missing in the response

## Steps to Reproduce
1. Create a simple Ktor server app with these features installed:
```
install(ContentNegotiation) {
        jackson()
}
install(LogbookServer) {
        logbook = defaultLogbook
}
```
2. Add a route that returns some JSON object
```
data class SomeData(val someProp: String)
route("/test") {
     get {
       call.respond(SomeData("someValue")) 
     }
}
```
4. Run the server then call the retour 
```
$curl http://localhost:8080/test
```

## Context
I had some integration tests assertions with RestAssured that failed after adding `logbook-ktor-server` because the `Content-Type` header was no longer returned, thus RestAssured is no longer able to parse the response.

## Your Environment
ktor 2.3.3
logbook 3.4.0

I haven't had time to make a sample yet sorry.

msdousti:
This PR registers the Ktor plugin interceptor in the right pipeline, so that requests with empty body (such as GET) are handled properly.

## Description
As analyzied in [this comment](https://github.com/zalando/logbook/issues/1627#issuecomment-1726450902), a request with empty body (such as a GET request) does not trigger the [ApplicationReceivePipeline](https://api.ktor.io/ktor-server/ktor-server-core/io.ktor.server.request/-application-receive-pipeline/index.html). This PR fixes the issue by registering another interceptor into [ApplicationCallPipeline](https://api.ktor.io/ktor-server/ktor-server-core/io.ktor.server.application/-application-call-pipeline/index.html).

To allow for the request body to be processed by both our plugin and the application itself, [DoubleReceive plugin﻿](https://ktor.io/docs/server-double-receive.html) was added. (Thanks to @grassehh for introducing this.)

## Motivation and Context
Fixes #1627 

## Types of changes

## Checklist:


logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/LogbookServer.kt
```diff
@@ -32,25 +29,37 @@ class LogbookServer(
...
             val plugin = LogbookServer(config.logbook)
 
-            pipeline.receivePipeline.intercept(ApplicationReceivePipeline.Before) {
+            // This interceptor is executed before ApplicationReceivePipeline,
+            // and regardless of whether the request has body or not
+            pipeline.intercept(ApplicationCallPipeline.Call) {
                 val request = ServerRequest(call.request)
                 val requestWritingStage = plugin.logbook.process(request)
-                val proceedWith = when {
-                    request.shouldBuffer() && !call.request.receiveChannel().isClosedForRead -> {
```
grassehh, line 44: Any reason you removed the `request.shouldBuffer()` check? As I think I understand, the `Logbook.process()` decides wether or not the request should be buffered by changing the state of the `HttpRequest` object.

logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/LogbookServer.kt
```diff
@@ -32,25 +29,37 @@ class LogbookServer(
...
+                    val content = call.request.receiveChannel().readBytes()
+                    request.buffer(content)
+                    ByteReadChannel(content)
+                } else {
+                    ByteReadChannel.Empty
                 }
+
                 val responseProcessingStage = requestWritingStage.write()
                 call.attributes.put(responseProcessingStageKey, responseProcessingStage)
+                call.attributes.put(proceedWithKey, proceedWith)
```
grassehh, line 57: Do we actually need this attribute ? Isn't it still enough to add the content in the request buffer ?
Changing the type of interceptor might be enough for `responseProcessingStageKey` to always be filled inside the other one :)

logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/LogbookServer.kt
```diff
@@ -32,25 +29,37 @@ class LogbookServer(
...
+            }
+
+            // This interceptor is executed only when the request has body
+            pipeline.receivePipeline.intercept(ApplicationReceivePipeline.Transform) { subject ->
```
grassehh, line 61: Similarly to my comment on the attribute, do we actually need this pipeline interceptor ?

grassehh (commented):
Thanks for the MR !

logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/LogbookServer.kt
msdousti: It wasn't used anywhere except in this piece of code. Once I re-ordered the code in the way it is, I didn't see ay use for it and all the tests pass.

I'm not sure what it does. `Logbook.process()` is called before, so it cannot depend on it.

Can you provide a unit-test that fails without it?

logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/LogbookServer.kt
msdousti: I really struggled and couldn't due to the way Ktor is design. Please read the PR description carefully as why I did it the way I did.

Needless to say, I may have missed something, so if you have a concrete suggestion, by all means go ahead and add a comment.

logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/LogbookServer.kt
msdousti: Same as above :))

logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/LogbookServer.kt
```diff
@@ -32,25 +29,37 @@ class LogbookServer(
...
         override val key: AttributeKey<LogbookServer> = AttributeKey("LogbookServer")
+
         override fun install(pipeline: Application, configure: Config.() -> Unit): LogbookServer {
             val config = Config().apply(configure)
             val plugin = LogbookServer(config.logbook)
 
-            pipeline.receivePipeline.intercept(ApplicationReceivePipeline.Before) {
+            // This interceptor is executed before ApplicationReceivePipeline,
+            // and regardless of whether the request has body or not
+            pipeline.intercept(ApplicationCallPipeline.Call) {
```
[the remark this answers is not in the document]
msdousti: Fixed in a58913b.

logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/LogbookServer.kt
grassehh: I don't have any unit-test, I was just wondering, because I thought call to `Logbook.process` configures the request `State` depending on the configuration (including the buffer state).
But I might just be wrong, I just guessed it by reading the code.

logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/LogbookServer.kt
grassehh: Well I agree Ktor design seems a little weird when it comes to creating plugins.. They simplified it a little bit in the new API but reading request/response bodies is still a headache.
I will experiment using the [`ResponseObserver`](https://api.ktor.io/ktor-client/ktor-client-core/io.ktor.client.plugins.observer/-response-observer.html) plugin to deal with the response (which could help avoid using internal API from `request.content`.

Back to your question, I'm just wondering if you have spimply tried to test it without having this attribute and without the `ApplicationReceivePipeline.Transform` phase interceptor ?

Now that you managed to have the `responseProcessingStageKey` always added wether or not a body is received, this might not be useful anymore.

logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/LogbookServer.kt
grassehh: Alright, I'm not the Logbook expert there so I'll let you and the others decide about it ^^

logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/LogbookServer.kt
grassehh: Hello @msdousti I have pushed a sample here: ([url](https://github.com/grassehh/issues/blob/main/zalando-logbook-1627/src/test/kotlin/com/grassehh/ServerTest.kt))

Basically, I only changed the intecepted phase to `ApplicationCallPipeline.Monitoring` and removed the `proceedWith` which is not necessary and it seems to work.

msdousti: @grassehh 
I replaced my code with your `LogbookServer`, and ran the tests. Two tests fail as follows:

image

msdousti: Basically, if you don't use `proceedWith`, the request body is consumed by th "request logger", and it is no longer visible down the pipeline.

grassehh: I see @msdousti you are right. So I found a way by installing the [DoubleReceive](https://ktor.io/docs/server-double-receive.html) inside the custom plugin, and then simply calling `call.receiveChannel().readBytes()` inside the interceptor in order to read the request body.
I updated my sample and added the `LogbookServerTest` and they all pass now :)

msdousti: Thanks, good to know about this plugin!

I've a question though: Is it worth adding another dependency to the project, and add another plugin to the pipeline, while my code is essentially doing that without all the fuss?

This is not a rhetorical question, I'm genuinely interested in the answer. The `DoubleReceive` plugin makes the code somewhat more readable, but the extra dependency is a downside, and the extra plugin **may** also make it less performant/more memory hungry (not sure).

grassehh: I understand. In my opinion I can see several reason of using that plugin:
- It is an official one, thus it could might make it easier to maintain the custom plugin
- Ktor is a complex and sensitive machinery so we don't know if there could be any side-effect or things we didn't think of about reading requests bodies. As an example you can check out my other [issue](https://github.com/zalando/logbook/issues/1822) in the `logbook-ktor-client` that can lead to some deadlocks when reading the response.

Now if you really don't want to add this dependency, which I can understand, it is worth taking a look at the `DoubleReceive` implementation.
Also, could it be a requirement for the users to add the plugin themselves ? We could eventually have two versions of the `logbook-ktor-server`. One that is fully custom and another that relies on the `Ktor` official plugin.

msdousti: Thanks for the explanations. I trust if we load the plugin and the user loads it independently, there will be no conflicts? 

@kasmarian Could you chime in here: Comparing two approaches above (with or without additional plugin), which one do you prefer?

kasmarian: I agree with @grassehh on this. The `DoubleReceive﻿` plugin seems like a good way to go. Overall Logbook has to overcome issues related to double-reads of request/response streams in various modules, and here the native plugin provides a solution out of the box.

msdousti: Thanks! Then I'll make the necessary changes and push sometime tomorrow.

msdousti:
Applied the `DoubleReceive` plugin﻿, and also migrated to the new Ktor API.

Everything's fine, except coverage fails due to a [JaCoCo issue with Kotlin](https://stackoverflow.com/a/53568092/459391) coroutines 😒

image

image

grassehh:
> Applied the `DoubleReceive` plugin﻿, and also migrated to the new Ktor API.
> 
> Everything's fine, except coverage fails due to a [JaCoCo issue with Kotlin](https://stackoverflow.com/a/53568092/459391) coroutines 😒
> 
> image
> 
> image

Good job ! It looks good with the new API too !

msdousti:
After spending sometime, I found a workaround: If the function JaCoCo is nagging about is marked as `suspend`, the issue will be resolved.

Thanks a lot @grassehh for allthe effort you put into this ❤️ 

@kasmarian It's ready to be reviewed / merged 🙏

kasmarian:
Looks very well to my limited Ktor knowledge, thank you, @msdousti!

kasmarian:
:+1:

kasmarian:
@msdousti could you make the commits to be signed for this PR?

msdousti:
@kasmarian Done 🙂

kasmarian:
:+1:

grassehh:
Thank you for the fix @msdousti @kasmarian :) 
Please if you can keep up the rythm, take a look at my other issues about client and server plugin: [1821](https://github.com/zalando/logbook/issues/1821), [1822](https://github.com/zalando/logbook/issues/1822), [1823](https://github.com/zalando/logbook/issues/1823)

For each of them I have already provided a possible solution and working example. I'd love to see them as well in the future release !
Thank you !

Issue #1822, opened 2024-04-29
[logbook-ktor-client] Dead locks on large request/response bodies


grassehh:
## Description
Using `logbook-ktor-client` plugin, if you send/receive request/response with large string bodies (approx ~4000+ characters), the request will never be sent or the response will never be received.
The issue seems to only occur when using streamed content serialized/deserialized through `jackson`.

## Expected Behavior
Request/response with large content should be sent/received

## Actual Behavior
Requests threads will be stuck and will never be sent.
Responses will timeout.

## Possible Fix
Fix available here: [https://github.com/grassehh/issues/blob/main/zalando-logbook-1822/src/test/kotlin/com/grassehh/logbook/LogbookClient.kt](https://github.com/grassehh/issues/blob/main/zalando-logbook-1822/src/test/kotlin/com/grassehh/logbook/LogbookClient.kt)

In summary, as for the request, the root cause seems to reside in the `Unconfined` dispatcher used in the [`readBytes`](https://github.com/zalando/logbook/blob/f29dad2ca372dffa389d296bd0acf8ed352a3da9/logbook-ktor-common/src/main/kotlin/org/zalando/logbook/common/ContentUtils.kt#L19) method. I don't have a clear explaination of what it does not work, but using the default or `IO` seems to fix it.

As for the response, after comparing with the official Ktor [`ObserverResponse`](https://github.com/ktorio/ktor/blob/main/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/observer/ResponseObserver.kt) plugin, simply executing the process inside a coroutine seems to fix the issue.
Another point to not is that it is recommanded to call the `dicard()` method on the `ByteReadChannel` once we are done with it.

In the fix, I eventually also placed the request process inside a coroutine aswell improve overall request/response times.

## Steps to Reproduce
- Checkout this repository: [https://github.com/grassehh/issues](https://github.com/grassehh/issues)
- Run the `ClientTest` test inside the **zalando-logbook-1822** folder

The test will fail. If you replace the `LogbookClient` with `com.grassehh.logbook.LogbookClient`, the test will pass.

## Your Environment
- Ktor 2.3.10
- Logbook 3.8.0

kasmarian:
Thank you for the well defined description of the issue, @grassehh ! It turned out that the only part that mattered was to use the coroutine scope from `scope: HttpClient`. I added the changes in https://github.com/zalando/logbook/pull/1832

kasmarian:
[...]

PR #1832, opened 2024-05-13
wrap ktor client response handling into HttpClient context

kasmarian:
## Description
Wrapping Ktor client response handling in the coroutine context of HttpClient scope. 

## Motivation and Context

Addresses #1822 

## Types of changes

## Checklist:

### DIFF
--- a/logbook-ktor-client/src/main/kotlin/org/zalando/logbook/client/LogbookClient.kt
+++ b/logbook-ktor-client/src/main/kotlin/org/zalando/logbook/client/LogbookClient.kt
@@ -14,6 +14,8 @@ import io.ktor.http.content.OutgoingContent
 import io.ktor.util.AttributeKey
 import io.ktor.util.InternalAPI
 import io.ktor.util.split
+import io.ktor.utils.io.discard
+import kotlinx.coroutines.launch
 import org.apiguardian.api.API
 import org.apiguardian.api.API.Status.EXPERIMENTAL
 import org.zalando.logbook.Logbook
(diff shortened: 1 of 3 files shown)


grassehh:
Thanks for your review :) 
However, are you sure the `Dispatcher.Unconfined` is not an issue too though? Because somehow [this test](https://github.com/grassehh/issues/blob/30a6f58be310473edf2a138b91bcf2a6fd003ded/zalando-logbook-1822/src/test/kotlin/com/grassehh/ClientTest.kt#L45) never responds when I try with your code. And when I just change the `Dispatcher` to `IO` it works again.

Note that in Kotlin [CoroutineDispatcher](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-dispatcher/) documentation they indicate that _"The Unconfined dispatcher should not normally be used in code"_

kasmarian:
Interesting, I tried to replicate this test with `Should log request and response with big body` one. And it was failing before the fix. But I didn't re-run the tests on your repo. Let me check that.

I'm not sure about the IO Dispatcher just because not all clients may want it to be used. I see that if I change the context to the current one (coroutineContext) your test also passes. I wonder why my test didn't capture this? :thinking:

grassehh:
> Interesting, I tried to replicate this test with `Should log request and response with big body` one. And it was failing before the fix. But I didn't re-run the tests on your repo. Let me check that.
> 
> I'm not sure about the IO Dispatcher just because not all clients may want it to be used. I see that if I change the context to the current one (coroutineContext) your test also passes. I wonder why my test didn't capture this? 🤔

Good catch ! Seems like a better solution indeed.

ChristianLohmann:
:+1:

PR #1839, opened 2024-05-16
Improve performance of query filters (especially on large bodies)

### LINKED ISSUE #1838: Default `QueryFilter` has a very slow RegEx on certain inputs
I was sending some requests with `curl` to an application that had logbook enabled and noticed quite terrible performance. 

`curl` by default will send data with a `Content-Type: application/x-www-form-urlencoded`, by default this triggers a QueryFilter that will try to parse the content and look for `client_secret` and `password` in the body and obfuscate that. 

In the end this will end up parsing the body with this Pattern: https://github.com/zalando/logbook/blob/8fa25bb682d7c3d9ef5054d7f7efe4c2ea760ece/logbook-core/src/main/java/org/zalando/logbook/core/QueryFilters.java#L60-L60

I debugged the regex (with PCRE, not Java but seems to have roughly the same performance) and the amount of steps is exponential with the body size. This matches my experience as well. Sending a 100K body took about 4s if I recall correctly, sending 1MB would take _minutes_.

Btw: it largely depends on the body. If your body actually has a lot of parameters, so lots of stuff separated by `&`, it's much less of a problem. If it has just a large text of some sort, it will blow up.

## Description
Why I consider this a bug: I think there is no reason for a regex that will has `O(n^2)` complexity.

## Expected Behavior
To be faster! :-D 

## Actual Behavior
It's slow.

## Steps to Reproduce

1. Have an application that you can send big request bodies to
2. Send a request with a large body
3. _wait_

aukevanleeuwen:
Fixes: #1838

Rewrite two `QueryFilter`s with an implementation based on `StringTokenizer` rather than a possible `O(n^2)` regex (for more information see the linked issue).

## Description
I've looked at the description for a `application/x-www-form-urlencoded` encoded body here: https://url.spec.whatwg.org/#urlencoded-parsing and tried to follow those lines. I did *not* put in the percentage encoding, since that wasn't in the current implementation and I didn't want to break backwards compatibility.

I've added a few test cases that weren't covered (and also not all handled correctly in the previous implementation), but I think they are rather obscure. Where normally you have `key=value`, it's also possible to just have `key`, if there is no `=` there, `key` is treated as the key and there is no value. Also `key=` and `=value` are allowed even, resulting in an empty key or empty value respectively. 

## Motivation and Context
Fixes #1838 

## Types of changes

## Checklist:

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/core/QueryFilters.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/core/QueryFilters.java
@@ -3,12 +3,13 @@ package org.zalando.logbook.core;
 import org.apiguardian.api.API;
 import org.zalando.logbook.QueryFilter;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.StringTokenizer;
+import java.util.function.BiFunction;
 import java.util.function.Predicate;
 import java.util.function.UnaryOperator;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
-import static java.util.regex.Pattern.compile;
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 import static org.apiguardian.api.API.Status.MAINTAINED;
 import static org.apiguardian.api.API.Status.STABLE;
@@ -45,7 +46,6 @@ public final class QueryFilters {
         return replaceQuery(name::equals, replacementFunction);
     }
 
-
     @API(status = EXPERIMENTAL)
     public static QueryFilter replaceQuery(
             final Predicate<String> predicate, final String replacement) {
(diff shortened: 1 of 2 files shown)


aukevanleeuwen:
@whiskeysierra perhaps? :-)

2k16daniel:
Hoping that this commit will be merged

kasmarian:
Thank you for the PR @aukevanleeuwen, it looks great, with `StringTokenizer` the performance will be way faster!

kasmarian:
In order for the PR to be compliant with our merging rules, could you please make sure your commits are signed?

aukevanleeuwen:
@kasmarian ✅

kasmarian:
:+1:

lukasniemeier-zalando:
:+1:

Issue #1868, opened 2024-07-10
ktor-client: response replaces request body


noffke:
## Description
When using logbook-ktor-client together with a custom sink that merges request and response, the request body is always the response body, i.e. response overwrites request. I think this is caused by the `State` (https://github.com/zalando/logbook/blob/main/logbook-ktor-common/src/main/kotlin/org/zalando/logbook/common/State.kt) subclasses being `object`s, which makes them a singleton.

## Expected Behavior
I expect that at the point of calling the sink, the request contains the request body and the response contains the response body.

## Actual Behavior
At the point of calling the sink, the request's body is that of the response.

## Possible Fix
Make `State` subclasses a `class` that has to be instanciated so that each request and response has their own `State` instance.

## Steps to Reproduce
Please see `LogbookClientWithSinkTest` in my upcoming PR.

## Context
I'm trying to log http requests and responses when using ktor client

## Your Environment
Logbook version 3.9.0

noffke:
@kasmarian thank you for approving and merging my PR!

PR #1869, opened 2024-07-10
Fix for #1868 - make State subclasses real classes

noffke:
make State subclasses real classes (not objects, i.e. singletons) so that each request and response can have their own state instance

## Description
State subclasses are `object`s which make them singletons. Hence, all requests and responses share the same state instance. I think the subclasses should be `class`es, so that requests and responses have their own `State` instances.

## Motivation and Context
Please see #1868

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ?] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ x] I have added tests to cover my changes.


msdousti (approved):

kasmarian:
Thank you for the fix, @noffke !

Issue #1384, opened 2022-12-07
Quarkus Resteasy endpoint returning 500 to client after HEAD request because of Logbook NullPointerException


fenix-hub:
I'm writing a webservice using Logbook to log requests and responses in a Quarkus application.
I'm using Resteasy implementation for the RESTful endpoints.
Exposing an `@HEAD` endpoint causes Logbook to throw a `NullPointerException` right after logging the response.
Even though the request is sucessfully handled, whatever the response code is (be it `200` or `404`) the exception causes the client to not receive the webservice response, instead 500 is received.
 
## Description
By my understanding, Logbook uses `TeeStream` to overcome the request/response stream to be emptied on requests and responses.
This is default behavior at least; however this should not be *always* done, because of some requests type not supporting (or not expecting) a response body to be returned by standard.
An `HEAD` request is one of these. 

## Expected Behavior
Logbook should just log request/response flow without trying to catch bodies in `HEAD` requests, eventually directly inspecting the body itself or just checking if the `Content-Type` header is present (it isn't for `HEAD` requests).

## Actual Behavior
Logbook logs request/response correctly, but a `NullPointerException` is thrown and causes any client to receive 500 instead of the actual response.

<details>
<summary>Stacktrace</summary>

```
2022-12-07 10:53:01,001 TRACE [org.zal.log.Logbook] (executor-thread-0) {"origin":"local","type":"response","correlation":"c0f740b400ab062d","duration":566,"protocol":"HTTP/1.1","status":204}
2022-12-07 10:53:01,012 DEBUG [io.ver.ext.web.RoutingContext] (executor-thread-0) RoutingContext failure (500): java.lang.NullPointerException
        at org.zalando.logbook.jaxrs.TeeOutputStream.close(TeeOutputStream.java:39)
        at io.quarkus.resteasy.runtime.standalone.VertxHttpResponse.finish(VertxHttpResponse.java:145)
        at io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.dispatch(VertxRequestHandler.java:166)
        at io.quarkus.resteasy.runtime.standalone.VertxRequestHandler$1.run(VertxRequestHandler.java:93)
        at io.quarkus.vertx.core.runtime.VertxCoreRecorder$14.runWith(VertxCoreRecorder.java:576)
        at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2449)
[...]

PR #1909, opened 2024-09-22
Fix JAX-RS filter NPE for empty entities

msdousti:
In JAX-RS, when Logbook is registered as a response filter, and the response entity is empty, an NPE is thrown.
Same seems to be the case for request filter and an empty request.

## Description
The issue was reported to us but marked as not reproducible (see below). However, I could reproduce it here:
https://github.com/msdousti/quarkus-logbook

If you use install this branch to your local Maven, it works as expected. However, if you change `logbook-bom.version` in the above repo to `3.9.0` (current Logbook version), and run the test:

```bash
curl -iI http://localhost:8080/hello
```

```
HTTP/1.1 500 Internal Server Error
content-type: text/plain; charset=utf-8
```

In the server logs:

```
2024-09-22 02:13:20,426 TRACE [org.zal.log.Logbook] (executor-thread-1) Incoming Request: f0ce7f5363945e86
Remote: localhost:8080
HEAD http://localhost:8080/hello HTTP/1.1
Accept: */*
Host: localhost:8080
User-Agent: curl/8.5.0

2024-09-22 02:13:20,433 TRACE [org.zal.log.Logbook] (executor-thread-1) Outgoing Response: f0ce7f5363945e86
Duration: 11 ms
HTTP/1.1 200 OK
2024-09-22 02:13:20,437 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (executor-thread-1) HTTP Request to /hello failed, error id: e5624c9d-1e32-48e8-b1b3-56183b37467c-1: java.lang.NullPointerException: Cannot invoke "java.io.OutputStream.close()" because "this.original" is null
        at org.zalando.logbook.jaxrs.TeeOutputStream.close(TeeOutputStream.java:39)
        at io.quarkus.resteasy.runtime.standalone.VertxHttpResponse.finish(VertxHttpResponse.java:145)
        at io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.dispatch(VertxRequestHandler.java:174)
...
```

## Root Cause

1. `LogbookServerFilter#filter` calls [response.expose()](https://github.com/zalando/logbook/blob/05ede67de9953342d6fb8992a56e73b02915dba7/logbook-jaxrs/src/main/java/org/zalando/logbook/jaxrs/LogbookServerFilter.java#L51).
2. `expose` calls [state.buffer(context)](https://github.com/zalando/logbook/blob/05ede67de9953342d6fb8992a56e73b02915dba7/logbook-jaxrs/src/main/java/org/zalando/logbook/jaxrs/LocalResponse.java#L140)
3. `buffer` is implemented as follows:

    https://github.com/zalando/logbook/blob/05ede67de9953342d6fb8992a56e73b02915dba7/logbook-jaxrs/src/main/java/org/zalando/logbook/jaxrs/LocalResponse.java#L57-L62

4. When the response has no entity, then `context.getEntityStream()` is null. As such, `TeeOutputStream` will have a `null`  stream (called "original"). When [original.close](https://github.com/zalando/logbook/blob/05ede67de9953342d6fb8992a56e73b02915dba7/logbook-jaxrs/src/main/java/org/zalando/logbook/jaxrs/TeeOutputStream.java#L39) is called, an NPE is thrown.

The exact same issue happens for Request Filters, when [request.expose](https://github.com/zalando/logbook/blob/05ede67de9953342d6fb8992a56e73b02915dba7/logbook-jaxrs/src/main/java/org/zalando/logbook/jaxrs/LogbookClientFilter.java#L44) is called.

## Fix
The fix is to call expose methods conditionally, only when the request/response entity exists.

## Test
The existing tests use `ResourceConfig` ([example](https://github.com/zalando/logbook/blob/05ede67de9953342d6fb8992a56e73b02915dba7/logbook-jaxrs/src/test/java/org/zalando/logbook/jaxrs/ClientAndServerWithoutBodyTest.java#L55)), and they don't seem to suffer from this issue.

[Quarkus does not seem to support ResourceConfig](https://stackoverflow.com/a/71642870/459391), so in my Quarkus project, I used a `@Provide` (see [LogbookFilterSetup](https://github.com/msdousti/quarkus-logbook/blob/main/src/main/java/io/msdousti/LogbookFilterSetup.java)).

I wasn't able to write an automated test to cover this, but the existing tests plus cover for all branches already. Suggestions are welcome.

## Motivation and Context

https://github.com/zalando/logbook/issues/1384

## Types of changes

## Checklist:


msdousti:
:+1:

msdousti:
:+1:

logbook-jaxrs/src/main/java/org/zalando/logbook/jaxrs/LogbookClientFilter.java
```diff
@@ -37,11 +37,10 @@ public void filter(final ClientRequestContext context) throws IOException {
         if (context.hasEntity()) {
             context.setProperty("request", request);
             context.setProperty("write-request", stage);
+            request.expose();
```
kasmarian, line 40: I'm a bit puzzled why we'd only do with conditional buffering to a request here and to a response in the server filter? Shouldn't the same check be done on line 61 here
https://github.com/zalando/logbook/blob/c9e8ba9922e6f4e46afb487c4e2ac9ea69b4212d/logbook-jaxrs/src/main/java/org/zalando/logbook/jaxrs/LogbookClientFilter.java#L61

And line 35 in LogbookServerFilter?
https://github.com/zalando/logbook/blob/c9e8ba9922e6f4e46afb487c4e2ac9ea69b4212d/logbook-jaxrs/src/main/java/org/zalando/logbook/jaxrs/LogbookServerFilter.java#L35

msdousti24: You have a point.
The main issue is that the tests do not reflect what Quarkus does in reality. 

Debugging the tests, I noted that the default implementation of `State#buffer` is executed, which doesn't do anything: 

https://github.com/zalando/logbook/blob/05ede67de9953342d6fb8992a56e73b02915dba7/logbook-jaxrs/src/main/java/org/zalando/logbook/jaxrs/LocalResponse.java#L31-L33

But running with live Quarkus, the overridden implementation is picked, and that's the source of the NPE:

https://github.com/zalando/logbook/blob/05ede67de9953342d6fb8992a56e73b02915dba7/logbook-jaxrs/src/main/java/org/zalando/logbook/jaxrs/LocalResponse.java#L58-L62

I just found two examples of `expose` and fixed them, assuming they are the only ones. My real-world test was with `HEAD` so I only tested the "empty response" case.

I tried adding Quarkus machinery to our tests, but they always succeeded by following the default implementation of `State#buffer`.

kasmarian: Regarding the default implementation of State#buffer, shouldn't that be changed as soon as strategy.process() is called?
 
https://github.com/zalando/logbook/blob/65a12b97b8538a25b284e9edf107ea547dcb59c8/logbook-core/src/main/java/org/zalando/logbook/core/DefaultLogbook.java#L52

But I also couldn't find a way to break anything with this fix, maybe it's sufficient. 

Alternatively, what if the fix is applied to `TeeOutputStream` instead to tolerate `null` OutputStream parameter? 

Overall I'm good with the fix, just trying to understand how to make it regression tolerant and reproducible.

msdousti: I reverted all my changes, and went with your suggestion to make `TeeOutputStream` robust: It now accepts a nullable stream, and handles null properly.

msdousti:
:+1:

kasmarian (approved):

kasmarian:
:+1:

kasmarian:
@lukasniemeier-zalando @ChristianLohmann could you have a look?

lukasniemeier-zalando (approved):

Issue #1963, opened 2024-11-19
NullPointerException when trying to use webClient with wildfly in a gateway service


magyari44:
## Description
I am trying to use logbook for logging http requests and response in a gateway module, in such a way, that the gateway logs the incoming requests from the internet ("origin":"remote","type":"request") and the outgoing requests and responses ("origin":"local","type":"request"; "origin":"remote","type":"response") to inner-services, and at the end the outgoing response to the internet too ("origin":"local","type":"response").

This gateway module uses spring webClient for handling asynchronous requests for inner-services.

I have used the following dependencies, where I am not able to upgrade them:
```xml
<groupId>org.springframework.boot</groupId>
<artifactId>spring-boot-starter-parent</artifactId>
<version>2.4.13</version>
<relativePath/>

<dependency>
	<groupId>org.zalando</groupId>
	<artifactId>logbook-spring-boot-starter</artifactId>
	<version>2.15.0</version>
</dependency>
<dependency>
	<groupId>org.zalando</groupId>
	<artifactId>logbook-netty</artifactId>
	<version>2.15.0</version>
</dependency>
```

It works fine for logging all the information but I am getting a `NullPointerException` too, related to the LogbookFilter, and I am not sure why is this, and what causes it.

## Expected Behavior
 - Log everything, but do not throw `NullPoinerException`

## Actual Behavior

- It throws a NullPointerException:
Example:
```
2024-11-19 11:52:13 10:52:13,258 INFO  [stdout] (default task-1) 2024-11-19 10:52:13.238  INFO 185 --- [ default task-1] c.logbook.async.logbookdemo.DemoLogbook  : {"origin":"remote","type":"request","correlation":"d742767227a71985","protocol":"HTTP/1.1","remote":"172.23.0.1","method":"GET","uri":"http://localhost:8080/gateway/api","host":"localhost","path":"/gateway/api","scheme":"http","port":"8080","headers":{"Accept":["*/*"],"Accept-Encoding":["br, deflate, gzip, x-gzip"],"content-length":["0"],"Content-Type":["application/json"],"Cookie":["JSESSIONID=681FBEB58818F31657E3BA6C693072C1.5a7e07a78e2a; JSESSIONID=681FBEB58818F31657E3BA6C693072C1"],"host":["localhost:8080"],"User-Agent":["IntelliJ HTTP Client/IntelliJ IDEA 2024.2.4"]}}
[...]

PR #1997, opened 2025-01-06
Attempt to write a response log when flushBuffer() fails

kasmarian:
## Description
In cases when a reactive client closes a connection, flushBuffer() operation may throw a "java.io.IOException: Broken pipe" exception, that in its own turn leads to a connection close, and `LogbookAsyncListener.onComplete` call. But as this happens AFTER `ServletResponse.freeResources()` is called (by HttpServerExchange.endExchange() in case of undertow, for example), the `responseWritingStageSynchronizationName` attribute is not present anymore in the response object, which leads to an NPE. This fix attempts to overcome this edge case, by trying to write the response anyway, even after the flush fails.

I wasn't able to add a reproducible unit test with MockMvc for this use case so far, but added an isolated one. Any suggestions are welcome

## Motivation and Context

#1963 

## Types of changes

## Checklist:


msdousti:
I suggest having a mock-based test to reproduce the NPE, in order to quench the thirst for 100% coverage.

kasmarian:
> I suggest having a mock-based test to reproduce the NPE, in order to quench the thirst for 100% coverage.

added the test in https://github.com/zalando/logbook/pull/1997/commits/12c63a7b150552ea8396e628124a1f42a5476f4a ^^

ChristianLohmann (approved):

Issue #1876, opened 2024-08-09
Missing Logging for Sent Requests During ReadTimeoutException in WebClient


NavruzshoevDaniel:
## Description
When sending requests via WebClient, if the request is successfully sent but a `ReadTimeoutException` occurs, the event of the request being sent is not logged. Additionally, the logging of the request occurs when the response is returned, rather than at the moment the request is made.

## Expected Behavior
The request should be logged at the moment it is sent, regardless of whether a `ReadTimeoutException` occurs. 

## Actual Behavior
Currently, if a `ReadTimeoutException` is thrown, there is no log entry indicating that the request was sent. Logging occurs only when the response is received, which might never happen if a timeout occurs.

## Possible Fix
Log the request immediately when the function is called, not after the response is returned.

## Steps to Reproduce
1. Send a request via WebClient.
2. Simulate a scenario where the request is sent but the response is delayed or results in a `ReadTimeoutException`.
3. Check the logs to see that the request was not logged at the time it was sent.
4. Observe that the logging only occurs upon receiving a response (or not at all in the case of a timeout).

## Context
This bug affects the ability to track and debug issues related to outgoing requests, particularly when responses are delayed or result in timeouts.

## Your Environment
* Version used: org.zalando:logbook-spring-boot-webflux-autoconfigure:2.16.0

## Draft decision

Maybe you should add `doOnError`
```java

package org.zalando.logbook.spring.webflux;

import lombok.RequiredArgsConstructor;
import org.apiguardian.api.API;
import org.springframework.core.io.buffer.DefaultDataBufferFactory;
import org.springframework.http.HttpHeaders;
import org.springframework.web.reactive.function.client.ExchangeFilterFunction;
import org.springframework.web.reactive.function.client.ExchangeFunction;
import org.zalando.logbook.Logbook;
import reactor.core.publisher.Flux;
import reactor.core.publisher.Mono;

import static org.apiguardian.api.API.Status.EXPERIMENTAL;
import static org.springframework.http.HttpHeaders.TRANSFER_ENCODING;
import static org.zalando.fauxpas.FauxPas.*;

@RequiredArgsConstructor
@API(status = EXPERIMENTAL)
@SuppressWarnings({"NullableProblems"})
public class LogbookExchangeFilterFunction implements ExchangeFilterFunction {

    private final Logbook logbook;

    @Override
[...]

PR #2000, opened 2025-01-07
Log webclient request on response errors

kasmarian:
## Description
When WebClient's call times out, the request is not logged together with the response, as both operations are delayed by the callback approach.  I'm calling a `requestWritingStage.write();` on timeouts errors. 

## Motivation and Context
#1876

## Types of changes

## Checklist:

### DIFF
--- a/logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/LogbookExchangeFilterFunction.java
+++ b/logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/LogbookExchangeFilterFunction.java
@@ -1,6 +1,5 @@
 package org.zalando.logbook.spring.webflux;
 
-import io.netty.handler.timeout.TimeoutException;
 import lombok.RequiredArgsConstructor;
 import org.apiguardian.api.API;
 import org.springframework.core.io.buffer.DefaultDataBufferFactory;
@@ -35,12 +34,8 @@ public class LogbookExchangeFilterFunction implements ExchangeFilterFunction {
                         .body((outputMessage, context) -> request.body().insert(new BufferingClientHttpRequest(outputMessage, clientRequest), context))
                         .build()
                 )
-                .doOnError(throwingConsumer(throwable -> {
-                            if (throwable.getCause() instanceof TimeoutException) {
-                                requestWritingStage.write();
-                            }
-                        }
-                )).flatMap(throwingFunction(response -> {
+                .doOnError(throwingConsumer(throwable -> requestWritingStage.write()))
+                .flatMap(throwingFunction(response -> {
                     Logbook.ResponseProcessingStage responseProcessingStage = requestWritingStage.write();
 
                     ClientResponse clientResponse = new ClientResponse(response);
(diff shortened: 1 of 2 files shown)


logbook-spring-webflux/src/main/java/org/zalando/logbook/spring/webflux/LogbookExchangeFilterFunction.java
```diff
@@ -34,7 +35,12 @@ public Mono<org.springframework.web.reactive.function.client.ClientResponse> fil
                         .body((outputMessage, context) -> request.body().insert(new BufferingClientHttpRequest(outputMessage, clientRequest), context))
                         .build()
                 )
-                .flatMap(throwingFunction(response -> {
+                .doOnError(throwingConsumer(throwable -> {
+                            if (throwable.getCause() instanceof TimeoutException) {
```
kasmarian, line 39: Not sure if it makes sense to capture only TimeoutException here, or try to log the request either way. 
On one hand, if the exchange failed prior to the sending, it may create a false assumption that the request happened. On the other hand, there can be other use cases, when the exchange failed after the request is sent, that are not related to timeouts.

kasmarian: Plus, it looks like `io.netty.handler.timeout` package is not in the classpath, addressing TimeoutException would require adjustements to the dependency tree.

msdousti: I'm not super familiar with Netty; just asking out of curiosity. Looking at the `TimeoutException` inheritance tree:

```
java.lang.Object
  java.lang.Throwable
    java.lang.Exception
      java.lang.RuntimeException
        io.netty.channel.ChannelException
          io.netty.handler.timeout.TimeoutException
```

Would it make sense to generalize error handling to `ChannelException`?

kasmarian: Seems like `io.netty.channel.ChannelException` is also not in the classpath. But I imagine it's not much farther away from catching any runtime exception at this point.

msdousti: Makes sense. What if we generalize it to `RuntimeException` then? In case the request is sent successfully, but any exception happens then (timeout, connection closed, etc.), we still want to log the request. This way, we won't need to change the classpath.

kasmarian: So no condition for exceptions like here https://github.com/zalando/logbook/pull/2000/commits/77432d29014a57e8f6acd2252f58744df7845779 ?

msdousti: Yes, I think it's the easiest way to fix it 👍

msdousti (approved):

ChristianLohmann:
:+1:

Issue #1993, opened 2024-12-26
Truncation of significant digits in floating point numbers in json formatting


bogatikov:
When a JSON body contains a float value like 0.40000000000000002, the log records it as 0.4.
This occurs when the content type is application/json and the request is processed by JacksonJsonFieldBodyFilter.

```
Content-Type: application/json

{
  "amount": 0.40000000000000002
}
```

Log output: 
```
{
  "origin": "remote",
  "type": "request",
  "correlation": "9c6416830b522967",
  "protocol": "HTTP/1.1",
  "remote": "/127.0.0.1:53879",
  "method": "POST",
  "uri": "http://localhost:8080/helloworld",
  "host": "localhost",
  "path": "/helloworld",
  "scheme": "http",
  "port": "8080",
  "headers": {
    ....
    "Content-Type": [
      "application/json"
    ],
    ....
  },
  "body": {
    "amount": 0.4
  }
}
```

The original float value is truncated, making the logs inconsistent with the actual request. This can lead to confusion when analyzing logs, as the exact input is not preserved

## Expected Behavior
The log should retain and display the original float value exactly as it appeared in the request body, without altering the data.

## Actual Behavior
Float values are truncated to 16 significant digits when logged

## Possible Fix
The issue stems from how Jackson processes and copies JSON tokens. When Jackson encounters a float value, it parses it as a double, resulting in a loss of precision:
```
generator.copyCurrentEvent(parser);
```

Starting from Jackson 2.15+, the following method can be used to handle float values more accurately:
```
    public void copyCurrentEventExact(JsonParser p) throws IOException
```
This method treats float values as BigDecimal, preserving their full precision.

**Note:**
•	This change may impact performance.
•	It is recommended to make this behavior configurable, allowing users to choose between exact precision or standard copying for performance reasons.

## Steps to Reproduce
1.	Send a request with a JSON body containing a float value with more than 16 significant digits.

## Context
The current behavior results in misleading log entries when float values are involved, potentially affecting debugging and auditing processes

## Your Environment
•	Framework: Spring WebFlux
•	Logging Library: Logbook 3.10.0
•	Spring Versions: 3.0.4 / 3.4.0
•	Jackson Versions: 2.14, 2.15, 2.18

kasmarian:
Added a suggestion to allow specifying if float precision is needed in body filters constructors in https://github.com/zalando/logbook/pull/2005. Please have a look

kasmarian:
[...]

PR #2005, opened 2025-01-10
Allow using precise floats in logs

kasmarian:
## Description

This PR allows using precise floats in logs. 

Several JSON body filters in Logbook use `JsonGenerator` to reconstruct the payload.  
[generator.copyCurrentEvent](https://github.com/FasterXML/jackson-core/blob/0d2b0f39200d466f49f1abb06d9027053d41483d/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java#L2515) method has a flaw when writing float numbers. This may result in logging a different float value than the actual value that was in the payload.

With this change, I'm adding an optional boolean flag to all JSON body filters to use [copyCurrentEventExact](https://github.com/FasterXML/jackson-core/blob/0d2b0f39200d466f49f1abb06d9027053d41483d/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java#L2515) instaed. That may come with a performance impact (not sure how significant).

## Motivation and Context

https://github.com/zalando/logbook/issues/1993

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -534,6 +534,17 @@ a JSON response body will **not** be escaped and represented as a string:
 }
 ```
 
+
+> [!NOTE]  
+> Logbook is using [BodyFilters](#Filtering) to inline json payload or to find fields for obfuscation. 
+> Filters for JSON bodies are using Jackson, which comes with a defect of dropping off precision from floating point 
+> numbers (see [FasterXML/jackson-core/issues/984](https://github.com/FasterXML/jackson-core/issues/984)).
+> 
+> This can be changed by setting the `usePreciseFloats` flag to true in the filter respective filters.
+> 
+> E.g. `new CompactingJsonBodyFilter(true)` will keep the precision of floating point numbers.
+
+
 ##### Common Log Format
 
 The Common Log Format ([CLF](https://httpd.apache.org/docs/trunk/logs.html#common)) is a standardized text file format used by web servers when generating server log files. The format is supported via
--- a/logbook-json/src/main/java/org/zalando/logbook/json/ParsingJsonCompactor.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/ParsingJsonCompactor.java
@@ -11,12 +11,23 @@ final class ParsingJsonCompactor implements JsonCompactor {
 
     private final JsonFactory factory;
 
+    private final boolean usePreciseFloats;
+
+    public ParsingJsonCompactor(final JsonFactory factory, final boolean usePreciseFloats) {
+        this.factory = factory;
+        this.usePreciseFloats = usePreciseFloats;
+    }
+
+    public ParsingJsonCompactor(final boolean usePreciseFloats) {
+        this(new JsonFactory(), usePreciseFloats);
+    }
+
     public ParsingJsonCompactor() {
         this(new JsonFactory());
     }
 
     public ParsingJsonCompactor(final JsonFactory factory) {
-        this.factory = factory;
+        this(factory, false);
     }
 
     @Override
@@ -27,7 +38,11 @@ final class ParsingJsonCompactor implements JsonCompactor {
                 final JsonGenerator generator = factory.createGenerator(output)) {
 
             while (parser.nextToken() != null) {
-                generator.copyCurrentEvent(parser);
+                if (usePreciseFloats) {
+                    generator.copyCurrentEventExact(parser);
+                } else {
+                    generator.copyCurrentEvent(parser);
+                }
             }
 
             generator.flush();
--- a/logbook-json/src/test/java/org/zalando/logbook/json/CompactingJsonBodyFilterTest.java
+++ b/logbook-json/src/test/java/org/zalando/logbook/json/CompactingJsonBodyFilterTest.java
@@ -12,12 +12,14 @@ class CompactingJsonBodyFilterTest {
     /*language=JSON*/
     private final String pretty = "{\n" +
             "  \"root\": {\n" +
-            "    \"child\": \"text\"\n" +
+            "    \"child\": \"text\",\n" +
+            "    \"float_child\" : 0.40000000000000002" +
             "  }\n" +
             "}";
 
     /*language=JSON*/
-    private final String compacted = "{\"root\":{\"child\":\"text\"}}";
+    private final String compacted = "{\"root\":{\"child\":\"text\",\"float_child\":0.4}}";
+    private final String compactedWithPreciseFloat = "{\"root\":{\"child\":\"text\",\"float_child\":0.40000000000000002}}";
 
     @Test
     void shouldIgnoreEmptyBody() {
@@ -48,6 +50,13 @@ class CompactingJsonBodyFilterTest {
     void shouldTransformValidJsonRequestWithCompatibleContentType() {
         final String filtered = unit.filter("application/custom+json", pretty);
         assertThat(filtered).isEqualTo(compacted);
+    }
+
+    @Test
+    void shouldPreserveBigFloatOnCopy() {
+        final String filtered = new CompactingJsonBodyFilter(true)
+                .filter("application/custom+json", pretty);
+        assertThat(filtered).isEqualTo(compactedWithPreciseFloat);
     }
 
     @Test
--- a/logbook-json/src/main/java/org/zalando/logbook/json/CompactingJsonBodyFilter.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/CompactingJsonBodyFilter.java
@@ -21,6 +21,10 @@ public final class CompactingJsonBodyFilter implements BodyFilter {
 
     private final JsonCompactor compactor;
 
+    public CompactingJsonBodyFilter(final boolean usePreciseFloats) {
+        this(new ParsingJsonCompactor(usePreciseFloats));
+    }
+
     public CompactingJsonBodyFilter() {
         this(new ParsingJsonCompactor());
     }
--- a/logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/JacksonJsonFieldBodyFilter.java
@@ -29,11 +29,20 @@ public class JacksonJsonFieldBodyFilter implements BodyFilter {
     private final String replacement;
     private final Set<String> fields;
     private final JsonFactory factory;
+    private final boolean usePreciseFloats;
 
-    public JacksonJsonFieldBodyFilter(final Collection<String> fieldNames, final String replacement, final JsonFactory factory) {
+    public JacksonJsonFieldBodyFilter(final Collection<String> fieldNames,
+                                      final String replacement,
+                                      final JsonFactory factory,
+                                      final boolean usePreciseFloats) {
         this.fields = new HashSet<>(fieldNames); // thread safe for reading
         this.replacement = replacement;
         this.factory = factory;
+        this.usePreciseFloats = usePreciseFloats;
+    }
+
+    public JacksonJsonFieldBodyFilter(final Collection<String> fieldNames, final String replacement, final JsonFactory factory) {
+        this(fieldNames, replacement, factory, false);
     }
 
     public JacksonJsonFieldBodyFilter(final Collection<String> fieldNames, final String replacement) {
@@ -54,7 +63,11 @@ public class JacksonJsonFieldBodyFilter implements BodyFilter {
                 JsonToken nextToken;
                 while ((nextToken = parser.nextToken()) != null) {
 
-                    generator.copyCurrentEvent(parser);
+                    if (usePreciseFloats) {
+                        generator.copyCurrentEventExact(parser);
+                    } else {
+                        generator.copyCurrentEvent(parser);
+                    }
                     if (nextToken == JsonToken.FIELD_NAME && fields.contains(parser.currentName())) {
                         nextToken = parser.nextToken();
                         generator.writeString(replacement);
--- a/logbook-json/src/main/java/org/zalando/logbook/json/PrettyPrintingJsonBodyFilter.java
+++ b/logbook-json/src/main/java/org/zalando/logbook/json/PrettyPrintingJsonBodyFilter.java
@@ -20,9 +20,16 @@ import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 public final class PrettyPrintingJsonBodyFilter implements BodyFilter {
 
     private final JsonFactory factory;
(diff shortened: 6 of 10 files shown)


logbook-json/src/main/java/org/zalando/logbook/json/ParsingJsonCompactor.java
```diff
@@ -27,7 +38,11 @@ public String compact(final String json) throws IOException {
                 final JsonGenerator generator = factory.createGenerator(output)) {
 
             while (parser.nextToken() != null) {
-                generator.copyCurrentEvent(parser);
+                if (usePreciseFloats) {
```
msdousti, line 41: Can this logic be refactored?

msdousti (commented):
Meta comment: Jackson can encode floats as double, BigDecimal or String. The last option might be both efficient and precise.

logbook-json/src/main/java/org/zalando/logbook/json/ParsingJsonCompactor.java
kasmarian: you mean to avoid 2 if blocks?  Did it in https://github.com/zalando/logbook/pull/2005/commits/27515ed39aab1825789c59480936ed78a3de2f82

README.md
```diff
@@ -534,6 +534,17 @@ a JSON response body will **not** be escaped and represented as a string:
 }
 ```
 
+
+> [!NOTE]  
+> Logbook is using [BodyFilters](#Filtering) to inline json payload or to find fields for obfuscation. 
+> Filters for JSON bodies are using Jackson, which comes with a defect of dropping off precision from floating point 
+> numbers (see [FasterXML/jackson-core/issues/984](https://github.com/FasterXML/jackson-core/issues/984)).
+> 
+> This can be changed by setting the `usePreciseFloats` flag to true in the filter respective filters.
```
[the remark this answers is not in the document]
kasmarian: added in https://github.com/zalando/logbook/pull/2005/commits/65bcb76374bf8fdebf78a82880bd07c6bd34c2d2

logbook-json/src/main/java/org/zalando/logbook/json/ParsingJsonCompactor.java
msdousti: Thanks, but we now have three places where `private void copyCurrentEvent()` is defined. What I meant is, could we define it once, and use it in all places?

msdousti:
> This may also lead to a confusion as it won't represent the actual payload.

We can give the library user the option to choose between float representations: Double, BigDecimal, or String. This is IMHO better than the boolean flag. 

* If they want speed + precision, and don't care about number represented as string, they use String.
* speed but not precision ==> Double
* low speed, high precision ==> BigDecimal

WDYT?

logbook-json/src/main/java/org/zalando/logbook/json/ParsingJsonCompactor.java
kasmarian: These 3 places are not really connected to each other. In another language I could potentially do it via an extension function, but here the only way would be to extract it further into a static method that would take not 2 but 3 arguments. Not sure if I like the idea. Please suggest a better way, if you have it in mind

kasmarian:
> > This may also lead to a confusion as it won't represent the actual payload.
> 
> We can give the library user the option to choose between float representations: Double, BigDecimal, or String. This is IMHO better than the boolean flag.
> 
>     * If they want speed + precision, and don't care about number represented as string, they use String.
> 
>     * speed but not precision ==> Double
> 
>     * low speed, high precision ==> BigDecimal
> 
> 
> WDYT?

Can you give me a hint hot to actually use floats as strings, when `JsonGenerator` is used? All I'm finding so far is a way to annotate DTO classes to use this feature, but can't find a low level approach that would fit in the current implementation.

whiskeysierra:
I'd suggest to use the Strategy pattern internally, eg FloatRepresentation
with a single method to modify a Jsongenerator (?) with two
implementations, one for float and one for Big decimal. Pass that to as an
argument to the three places where it's needed.

msdousti24:
> Can you give me a hint hot to actually use floats as strings, when `JsonGenerator` is used? All I'm finding so far is a way to annotate DTO classes to use this feature, but can't find a low level approach that would fit in the current implementation.

Haven't tested it, but from [their wiki](https://github.com/FasterXML/jackson-core/wiki/JsonWriteFeatures):

```java
JsonFactory f = JsonFactory.builder()
   .enable(JsonWriteFeature.WRITE_NUMBERS_AS_STRINGS)
   .build();
```

msdousti:
> Can you give me a hint hot to actually use floats as strings, when `JsonGenerator` is used? All I'm finding so far is a way to annotate DTO classes to use this feature, but can't find a low level approach that would fit in the current implementation.

Haven't tested it, but from [their wiki](https://github.com/FasterXML/jackson-core/wiki/JsonWriteFeatures):

```java
JsonFactory f = JsonFactory.builder()
   .enable(JsonWriteFeature.WRITE_NUMBERS_AS_STRINGS)
   .build();
```

logbook-json/src/main/java/org/zalando/logbook/json/ParsingJsonCompactor.java
kasmarian: Wrapped everything in strategy as suggested by @whiskeysierra in https://github.com/zalando/logbook/pull/2005#issuecomment-2596240190. 

Not sure if I like the overcomplication (in my view), but I'm OK with keeping it this way, if all in favor. Please have a look.

logbook-json/src/test/java/org/zalando/logbook/json/CompactingJsonBodyFilterTest.java
```diff
@@ -50,6 +51,22 @@ void shouldTransformValidJsonRequestWithCompatibleContentType() {
...
+                .filter("application/custom+json", pretty);
+        final String compactedWithPreciseFloat = "{\"root\":{\"child\":\"text\",\"float_child\":0.40000000000000002}}";
+        assertThat(filtered).isEqualTo(compactedWithPreciseFloat);
+    }
+
+    @Test
+    void shouldLogFloatAsString() {
+        final String filtered = new CompactingJsonBodyFilter(new NumberAsStringJsonGeneratorWrapperCreator())
+                .filter("application/custom+json", pretty);
+        final String compactedWithFloatAsString = "{\"root\":{\"child\":\"text\",\"float_child\":\"0.40000000000000002\"}}";
```
kasmarian, line 66: @msdousti as per `WRITE_NUMBERS_AS_STRINGS` feature, it looks like it doesn't work the way we thought. 

JsonToken type is still ID_NUMBER_FLOAT and JsonGenerator calls [_copyCurrentFloatValue](https://github.com/FasterXML/jackson-core/blob/e4ac4f85873dc5e9cad8201f4b5b659db03b1f73/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java#L2772), which calls 
[WriterBasedJsonGenerator.writeNumber(double d)](https://github.com/FasterXML/jackson-core/blob/e4ac4f85873dc5e9cad8201f4b5b659db03b1f73/src/main/java/com/fasterxml/jackson/core/json/WriterBasedJsonGenerator.java#L816-L826) and the double value is written down to the write, just wrapped in a string. 

I left this failing test as well as `NumberAsStringJsonGeneratorWrapperCreator` implementation as a showcase.

logbook-json/src/main/java/org/zalando/logbook/json/ParsingJsonCompactor.java
whiskeysierra: I'm not fully in the topic but this duality of the creator and wrapper doesn't seem needed to me. What do we ultimately need to do? Call some enable method on the JsonFactory, JsonGenerator, ...?

kasmarian: The reason is mostly because JsonGenerator has to be created every time `filter` method is called, as it's created per output. At the same time, the Json BodyFilter classes are created once the Logbook is created. 
So to save the intent (precise float or not) we need to insert this information on the class creation time, but apply it (call different methods of the JsonGenerator depending on the strategy) on every `filter` call with different instances of generators.

whiskeysierra: I'd have the strategy invoked at the point where a factory (and output) is given and a generator is needed.
The default returns `factory.createGenerator(output)`.
The string variant returns:

```
factory.rebuild()
    .enable(JsonWriteFeature.WRITE_NUMBERS_AS_STRINGS)
    .build()
    .createGenerator(output)
    .enable(JsonGenerator.Feature.WRITE_NUMBERS_AS_STRINGS)
```

(Do we actually need both features there?)

The precise variant returns a custom `JsonGenerator` that wraps around an existing one and delegates everything, except copyCurrentEvent which it implements by calling copyCurrentEventExact on the underlying generator. You can create an abstract `ForwardingJsonGenerator` that just delegates every method as-is and have the PreciseFloatJsonGenerator only override the single method. Makes it a bit easier to read.

kasmarian: > (Do we actually need both features there?)

No, it's a transition from one way to set the feature to another, `JsonGenerator.Feature.WRITE_NUMBERS_AS_STRINGS` is deprecated. But in another comment (https://github.com/zalando/logbook/pull/2005#discussion_r1920213285), I'm trying to convince @msdousti to give up on the idea to use `WRITE_NUMBERS_AS_STRINGS` feature all together, as it doesn't preserve the precision. I'd just leave precise and not precise floats for now. 


> I'd have the strategy invoked at the point where a factory (and output) is given and a generator is needed.
The default returns factory.createGenerator(output).

But that's what is happening. Or am I missing something? I can keep the boolean flag in the BodyFilter classes constructors to decide which strategy to call, but then it's no better to the previous version.

logbook-json/src/test/java/org/zalando/logbook/json/CompactingJsonBodyFilterTest.java
msdousti: @kasmarian 

I should admit `enable(JsonWriteFeature.WRITE_NUMBERS_AS_STRINGS)` was a bad suggestion as it is used during **writing** of an object to JSON, not **reading** of a JSON string.

Unfortunately, I could not find a similar [JsonReadFeature](https://github.com/FasterXML/jackson-core/wiki/JsonReadFeatures).

Alternatively, I came up with the following code. The [compact function](https://github.com/zalando/logbook/blob/6cd012b53b8b25ae323ca79ef97bd1549713b11a/logbook-json/src/main/java/org/zalando/logbook/json/ParsingJsonCompactor.java#L23-L37) can be adapted to look like this or something similar.

```java
import com.fasterxml.jackson.core.JsonFactory;
import com.fasterxml.jackson.core.JsonGenerator;
import com.fasterxml.jackson.core.JsonParser;

import java.io.CharArrayWriter;

import static com.fasterxml.jackson.core.JsonToken.VALUE_NUMBER_FLOAT;

enum JsonNumberFormat {
    AS_DEFAULT,
    AS_FLOAT,
    AS_DOUBLE,
    AS_STRING,
    AS_BIG_DECIMAL,
}

public class Example {


    private static String readJson(String json, JsonNumberFormat format) throws Exception {
        final JsonFactory factory = new JsonFactory();

        try (
                final CharArrayWriter output = new CharArrayWriter(json.length());
                final JsonParser parser = factory.createParser(json);
                final JsonGenerator generator = factory.createGenerator(output)) {

            while (parser.nextToken() != null) {
                if (parser.getCurrentToken() == VALUE_NUMBER_FLOAT) {
                    switch (format) {
                        case AS_FLOAT -> generator.writeNumber(parser.getFloatValue());
                        case AS_DOUBLE -> generator.writeNumber(parser.getValueAsDouble());
                        case AS_STRING -> generator.writeString(parser.getValueAsString());
                        case AS_BIG_DECIMAL -> generator.writeNumber(parser.getDecimalValue());
                        default -> generator.copyCurrentEvent(parser);
                    }
                } else {
                    generator.copyCurrentEvent(parser);
                }
            }

            generator.flush();
            return output.toString();
        }
    }

    public static void main(String[] args) throws Exception {
        final String json = """
            {
                "x": [0.40000000000000002, 1, "a"]
            }
            """;

        System.out.println(readJson(json, JsonNumberFormat.AS_DEFAULT));
        System.out.println(readJson(json, JsonNumberFormat.AS_FLOAT));
        System.out.println(readJson(json, JsonNumberFormat.AS_DOUBLE));
        System.out.println(readJson(json, JsonNumberFormat.AS_STRING));
        System.out.println(readJson(json, JsonNumberFormat.AS_BIG_DECIMAL));
    }
}
```

Output:
```
{"x":[0.4,1,"a"]}
{"x":[0.4,1,"a"]}
{"x":[0.4,1,"a"]}
{"x":["0.40000000000000002",1,"a"]}
{"x":[0.40000000000000002,1,"a"]}
```

logbook-json/src/main/java/org/zalando/logbook/json/ParsingJsonCompactor.java
whiskeysierra: > But that's what is happening. Or am I missing something?

Almost, yes. You have this dedicated wrapper class around the generator, I'm assuming to reduce the interface size? I'd just get rid of that and the duality of having a wrapper and a creator for each case. You can get away with just a creator in two out of three cases and only for one of them you need a custom generator.

logbook-json/src/test/java/org/zalando/logbook/json/CompactingJsonBodyFilterTest.java
kasmarian: Good suggestion with using bare `(parser.getValueAsString())`! It's straightforward, and the functionality is on the same level, that these filters are operating on, as opposed to factory feature flags, and I didn't like as much. I applied it in `NumberAsStringJsonGeneratorWrapper`. Please have a look. 
If you think we should switch from the strategy approach to feature flags (enums), let's discuss it in the other thread, please.

ChristianLohmann (approved):

msdousti (approved):

logbook-json/src/test/java/org/zalando/logbook/json/CompactingJsonBodyFilterTest.java
msdousti: Thanks a lot. I like it better the way it's implemented now, which is more OOP that using enums :)

PR #2049, opened 2025-03-19
Add HttpAttributes To FilteredHttpRequest

### LINKED ISSUE #2024: Strategy 'BodyOnlyIfStatusAtLeastStrategy' does not log request attributes
## Description
In our project we need to log the attributes which extracted from JWT token. When using BodyOnlyIfStatusAtLeastStrategy, the request attributes does not be logged when the response status is less than the minimum status.

## Expected Behavior
withoutBody() should remove only the body from the request but not the attributes.

## Actual Behavior
log without body removed body and attributes

## Possible Fix

We found in this class, the request.withoutBodhy() removed the attributes from the request object.
https://github.com/zalando/logbook/blob/main/logbook-core/src/main/java/org/zalando/logbook/core/BodyOnlyIfStatusAtLeastStrategy.java#L34C27-L35C10

## Steps to Reproduce

1. log with strategy BodyOnlyIfStatusAtLeastStrategy and status=400
2. exract attributes with conf  

```
logbook:
  strategy: body-only-if-status-at-least
  minimum-status: 400
  attribute-extractors:
    - type: JwtAllMatchingClaimsExtractor
      claim-names: [ "xxx" ]

```

3. log the information

Thanks ^^

ChristianLohmann:
## Description 
When a `withoutBody` strategy is used then possible HttpAttributes won't be written to the sink. The problem is the difference in the HttpRequest object used to  send to the `writeBoth` method. I added the httpAttributes to the `FilteredHttpRequest` and adapted the `withRequest` method which will return a new `FilteredHttpRequest`  containing the original `HttpAttributes`.  Fixes #2024

## Motivation and Context
see #2024

## Types of changes

## Checklist:

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/core/FilteredHttpRequest.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/core/FilteredHttpRequest.java
@@ -1,5 +1,6 @@
 package org.zalando.logbook.core;
 
+import java.io.IOException;
 import lombok.AllArgsConstructor;
 import org.apiguardian.api.API;
 import org.zalando.logbook.BodyFilter;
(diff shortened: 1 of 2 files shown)


kasmarian (approved):

ChristianLohmann:
:+1:

PR #2056, opened 2025-03-25
Fix ktor server plugin http codes propagation

shilenkoalexander:
## Description
While using of the library we found that our call responses are logged with incorrect status codes. Most of them are logged with 200, even if they shouldn't.

## Motivation and Context
We found that every time when we use `call.respond(HttpStatusCode)` in our ktor server the problem appears. It works the same for the default 404 handlers [here](https://github.com/ktorio/ktor/blob/7fe92611582d6774c322c9ed538ddc5e8d2f8ac6/ktor-server/ktor-server-core/common/src/io/ktor/server/engine/BaseApplicationEngine.kt#L107)

Appeared that ktor-server stores status in the body field in case of using `call.respond(HttpStatusCode)` method and handles this case [here](https://github.com/ktorio/ktor/blob/7fe92611582d6774c322c9ed538ddc5e8d2f8ac6/ktor-server/ktor-server-core/common/src/io/ktor/server/http/content/DefaultTransform.kt#L27)

The PR propagates the body to the `ServerResponse` constructor and covers these cases, so we can see the real status in the logbook functionality.

To reproduce this bug you can use [this repo](https://github.com/shilenkoalexander/logbook-ktor-server-status-bug-reproduction) and run requests from [requests.http](https://github.com/shilenkoalexander/logbook-ktor-server-status-bug-reproduction/blob/master/requests.http) file.

## Types of changes

## Checklist:

### DIFF
--- a/logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/LogbookServer.kt
+++ b/logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/LogbookServer.kt
@@ -56,7 +56,7 @@ private suspend fun handleCallRespond(
     responseProcessingStageKey: AttributeKey<ResponseProcessingStage>
 ) {
     val responseProcessingStage = call.attributes[responseProcessingStageKey]
-    val response = ServerResponse(call.response)
+    val response = ServerResponse(call.response, body)
     val responseWritingStage = responseProcessingStage.process(response)
     if (response.shouldBuffer() && body is ByteArrayContent) {
         response.buffer(body.bytes())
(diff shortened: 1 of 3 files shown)


kasmarian:
Thank you for the fix, @shilenkoalexander! TIL about `HttpStatusCode` passed a body object by ktor. The PR looks good to me.
Could you please make sure that all the commits in this PR are signed?

shilenkoalexander:
@kasmarian Thanks for the review. 
I've fixed the commit to be signed

kasmarian:
:+1:

ChristianLohmann:
:+1:

Issue #2053, opened 2025-03-23
Missing logs in case of errors


oemergenc:
Hi,

I have a spring webflux application with a simple Controller like this, this is just for demonstration purposes:

```
@RestController
open class ErrorController {

    @GetMapping("/abort")
    fun simulateAbort(): Mono<String> {
        logger.info { "Received connection abort simulation request" }
        return Mono.error(RuntimeException("Connection aborted"))
    }
}
```
I have setup Logbook in this application. Unfortunately I am missing logs for this endpoint. I have read that Error dispatch is not supported. Not sure if this means that errors are not supported. 
However, I am wondering what would be the best approach to also get Logbook logs in these cases?

oemergenc:
ping 😃

kasmarian:
Hey @oemergenc thank you for the report, I've prepared a possible fix for it, please have a look if you're interested - https://github.com/zalando/logbook/pull/2064.

kasmarian:
It doesn't look like the netty fix will address the issue you have with Logbook the way I understood the proposal there. `ErrorLogHandler` will log error details when a client closed a connection and the request wasn't processed fully yet, so that there are on access logs. Logbook adds a separate handler to the pipeline at the end, where HTTP messages are fully decoded but not yet consumed, so that Logbook can have access to the request attributes and body.  But at the same time, the request may not reach logbook's handler, if it's aborted before. That's where the `ErrorLogHandler` will give at least some insights about what happened.

Issue #1926, opened 2024-10-01
Logbook not working with Spring Boot 3.3.x + Webflux


karolhor:
Logbook does not work with Spring Boot 3.3.4 with webflux/netty dependencies.

## Description
I'm trying to integrate logbook with Spring Boot 3.3.x + webflux but it seems not to log requests/responses. I read the documentation and tried to find some working examples on the web, but without success.

Before that I worked with Spring Boot 2.7.x and Logbook 3.7.2, which worked as a charm. 
I've prepared a minimal repo with two versions of Spring Boot for testing to reproduce this issue.

## Steps to Reproduce

1. Clone repo https://github.com/karolhor/webflux-logbook-example
2. Run application for Spring Boot 3.3.x: `mvn spring-boot:run`
3. Open page in your browser:  http://localhost:8080/api/example
4. Check the logs - there is no request/response data
<img width="1800" alt="Screenshot 2024-10-02 at 01 31 09" src="https://github.com/user-attachments/assets/ebcf42fd-4b72-4ffb-bf92-b67658e56192">

5. Run application for Spring Boot 2.6.x: `mvn -f pom-2.6.xml spring-boot:run`
6. Open page in your browser:  http://localhost:8080/api/example
7. Check the logs - correct behaviour
<img width="1800" alt="Screenshot 2024-10-02 at 01 49 22" src="https://github.com/user-attachments/assets/181e23e2-f1c0-47a7-9d23-70057ece6d3e">

## Your Environment

* Version used: `3.9.0`
* Link to your project: https://github.com/karolhor/webflux-logbook-example

[pom.xml](https://github.com/karolhor/webflux-logbook-example/blob/master/pom.xml)
[pom-2.6.xml](https://github.com/karolhor/webflux-logbook-example/blob/master/pom-2.6.xml)
```xml
...
<properties>
  ...
  <logbook.version>3.9.0</logbook.version>
</properties>

<dependencyManagement>
  <dependencies>
    ...
    <dependency>
      <groupId>org.zalando</groupId>
      <artifactId>logbook-bom</artifactId>
      <version>${logbook.version}</version>
      <type>pom</type>
      <scope>import</scope>
      </dependency>
    </dependencies>
</dependencyManagement>

<dependencies>
  <dependency>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-starter-actuator</artifactId>
  </dependency>
  <dependency>
    <groupId>org.springframework.boot</groupId>
[...]

Issue #1555, opened 2023-06-20
Autoconfigured NettyServerCustomizer for webflux does not log for 4XX/5XX response code


gedankennebel:
Hi, not sure if this is a bug in generall or just an issue with our setup.

We are using spring-cloud-starter-gateway 3.0.6 (webflux netty based spring application) and by adding 
org.zalando:logbook-spring-boot-webflux-autoconfigure:3.1.0 we only see request/response logging for a 2XX successfull request/response call. By hitting 403 or 404 no request/response logs are printed. 

However, using the LogbookWebFilter from the logbook webflux component all incoming request/responses are printed as expected regardless of the resulting response status code. 

This currently brings us to the solution to not use the webflux autoconfigure package and add something like this to our application by our own:
```
Bean
@Order(Ordered.HIGHEST_PRECEDENCE)
@ConditionalOnWebApplication(type = REACTIVE)
public WebFilter logbookServerFilter(Logbook logbook) {
    return new LogbookWebFilter(logbook);
}
```

A debug session using the debug log level showed us that the LogbookServerHandler got removed if the application ran into an error, we assume that this is the reason why the final "write" is never happening.
```
 
reactor.netty.ReactorNetty               : [1c159a01-13, L:/[0:0:0:0:0:0:0:1]:8080 - R:/[0:0:0:0:0:0:0:1]:54112] Added decoder [LogbookServerHandler] at the end of the user pipeline, full pipeline: [reactor.left.httpCodec, reactor.left.httpTrafficHandler, LogbookServerHandler, reactor.right.reactiveBridge, DefaultChannelPipeline$TailContext#0] 
reactor.netty.http.server.HttpServer     : [1c159a01-13, L:/[0:0:0:0:0:0:0:1]:8080 - R:/[0:0:0:0:0:0:0:1]:54112] Handler is being applied: org.springframework.http.server.reactive.ReactorHttpHandlerAdapter@72bc5a26 
r.n.http.server.HttpServerOperations     : [1c159a01-13, L:/[0:0:0:0:0:0:0:1]:8080 - R:/[0:0:0:0:0:0:0:1]:54112] Last HTTP response frame 
r.n.http.server.HttpServerOperations     : [1c159a01-13, L:/[0:0:0:0:0:0:0:1]:8080 - R:/[0:0:0:0:0:0:0:1]:54112] No sendHeaders() called before complete, sending zero-length header 
[...]

PR #2064, opened 2025-04-14
LogbookServerHandler - attempt to write logs when the handler is prematurely removed from the ReactorNetty pipeline

kasmarian:
## Description

`LogbookServerHandler`'s `channelRead()` method is expecting to be invoked at least once with message object that is instance of `io.netty.handler.codec.http.LastHttpContent`. 

When the handler is added via `reactor.netty.Connection.addHandlerLast()` method, [ReactorNetty registers the handler for removal on context termination](https://github.com/reactor/reactor-netty/blob/535e76a5adce220d59c58be25c90946fab392ad9/reactor-netty-core/src/main/java/reactor/netty/ReactorNetty.java#L383), and this termination is sometimes triggered before the last `LastHttpContent` message is passed to the read method.  Leaving the `LogbookServerHandler`'s Sequence object only with response written. 

This change proposes to add a trigger on `handlerRemoved()` method of `io.netty.channel.ChannelHandlerAdapter` to see if the response writing stage was already submitted to the `Sequence` and add the request writing stage and trigger the runnables for both stages.

Below is the initial thought process, which proved to be wrong. The issue turned out to be not in the order of the LogbookServerHandler (it's added in the same place eventually for the bare spring boot webflux app between `HttpTrafficHandler` and `ReactiveBridge`), but in the way it is added.

~This happens when the handler is added using `addHandlerLast()`, which positioned it too late in the pipeline. By that point, other handlers (like content aggregators) had already consumed the `LastHttpContent` messages. When LogbookServerHandler was positioned incorrectly, it only saw the write() events but missed the complete read() events.~

~This change proposes to position the `LogbookServerHandler` in the pipeline at the point when the message is readable and bufferable, but before handlers that may skip lead to skipping the pipeline calls for the last closing message.~

~The solution also offers the following order of checks as none of the netty handlers that the solution relies on are mandatory (see [NettyPipeline](https://projectreactor.io/docs/netty/1.0.38/api/reactor/netty/NettyPipeline.html)):~

~**First priority:** After HttpTrafficHandler (where HTTP messages are fully decoded but not yet consumed)
**Second priority:** After HttpCodec (when HttpTrafficHandler isn't available)
**Last resort:** At the end of the pipeline (as it works now)~

## Motivation and Context

Relates to #2053, #1555, #1926

## Types of changes

## Checklist:

### DIFF
--- a/logbook-netty/src/main/java/org/zalando/logbook/netty/LogbookServerHandler.java
+++ b/logbook-netty/src/main/java/org/zalando/logbook/netty/LogbookServerHandler.java
@@ -75,4 +75,13 @@ public final class LogbookServerHandler extends ChannelDuplexHandler {
         context.write(message, promise);
     }
 
+    @Override
+    public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
+        // In cases when the LastHttpContent message is not received before the context is terminated,
+        // attempt to add the request writing stage to the Sequence to trigger runEagerly().
+        if (sequence.hasSecondTask()) {
+            sequence.set(0, throwingRunnable(requestStage::write));
+        }
+        super.handlerRemoved(ctx);
+    }
 }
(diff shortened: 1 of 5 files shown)


ChristianLohmann:
:+1:

ChristianLohmann (approved):

kasmarian:
:+1:

Issue #1897, opened 2024-09-14
Logbook logging Request info but not response


deejonz:
Using the example https://github.com/msdousti/quarkus-logbook I can see "Incoming Request" but not the Response. 
When using logbook with Spring boot I can see both.

deejonz:
issue still there, can you please respond?

kasmarian:
@msdousti as you looked at quarkus-logbook before, maybe you could help here?

msdousti:
@kasmarian I know what's wrong.

In my pr
https://github.com/zalando/logbook/pull/1909

I originally came up with a solution, that we decided to change after you reviewed the PR.

The original solution worked, but the revised one didn't, and it somehow slipped through my manual tests (we don't have automated tests that replicates this).

If you agree, I can recreate a PR with the original solution.

kasmarian:
> If you agree, I can recreate a PR with the original solution.


Please do

msdousti:
@deejonz The issue is addressed in https://github.com/zalando/logbook/pull/2068.
Please upgrade to Logbook 3.12.0 **once released**.

@kasmarian If you have an ETA on 3.12.0, will be great to mention it. I'll close the issue as done.

PR #2068, opened 2025-04-16
Refactor response processing in LogbookServerFilter

msdousti:
Moved `response.expose()` to conditional logic to ensure it is only called when `hasEntity()` is true. This improves the clarity and correctness of the response handling process.

## Description

This fixes the issue as per my original fix in https://github.com/zalando/logbook/pull/1909, which was later changed during the review process.

* **Fix:** The fix is to call expose methods conditionally, only when the request/response entity exists.
* **Test:** I wasn't able to write an automated test to cover this, but the existing tests plus cover for all branches already.

# Manual test
* Use the project https://github.com/msdousti/quarkus-logbook
    * Compile with the version of Logbook in this PR (`3.12.0-SNAPSHOT`)
    * Run the project
* Send a GET and a HEAD request, and observe the console

## GET
This succeeds even without the current PR - doing test to prevent regression:
 
```
curl -i http://localhost:8080/hello

HTTP/1.1 200 OK
Content-Type: text/plain;charset=UTF-8
content-length: 14

Hello, World!
``` 

image

## HEAD
This requires the change of this PR. Without it, the request just hangs.
 
```
curl -iI http://localhost:8080/hello

HTTP/1.1 200 OK
connection: keep-alive
``` 

image

## Motivation and Context

https://github.com/zalando/logbook/issues/1897

## Types of changes

## Checklist:

### DIFF
--- a/logbook-jaxrs/src/main/java/org/zalando/logbook/jaxrs/LogbookServerFilter.java
+++ b/logbook-jaxrs/src/main/java/org/zalando/logbook/jaxrs/LogbookServerFilter.java
@@ -43,12 +43,12 @@ public final class LogbookServerFilter implements ContainerRequestFilter, Contai
 
         read(request::getProperty, "process-response", ResponseProcessingStage.class)
                 .ifPresent(context.hasEntity() ?
-                        throwingConsumer(stage ->
-                                request.setProperty("write-response", stage.process(response))) :
+                        throwingConsumer(stage -> {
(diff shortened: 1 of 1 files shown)


msdousti:
👍

kasmarian (approved):

msdousti24:
@kasmarian Thanks for the approval.

Seems like I lost my "merge" permission, could you also merge it please?
image

Issue #2079, opened 2025-04-24
NullPointerException in LogbookFilter with Async Servlet Requests


3750:
# NullPointerException in LogbookFilter with Async Servlet Requests

## Description
We're experiencing a `NullPointerException` in `LogbookFilter` when processing asynchronous servlet requests. The error occurs when the `AtomicBoolean` attribute is null during the `LogbookAsyncListener.onComplete` callback.

The issue is relatively rare (76 errors out of ~70,000,000 requests over 6 days) but consistently results in HTTP 500 errors when it happens. All failures occur on GET requests.

## Expected Behavior
The LogbookFilter should properly handle asynchronous servlet requests without throwing exceptions, even in edge cases like client disconnections (broken pipes).

## Actual Behavior
When a certain race condition occurs during async request processing, we get:
```
UT015021: Failure dispatching async event
java.lang.NullPointerException: Cannot invoke "java.util.concurrent.atomic.AtomicBoolean.getAndSet(boolean)" because "attribute" is null
```

This happens in `LogbookFilter.write()` on line 87.

## Possible Fix
Looking at the `LogbookFilter` class, it appears the issue might be related to how request attributes are managed during async processing. The attribute that holds the `AtomicBoolean` is somehow null when the async completion listener is triggered.

A possible fix could be to add a null check before attempting to use the `attribute` variable in `LogbookFilter.write()`.

## Steps to Reproduce
The issue seems to be related to race conditions during async request processing, possibly when clients disconnect during response streaming, making it difficult to reproduce consistently.

## Context
This bug affects our web service reliability, causing occasional 500 errors that are logged but don't impact most users. However, we'd like to eliminate these errors completely.

## Your Environment
* Spring Boot 3.3.4
* Undertow 2.3.17.Final
* Zalando Logbook 3.11.0

### Full Dependencies List
```
"io.undertow:undertow-core:2.3.17.Final"
"io.undertow:undertow-servlet:2.3.17.Final"
"org.springframework.boot:spring-boot:3.3.4"
"org.springframework.boot:spring-boot-actuator:3.3.4"
"org.springframework.boot:spring-boot-actuator-autoconfigure:3.3.4"
"org.springframework.boot:spring-boot-autoconfigure:3.3.4"
"org.springframework.boot:spring-boot-starter:3.3.4"
"org.springframework.boot:spring-boot-starter-aop:3.3.4"
"org.springframework.boot:spring-boot-starter-data-mongodb-reactive:3.3.4"
[...]

PR #2086, opened 2025-04-28
Check for synchronization attribute to not be null

kasmarian:
## Description

In rare cases when an async flow is enabled, LogbookFilter throws NPE because the synchronization attribute is missing in the request context. The root cause for such scenarios is not clear and happens rarely. The hypothesis is that it happens due to possible race conditions that happen when the calling client cancels the request and closes the connection. But it wasn't reproduced locally.

With this change the logging of the request & response will not happen in case of the flow described above.
## Motivation and Context
https://github.com/zalando/logbook/issues/2079

## Types of changes

## Checklist:


msdousti (approved):

kasmarian:
:+1:

PR #2087, opened 2025-04-28
Use default coroutine context in ktor client

kasmarian:
## Description
It appears that when nesting coroutine contexts via `scope.launch(coroutineContext) {}` LogbookClient may lead to deadlocks on big responses, as `loggingContent.discard()` uses blocking operations underneath and if other plugins interfered with the current coroutine scope.

When inheriting the scope it seems that the issue is resolved, as Dispatchers.IO is used and not [kotlinx.coroutines.BlockingEventLoop](https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/src/EventLoop.kt) as the coroutine element.

## Types of changes

## Checklist:

- [ ] All commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification)

### DIFF
--- a/logbook-ktor-client/src/main/kotlin/org/zalando/logbook/client/LogbookClient.kt
+++ b/logbook-ktor-client/src/main/kotlin/org/zalando/logbook/client/LogbookClient.kt
@@ -56,7 +56,7 @@ class LogbookClient(
 
             scope.receivePipeline.intercept(HttpReceivePipeline.After) { httpResponse ->
                 val (loggingContent, responseContent) = httpResponse.rawContent.split(httpResponse)
-                scope.launch(coroutineContext) {
+                scope.launch {
                     val responseProcessingStage = httpResponse.call.attributes[responseProcessingStageKey]
                     val clientResponse = ClientResponse(httpResponse)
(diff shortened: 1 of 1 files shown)


logbook-ktor-client/src/main/kotlin/org/zalando/logbook/client/LogbookClient.kt
```diff
@@ -56,7 +56,7 @@ class LogbookClient(
 
             scope.receivePipeline.intercept(HttpReceivePipeline.After) { httpResponse ->
                 val (loggingContent, responseContent) = httpResponse.rawContent.split(httpResponse)
-                scope.launch(coroutineContext) {
+                scope.launch {
```
kasmarian, line 59: I'm not sure how to test this with a unit test for now. If anyone has ideas, please suggest a way.

msdousti (approved):

kasmarian:
:+1:

pinguinjkeke:
Why do we run logging in a coroutine? What are we winning here? I think it's a pretty fast operation.

PR #2144, opened 2025-08-22
Switch from rest client driver to wiremcok

kasmarian:
## Description

In this PR I'm suggesting to switch from partially using rest-client-driver and partially using wiremock for tests, to only use wiremock.

This PR depends on https://github.com/zalando/logbook/pull/2143, as wiremock surfaced the same issue.

## Motivation and Context
- [rest-client-driver](https://github.com/rest-driver/rest-driver) is not actively supported.
- wiremock allows more flexibility

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification)

Issue #2103, opened 2025-06-03
Upgrading to 3.12.1 from 3.7.2 : problem with wiremock's MALFORMED_RESPONSE_CHUNK and okhttp


rpiterman:
## Description
For our http-intensive application we wrap okhttp with a domain-specific client using logbook for logging.
To test end-to-end, we use wiremock faults to simulate IO Problems and validate the application level result is as required.

While upgrading from logbook 3.7.2 to 3.12.1, the MALFORMED_RESPONSE_CHUNK test fails and seems to crush okhttp own thread.

## Expected Behavior
logbook okhttp interceptor does not affect, in no circumstances, the result of the request or response in the application.

## Actual Behavior
a MALFORMED_RESPONSE_CHUNK simulation of wiremock causes an exception to be thrown in okhttp managed thread.

rpiterman:
Using 3.12.1, this is the output (sorry for that, don't know why github does not preserve whitespace in code):

`13:07:08.520 [OkHttp http://localhost:52511/...] TRACE org.zalando.logbook.Logbook - Outgoing Request: bbc7130e69cb236f
Remote: localhost
GET http://localhost:52511/foo/no-response HTTP/1.1
accept: application/json
accept-encoding: gzip
authorization: XXX
connection: Keep-Alive
host: localhost:52511
ocpi-from-country-code: DE
ocpi-from-party-id: SNH
user-agent: okhttp/4.12.0
x-correlation-id: d6a620a7-8310-4a10-9ddd-3b8d8a505a74-c
x-request-id: d6a620a7-8310-4a10-9ddd-3b8d8a505a74-r
13:07:08.542 [OkHttp http://localhost:52511/...] WARN  o.z.l.okhttp.LogbookInterceptor - Unable to log response. Will skip the response logging step.

java.net.ProtocolException: Expected leading [0-9a-fA-F] character but was 0x6c
	at okhttp3.internal.http1.Http1ExchangeCodec$ChunkedSource.readChunkSize(Http1ExchangeCodec.kt:444)
	at okhttp3.internal.http1.Http1ExchangeCodec$ChunkedSource.read(Http1ExchangeCodec.kt:416)
	at okhttp3.internal.connection.Exchange$ResponseBodySource.read(Exchange.kt:281)
	at okio.RealBufferedSource.read(RealBufferedSource.kt:192)
	at okio.RealBufferedSource.request(RealBufferedSource.kt:209)
[...]

PR #2145, opened 2025-08-22
Issue 2103/malformed chunked response

kasmarian:
## Description
Logbook wraps the original `okhttp3.Response` in `org.zalando.logbook.okhttp.RemoteResponse` in it's LogbookInterceptor. When a response body is requested, `org.zalando.logbook.okhttp.RemoteResponse` attempts to read the original response buffer stream and buffer it during a state transition from `Offering` to `Buffering`. When the response is chunked and the body is malformed, okhttp throws `ProtocolException` and closes the underlying stream. In the result, not only the response is not logged by Logbook, but the caller doesn't get access to the original response (with faulty body).

To address this I attempt to:
-  okio.BufferedSource.peek() method when buffering the response body, and when it fails to fallback to a string `"!This is not a real response body. Logbook was unable to read the response body!"`, that will appear in the Logbook's log as the response body. This is done to not omit logging the response completely and to avoid misleading the client that the body is empty.
- when LogbookInterceptor returns the response back from the intercept call, it calls `toResponse()` method, that will also trigger the buffering flow that would return the `"!This is not a real response body. Logbook was unable to read the response body!"` as body. To avoid this, as because the `peek` method was used on the BufferedSource, the BufferedSource wasn't closed, so returning the original `okhttp3.Response` should be safe in this case.

Depends on https://github.com/zalando/logbook/pull/2144

## Motivation and Context
#2103

## Types of changes

## Checklist:

### DIFF
--- a/logbook-common/pom.xml
+++ b/logbook-common/pom.xml
@@ -20,10 +20,6 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-nop</artifactId>
         </dependency>
-        <dependency>
-            <groupId>com.github.rest-driver</groupId>
-            <artifactId>rest-client-driver</artifactId>
-        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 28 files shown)


lukasniemeier-zalando:
:+1:

kasmarian:
:+1:

Issue #1865, opened 2024-06-26
Unable to read request body from ServletRequest twice because InputStream is getting exhausted after first use


mutyasaisrikar:
I am using Java 17, Spring Boot 3 and logbook-spring-boot-starter:3.9.0

I have a Spring interceptor where I have to read the request body in my application. I also use Logbook to log the request and the response in custom format.

I was previously using the `1.13.0` version of Logbook. After migrating to the latest version, I can read the request body in my Spring interceptor but it is failing at `@RequestBody` annotation in the Spring Controller class with an error saying `Required request body is missing`

## Possible Fix
After going through the code, I can see that previously with logbook 1.x, the underlying logic of `httpServletRequest.getInputStream` method was creating a `ByteArrayInputStream` object every time it gets called. So If I were to call getInputStream 2 times, it returns 2 different `ByteArrayInputStream` objects.

But now It is getting created once per request. Because of that, the first time getInputStream is called, the stream is closed and the second time we get an empty response body.

## Steps to Reproduce
https://github.com/mutyasaisrikar/logbook-error
1. SpringConfig.java -> TestInterceptor class - reads request body and prints
2. `RestApiTest` test fails because the Controller class is not able to read the request body

mutyasaisrikar:
Hello @whiskeysierra,

This links to a similar issue I have raised before.

https://github.com/zalando/logbook/issues/974

 At that time I was not able to understand. But now I have more information about this. Can you please help with this?

artpaym:
@mutyasaisrikar Afaik, it's a default behavior that reading from request stream drains making controllers fail, and it requires [manual handling of that](https://www.baeldung.com/spring-reading-httpservletrequest-multiple-times) unfortunately. It seems that if Logbook in previous versions did it silently, it was only a fortunate coincidence.

kasmarian:
I missed that part when GH actions closed the issue, sorry for that. But the good news is that the fix (https://github.com/zalando/logbook/pull/2195) will be part of 4.0.0 release.

PR #2150, opened 2025-09-01
Return buffered InputStream from RemoteRequest.Buffering

kasmarian:
## Description
In Logbook's Servlet implementation,  `RemoteRequest` wrapper returns the same object of ServletInputStream when getInputStream() is called. When/If an interceptor reads the request input stream before the request reaches the handler, the stream will be closed. This is reproducible in https://github.com/mutyasaisrikar/logbook-error. 

The proposed change is to override the implementation of `getInputStream` in `RemoteRequest.Buffering` class, so that it returns a new instance of InputStream with the buffered body array on every call.

I'm also removing the `Streaming` abstract class that had 2 implementations: Ignoring and Buffering. As now Buffering will have its own implementations, I couldn't find a use for `Ignoring`, as I assume the InputStream won't be used by Logbook in case of `Ignoring`, and it can leave the implementation of `State` that is returning `request.getInputStream()`.

## Motivation and Context
#1865

## Types of changes

## Checklist:

- [ ] All commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification)

### DIFF
--- a/logbook-servlet/src/main/java/org/zalando/logbook/servlet/RemoteRequest.java
+++ b/logbook-servlet/src/main/java/org/zalando/logbook/servlet/RemoteRequest.java
@@ -8,7 +8,6 @@ import jakarta.servlet.ServletResponse;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletRequestWrapper;
 import lombok.AllArgsConstructor;
-import lombok.Getter;
 import lombok.SneakyThrows;
 import org.zalando.logbook.ContentType;
 import org.zalando.logbook.HttpHeaders;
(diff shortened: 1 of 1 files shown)


logbook-servlet/src/main/java/org/zalando/logbook/servlet/RemoteRequest.java
```diff
@@ -166,20 +146,24 @@ public byte[] getBody() {
             return body;
         }
 
+        @Override
+        public ServletInputStream getInputStream(final ServletRequest request) {
+            return new ServletInputStreamAdapter(new ByteArrayInputStream(body));
```
msdousti, line 151: Using `new ByteArrayInputStream(body)` can be a slow operation, especially for a large `body`. Is resolving this bug (which seems a rather edge case) worth doing it always?

An alternative might be giving the user an option to deep-clone the body, or simply return it.

ChristianLohmann:
:+1:

lukasniemeier-zalando:
@kasmarian small conflict after merging https://github.com/zalando/logbook/pull/2195

kasmarian:
@lukasniemeier-zalando actually, it looks like https://github.com/zalando/logbook/pull/2195 addressed this issue as well. I verified it with https://github.com/mutyasaisrikar/logbook-error. I'll close this PR

PR #2169, opened 2025-09-10
Keep Http Attributes in a FilteredResponse

ChristianLohmann:
This PR shall fix the issue that extracted HTTP Attributes are not logged in a filtered Response

## Types of changes

## Checklist:

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/core/FilteredHttpResponse.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/core/FilteredHttpResponse.java
@@ -1,24 +1,32 @@
 package org.zalando.logbook.core;
 
-import lombok.AllArgsConstructor;
+import java.io.IOException;
 import org.apiguardian.api.API;
 import org.zalando.logbook.BodyFilter;
 import org.zalando.logbook.ForwardingHttpResponse;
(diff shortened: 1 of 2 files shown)


logbook-core/src/main/java/org/zalando/logbook/core/FilteredHttpResponse.java
```diff
@@ -1,24 +1,32 @@
...
     private final BodyFilter bodyFilter;
+    private final HttpAttributes attributes;
+
+    FilteredHttpResponse(final HttpResponse response,
+                         final HeaderFilter headerFilter,
+                         final BodyFilter bodyFilter) {
+        this.response = response;
+        this.headerFilter = headerFilter;
+        this.bodyFilter = bodyFilter;
+        this.attributes = response.getAttributes();
```
kasmarian, line 28: Could you add a bit more context about how/why does it not work right now? [ForwardingHttpResponse would also just get attributes from the original response](https://github.com/zalando/logbook/blob/f434a7e22993cf18f119dc87100051337f2e3e4b/logbook-api/src/main/java/org/zalando/logbook/ForwardingHttpResponse.java#L38), how does extracting the attributes earlier help?

PR #2192, opened 2025-12-18
Uncompress compressed body before return

antsyganov:
## Description
The byte[] getBody() method decompresses the body before returning it. It checks if the response contains a Content-Encoding header with the value gzip or x-gzip. If so, the method decompresses the body before returning it. As a result, Logbook will print the decompressed content, instead of displaying unreadable compressed data such as �0��=W��Q.... This ensures that the logged output is human-readable.

## Motivation and Context
If the response body is compressed, calling new String(getBody(), getCharset()) will cause Logbook to treat the body as plain text (or JSON), rather than as compressed bytes. As a result, Logbook will print unreadable compressed data such as �0��=W��Q.... 

## Types of changes

## Checklist:


antsyganov:
👍

PR #2193, opened 2025-12-18
Migrate away from javax

kasmarian:
## Description

This PR completes the Jakarta EE migration by removing leftover javax annotations and adding JSpecify annotations for improved null-safety support. It also updates documentation and build scripts to reflect the changes from the base release.

  Changes

  1. Remove Deleted Jakarta Annotations

  With the migration to Jakarta EE, these javax annotations were removed and need replacement:
  - @javax.annotation.CheckReturnValue - for methods whose return values shouldn't be ignored
  - @javax.annotation.concurrent.Immutable - for immutable classes
  - @javax.annotation.concurrent.NotThreadSafe - for non-thread-safe classes
  - @javax.annotation.ParametersAreNonnullByDefault - for package-level null defaults

  2. Add JSpecify for Null-Safety

  Replaced deleted annotations with JSpecify 1.0.0 as a compile-time dependency:
  - Added @NullMarked package-level annotations to 18 package-info.java files
  - Provides package-level null-non-null semantics for static analysis tools
  - JSpecify included as optional dependency

  3. Fix Kotlin Null-Safety Issues

  Updated 4 Kotlin files to handle nullable content types:
  - logbook-ktor-client/src/main/kotlin/org/zalando/logbook/client/ClientRequest.kt
  - logbook-ktor-client/src/main/kotlin/org/zalando/logbook/client/ClientResponse.kt
  - logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/ServerRequest.kt
  - logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/ServerResponse.kt

  Added @file:Suppress("RETURN_TYPE_MISMATCH_ON_OVERRIDE") to handle Kotlin's strict null-checking when returning nullable String? from interface methods declaring non-nullable String.

  4. Build.sh

  - build.sh: Removed legacy package command as it was used because of shade plugin, which is dropped here as well.

 
## Types of changes

## Checklist:

### DIFF
--- a/logbook-ktor-client/src/main/kotlin/org/zalando/logbook/client/ClientRequest.kt
+++ b/logbook-ktor-client/src/main/kotlin/org/zalando/logbook/client/ClientRequest.kt
@@ -1,6 +1,5 @@
 @file:Suppress(
     "SimpleRedundantLet", // jacoco workaround
-    "RETURN_TYPE_MISMATCH_ON_OVERRIDE", // nullable content type in Kotlin
 )
 
 package org.zalando.logbook.client
--- a/logbook-ktor-client/src/main/kotlin/org/zalando/logbook/client/ClientResponse.kt
+++ b/logbook-ktor-client/src/main/kotlin/org/zalando/logbook/client/ClientResponse.kt
@@ -1,6 +1,5 @@
 @file:Suppress(
     "SimpleRedundantLet", // jacoco workaround
-    "RETURN_TYPE_MISMATCH_ON_OVERRIDE", // nullable content type in Kotlin
 )
 
 package org.zalando.logbook.client
--- a/logbook-ktor-common/pom.xml
+++ b/logbook-ktor-common/pom.xml
@@ -58,6 +58,11 @@
             <plugin>
                 <groupId>org.jetbrains.kotlin</groupId>
                 <artifactId>kotlin-maven-plugin</artifactId>
+                <configuration>
+                    <args>
+                        <arg>-Xjsr305=strict</arg>
+                    </args>
+                </configuration>
             </plugin>
         </plugins>
     </build>
--- a/logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/ServerRequest.kt
+++ b/logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/ServerRequest.kt
@@ -1,6 +1,5 @@
 @file:Suppress(
     "SimpleRedundantLet", // jacoco workaround
-    "RETURN_TYPE_MISMATCH_ON_OVERRIDE", // nullable content type in Kotlin
 )
 
 package org.zalando.logbook.server
--- a/logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/ServerResponse.kt
+++ b/logbook-ktor-server/src/main/kotlin/org/zalando/logbook/server/ServerResponse.kt
@@ -1,6 +1,5 @@
 @file:Suppress(
     "SimpleRedundantLet", // jacoco workaround
-    "RETURN_TYPE_MISMATCH_ON_OVERRIDE", // nullable content type in Kotlin
 )
 
 package org.zalando.logbook.server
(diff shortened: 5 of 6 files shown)


logbook-api/src/main/java/org/zalando/logbook/HttpHeaders.java
```diff
@@ -1,7 +1,7 @@
 package org.zalando.logbook;
 
-import javax.annotation.CheckReturnValue;
```
msdousti: `CheckReturnValue` can be imported by using `com.github.spotbugs:spotbugs-annotations`; see:

https://github.com/findbugsproject/findbugs/blob/master/findbugs/src/java/edu/umd/cs/findbugs/annotations/CheckReturnValue.java

logbook-core/src/main/java/org/zalando/logbook/core/package-info.java
```diff
@@ -1,4 +1,2 @@
-@ParametersAreNonnullByDefault
 package org.zalando.logbook.core;
 
-import javax.annotation.ParametersAreNonnullByDefault;
```
msdousti: I vaguely remember this added some compile-time null-safety guarantees.by adding nonnull annotation to all parameters, unless specified otherwise. (i.e., via `@Nullable`).

I couldn't readily find a replacement; but would be great to check more.

logbook-core/src/test/java/org/zalando/logbook/core/StreamHttpLogWriterTest.java
```diff
@@ -1,6 +1,5 @@
 package org.zalando.logbook.core;
 
-import net.jcip.annotations.NotThreadSafe;
```
msdousti: I think this should not be removed; rather, any place we have

```java
javax.annotation.concurrent.NotThreadSafe;
```

it should be replaced by 
```java
import net.jcip.annotations.NotThreadSafe;
```

`@NotThreadSafe` is merely a documentation annotation for our users.

logbook-api/src/main/java/org/zalando/logbook/HttpHeaders.java
msdousti: After some research, I was also going to suggest JSpecify. 

I think the proper annotation would be `@NullMarked`.

I guess optional is fine for our package compilation, but do our clients benefit from it if they forget to add it?

logbook-core/src/test/java/org/zalando/logbook/core/StreamHttpLogWriterTest.java
msdousti: > no updates since 2013

It's a simple set of documentation annotations based on the famous JCIP book. IMO, no update was ever needed as concepts like ThreadSafe, ThreadUnsafe and Immutable have not changed since 😅

kasmarian: switched to jcip dependency in https://github.com/zalando/logbook/pull/2193/commits/744bbcbee511eae93e370709def5d1488dc29c62

logbook-ktor-client/src/main/kotlin/org/zalando/logbook/client/ClientRequest.kt
```diff
@@ -1,5 +1,6 @@
 @file:Suppress(
     "SimpleRedundantLet", // jacoco workaround
+    "RETURN_TYPE_MISMATCH_ON_OVERRIDE", // nullable content type in Kotlin
```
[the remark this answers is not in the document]
kasmarian: the issue was that kotlin is not yet recognizing jakarta.annotations, and despite we're using Nullable annotation on this method, the compiler complained. Seems like I found a way to make it work in https://github.com/zalando/logbook/pull/2193/commits/71bdd2b2d7ed753d13d585e007d8ebce54167d17

logbook-api/src/main/java/org/zalando/logbook/HttpHeaders.java
msdousti: To my knowledge, `@NullMarked` is a per-package annotation; it automatically applies to all parameters and return values in that package.

My concern,as stated in previous comment, was about making JSpecify optional.

kasmarian: The dilemma I have with making it not optional is: should we make it compile or provided? Compile would bring the transitive dependency to deal with to the clients, but provided would force them to add one, which is somehow feels even worse.

msdousti:
👍

kasmarian:
👍

PR #2234, opened 2026-02-09
Added LogbookCustomizer for Spring

paul-kraftlauget:
## Description
Added LogbookCustomizer functional interface and implemented the standard usage pattern

## Motivation and Context
This gives the opportunity to have custom code to change the behavior of the builder. This is a common pattern already seen across many Spring components, giving the consuming application a lot of flexibility. 

Furthermore, the LogbookCreator.Builder is now a prototype bean, applying the standard customizers. This means that if the application wants two different LogBook instances, for example one instance for WebFlux and a slightly different instance used with WebClient, this is very easy to acheive.

## Types of changes

I am a bit unsure if this is considered a breaking change. It changes methods marked as internal APIs, so I think it should be ok. 

## Checklist:

### DIFF
--- a/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
+++ b/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
@@ -7,6 +7,7 @@ import jakarta.servlet.Servlet;
 import lombok.Generated;
 import org.apache.http.client.HttpClient;
 import org.apiguardian.api.API;
+import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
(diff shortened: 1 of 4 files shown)


logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -22,21 +23,11 @@
...
-import org.zalando.logbook.HttpLogWriter;
-import org.zalando.logbook.HttpRequest;
-import org.zalando.logbook.Logbook;
-import org.zalando.logbook.PathFilter;
-import org.zalando.logbook.QueryFilter;
-import org.zalando.logbook.RequestFilter;
-import org.zalando.logbook.ResponseFilter;
-import org.zalando.logbook.Sink;
-import org.zalando.logbook.Strategy;
+import org.zalando.logbook.*;
```
kasmarian: we tend to not use start imports, could you please change it?

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -111,7 +101,24 @@ public LogbookAutoConfiguration(final LogbookProperties properties) {
...
-    public Logbook logbook(
+    public Logbook logbook(LogbookCreator.Builder builder) {
+        return builder.build();
+    }
+
+    @API(status = INTERNAL)
+    @Bean
+    @ConditionalOnMissingBean(LogbookCreator.Builder.class)
+    @Scope("prototype")
+    public LogbookCreator.Builder logbookBuilder(final ObjectProvider<LogbookCustomizer> customizers) {
```
kasmarian: I wonder if we even need this prototype bean. Wouldn't it work the same if logbook bean would accept customizers and apply them? 

e.g. 
```
    @API(status = INTERNAL)
    @Bean
    @ConditionalOnMissingBean(Logbook.class)
    public Logbook logbook(List<LogbookCustomizer> customizers) {
        var builder = Logbook.builder();
        customizers.forEach(customizer -> customizer.customize(builder));
        return builder.build();
    }
```

kasmarian (commented):
I like the idea, thank you for you contribution! 

Could you please also add a test (or a few) in the autoconfiguration module where the customizer will be used? To avoid regression in this field. Could be a simple customization that's verified in the test, that checks that LogbookCustomizer beans are actually applied as expected.

kasmarian:
@paul-kraftlauget are you still interested in this PR?

paul-kraftlauget:
Yes, I am still interested.

I have been vacationing recently.

I hope to find time soon to complete the PR.

kasmarian:
@paul-kraftlauget are you still interested in this PR?

PR #2291, opened 2026-04-16
Fix logging of large response bodies in httpclient5 implementation LogbookHttpAsyncResponseConsumer

sfilipiak-inpost:
## Description
Current implementation of httpclient5 `LogbookHttpAsyncResponseConsumer` is incorrectly handling large response bodies (over 8192 bytes). This fix addresses this issue.

## Motivation and Context
httpclient5 changed its model to reactive/nio comparing to httpclient. In the new approach, `AsyncResponseConsumer` receives raw http chunks in `consume(ByteBuffer)` contrary to the old `responseCompleted(HttpContext)` which received full response.

This caused logbook to call `Sink.write()` on each http chunk it received. This worked fine for small response bodies but is causing errors for large responses.

This fix addresses this issue by using `FutureCallback<T> resultCallback` which allows logbook to log entire response once at the end of response processing. Unfortunately, since the delegate is a generic interface, additional `Function<T, byte[]> bodyExtractor` is required which causes the new constructor of `LogbookHttpAsyncResponseConsumer` to be incompatible with the old version (additional parameter is required).

I could try to think of some kind of workaround, but I believe it would make the implementation difficult to understand the maintain. I consider the current implementation incomplete/incorrect, therefore it's ok in my opinion to introduce this breaking change.

## Types of changes

## Checklist:

### DIFF
--- a/logbook-httpclient5/src/main/java/org/zalando/logbook/httpclient5/LogbookHttpAsyncResponseConsumer.java
+++ b/logbook-httpclient5/src/main/java/org/zalando/logbook/httpclient5/LogbookHttpAsyncResponseConsumer.java
@@ -1,5 +1,7 @@
 package org.zalando.logbook.httpclient5;
 
+import lombok.RequiredArgsConstructor;
+import lombok.experimental.Delegate;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.hc.core5.concurrent.FutureCallback;
 import org.apache.hc.core5.http.EntityDetails;
@@ -12,6 +14,7 @@ import org.zalando.logbook.Logbook.ResponseProcessingStage;
 
 import java.io.IOException;
 import java.nio.ByteBuffer;
+import java.util.function.Function;
 
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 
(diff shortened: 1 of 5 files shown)


logbook-httpclient5/src/main/java/org/zalando/logbook/httpclient5/LogbookHttpAsyncResponseConsumer.java
```diff
@@ -20,13 +23,15 @@
...
 
     private final AsyncResponseConsumer<T> consumer;
+    private final Function<T, byte[]> bodyExtractor;
     private final boolean decompressResponse;
     private HttpResponse response;
     private EntityDetails entityDetails;
     private ResponseProcessingStage stage;
 
-    public LogbookHttpAsyncResponseConsumer(final AsyncResponseConsumer<T> consumer, boolean decompressResponse) {
+    public LogbookHttpAsyncResponseConsumer(final AsyncResponseConsumer<T> consumer, final Function<T, byte[]> bodyExtractor, final boolean decompressResponse) {
```
lukasniemeier-zalando, line 32: What `bodyExtractor` may users specify here? I saw the test is using `SimpleHttpResponse::getBodyBytes` - is that a good choice in all cases?

lukasniemeier-zalando, line 32: What `bodyExtractor` may users specify here? I saw the test is using `SimpleHttpResponse::getBodyBytes` - is that a good choice in all cases?

sfilipiak-inpost: It solely depends on the type of `consumer` provided by the user. In tests we are using `SimpleResponseConsumer` with its `<T>` type as `SimpleHttpResponse`, hence `bodyExtractor` needs to extract response body from `SimpleHttpResponse`.

Other possible usage is (using existing implementations) e.g.
```java
import org.apache.hc.core5.http.Message;
import org.apache.hc.core5.http.nio.entity.BasicAsyncEntityConsumer;
import org.apache.hc.core5.http.nio.support.BasicResponseConsumer;

new LogbookHttpAsyncResponseConsumer<>(
    new BasicResponseConsumer<>(new BasicAsyncEntityConsumer()),
    Message::getBody,
    true
);
```

lukasniemeier-zalando: Thanks, makes sense :+1:

lukasniemeier-zalando (approved):

sfilipiak-inpost:
Is there a chance of getting another approval soon and having this change released in the next version?

ChristianLohmann:
:+1:

PR #2301, opened 2026-04-22
Add StatusCodeBasedSink for HTTP status-code-aware log levels

### LINKED ISSUE #2232: Use different log level based on the HTTP status
By default, `org.zalando.logbook.Logbook`  writes request and response entries at `TRACE` level only. It would be nice that based on the HTTP response code, the corresponding HTTP response entry could be logged on a `WARN`/`ERROR` level.

## Detailed Description
Use different log levels, to clearly separate successful HTTP responses from error ones. At the moment this requires creating multiple custom `HttpLogWriter` for every log level and a custom `Sink` to trigger the corresponding writer based on the HTTP response code. Configuring this for multiple microservices quickly becomes a maintenance job, because you either have to copy paste the same configuration everywhere, or try and maintain some custom Spring Boot Starter that exposes the `Sink` `@Bean`

## Context
Logging 4xx responses as `WARN` and 5xx as `ERROR`  (or consolidating both of them under `ERROR`) improves observability and alerting while keeping normal traffic at `TRACE`.

## Possible Implementation
Maybe a `Sink` which resolves the HTTP response code and delegates to corresponding `HttpLogWriter`?

Vinu2111:
## Description
Adds `StatusCodeBasedSink`, a new `Sink` implementation that 
selects the SLF4J log level based on the HTTP response status code:

- `2xx / 3xx` → `TRACE`
- `4xx` → `WARN`
- `5xx` → `ERROR`

Requests are always logged at `TRACE` level.

The sink is enabled via:
```properties
logbook.write.status-code-based=true
```

## Motivation and Context
Fixes #2232

Currently all HTTP traffic is logged at `TRACE` level regardless 
of the response status. This makes it difficult to observe errors 
in production without enabling TRACE logging for all traffic.

This change allows teams to log successful requests quietly at 
TRACE while surfacing 4xx and 5xx responses at WARN/ERROR — 
improving observability without any additional configuration burden.

## Types of changes

## Checklist
- [x] All commits are signed


logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -320,6 +321,20 @@ public Sink chunkingSink(final Sink sink) {
         return new ChunkingSink(sink, properties.getWrite().getChunkSize());
     }
 
+    @Bean
+    @Primary
```
kasmarian: Are you adding the `@Primary` annotation here because otherwise the DefaultSink bean is initialized first? I'm thinking if there's a way to avoid adding more @Primary beans. 
One way would be to wrap both beans into internal configuration classes and use @AutoConfigureBefore annotation to set the order. E.g. 

```
 @Configuration
    public static class DefaultLogbookSinkConfiguration {
        @Bean
        @ConditionalOnMissingBean(Sink.class)
        public Sink sink(
                final HttpLogFormatter formatter,
                final HttpLogWriter writer) {
            return new DefaultSink(formatter, writer);
        }
    }

    @Configuration
    @AutoConfigureBefore(DefaultLogbookSinkConfiguration.class)
    public static class StatusCodeBasedLogbookSinkConfiguration {
        @Bean
        @ConditionalOnMissingBean(Sink.class)
        public Sink sink(
                final HttpLogFormatter formatter) {
            return new StatusCodeBasedSink(formatter);
        }
    }
```

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookProperties.java
```diff
@@ -50,6 +50,7 @@ public static class Obfuscate {
     public static class Write {
         private int chunkSize;
         private int maxBodySize = -1;
+        private boolean statusCodeBased = false;
```
kasmarian, line 53: Please document the new property in `README.md` as well as in `spring-configuration-metadata.json`, so that clients are aware of the toggle and when to use it.

logbook-core/src/main/java/org/zalando/logbook/core/StatusCodeBasedSink.java
```diff
@@ -0,0 +1,52 @@
...
+@RequiredArgsConstructor
+public final class StatusCodeBasedSink implements Sink {
+
+    private static final Logger log = LoggerFactory.getLogger(Logbook.class);
+
+    private final HttpLogFormatter formatter;
+
+    @Override
+    public boolean isActive() {
+        return log.isEnabledForLevel(Level.TRACE);
```
kasmarian: A bit of overarching concern: you're here skipping the level of `HttpLogWriter` abstraction completely and partially replicate it's responsibility. But it has a few downsides:
a) here, for example, you check that trace log level is enabled, but warn and error log levels are also used in this implementation. It could be the case that clients would only want to have warn/error log entries to be recorded.
b) clients may use their own writer implementations, but they won't have a way to inject it here.


I'm thinking if it's sensible to introduce 3 writer objects - one per each log level, and use them here conditionally instead.

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -320,6 +321,20 @@ public Sink chunkingSink(final Sink sink) {
...
+    @ConditionalOnBean(Sink.class)
+    @ConditionalOnMissingBean(name = "statusCodeBasedSink")
+    @ConditionalOnProperty(
+            prefix = "logbook.write",
+            name = "status-code-based",
+            havingValue = "true"
+    )
+    public StatusCodeBasedSink statusCodeBasedSink(
```
kasmarian: Please create a test for the bean in this module also that will verify that when the conditional property is set to `true` it's being used.

Vinu2111:
Thank you for the detailed review @kasmarian! I'll address all four points:

1. Refactor `StatusCodeBasedSink` to use 3 `HttpLogWriter` instances (one per level) and delegate conditionally
2. Replace `@Primary` with nested `@Configuration` classes using `@AutoConfigureBefore` ordering as suggested
3. Add README documentation and `spring-configuration-metadata.json` entry
4. Add an autoconfiguration integration test

Will push the updated version shortly.

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
Vinu2111: Fixed - replaced @Primary with a conditional if/else branch in the sink() method based on the property value.

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookProperties.java
Vinu2111: Done - added documentation to README.md and spring-configuration-metadata.json.

logbook-core/src/main/java/org/zalando/logbook/core/StatusCodeBasedSink.java
Vinu2111: Addressed - introduced LevelBasedHttpLogWriter which accepts an SLF4J Level and logs at that level. StatusCodeBasedSink now holds 3 writer instances (traceWriter, warnWriter, errorWriter) and delegates to the appropriate one based on response status.

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
Vinu2111: Added StatusCodeBasedSinkTest in the autoconfigure module verifying the bean is activated when logbook.write.status-code-based=true is set.

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookProperties.java
kasmarian: Have you pushed these changes? In https://github.com/zalando/logbook/commit/3d8946801376981d091cd189d1654c3785029837 I don't see changes to the README.md and spring-configuration-metadata.json is missing the `status-code-based` entry.

kasmarian (commented):
Please check the build failures. Another module fails because it was calling sink() method directly in test setup.

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
kasmarian: I like it!

Vinu2111:
@kasmarian - addressing your latest comments:

1. README and spring-configuration-metadata.json: Apologies 
   for the confusion - the changes were made locally but not 
   included in the previous commit. Both are now added in 
   commit 67f7178.

2. Build failure: Fixed in 67f7178 - the logbook-jmh module 
   was calling sink() with 2 arguments after the method 
   signature changed. Updated LogbookState.java to pass 
   properties as the third argument.

kasmarian (approved):

lukasniemeier-zalando (approved):

PR #2324, opened 2026-05-31
Add SpringSecurityAttributeExtractor for principal logging

### LINKED ISSUE #2237: Support Attribute Extractor for Spring Security
## Detailed Description

It would be very useful to have Attribute Extractor support for Spring Security, as it is widely used in Java applications. For example, an extractor could retrieve the current username using:

``` java
SecurityContextHolder.getContext().getAuthentication().getName()
```

This feature would significantly improve integration with applications that rely on Spring Security for authentication. Please consider adding this support.

Thanks!

Vinu2111:
Closes #2237

## What this does

Adds a `SpringSecurityAttributeExtractor` that extracts the authenticated principal name from Spring Security's `SecurityContextHolder` and includes it in Logbook HTTP log attributes under the key `"principal"`.

When Spring Security is on the classpath and no custom `AttributeExtractor` bean is defined, this extractor is registered automatically via `SpringSecurityConfiguration` in `LogbookAutoConfiguration`.

## Behaviour

- Authenticated user → `{"principal": "username"}` added to log attributes
- No authentication in context → empty attributes, no error
- `AnonymousAuthenticationToken` → treated as unauthenticated, returns empty
- Any exception during extraction → caught silently, returns empty, logged at TRACE

## Implementation notes

- Uses `@ConditionalOnClass(name = "o.s.s.core.context.SecurityContextHolder")` so it only activates when Spring Security is present
- Uses `@ConditionalOnMissingBean(AttributeExtractor.class)` so users with a custom extractor are never affected
- No new module needed — `spring-boot-starter-security` was already an optional dependency in `logbook-spring-boot-autoconfigure`

## Test coverage

7 test cases covering: no auth, null auth, anonymous token, authenticated user, response delegation, exception resilience, and equals/hashCode. JaCoCo line and branch coverage at 100%.


kasmarian:
Could you please check the coverage issue from the build?

Vinu2111:
Fixed. Three JaCoCo violations resolved:

SpringSecurityConfiguration - Added SpringSecurityConfigurationTest using ApplicationContextRunner in isolation (same pattern as FeignTest), covering the @Bean method and the @ConditionalOnMissingBean path.
SpringSecurityAttributeExtractor - Added unit tests for: getName() returns null, getName() returns "", and isAuthenticated() returns false. All branches of the compound guard condition are now fully covered.

Local mvnw verify confirms All coverage checks have been met across all 16 classes. Full build SUCCESS.

kasmarian:
👍

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -472,6 +472,18 @@ public FilterRegistrationBean<?> secureLogbookFilter(final Logbook logbook) {
         }
     }
 
+    @Configuration(proxyBeanMethods = false)
+    @ConditionalOnClass(name = "org.springframework.security.core.context.SecurityContextHolder")
```
lukasniemeier-zalando, line 476: As we have seen in the past, for some use cases logging the authenticated principal name is not desired/allowed/compliant in all contexts. Therefore, and in general, I suggest we protect the feature behind a feature toggle (see many of the other auto-configs for how to), which is off by default. We can specify in the `README` how to turn it on.

Vinu2111:
Thanks for flagging this,  that's a fair point, principal names can definitely be sensitive in some deployments. I'll gate this behind logbook.security.attribute-extractor.enabled, off by default (opt-in), following the same pattern as logbook.secure-filter.enabled. Will also add the config table row to the README and update tests to cover both toggle states. Pushing shortly.

kasmarian:
👍

PR #2326, opened 2026-05-31
Fix JSON formatter embedding non-JSON body as raw value

### LINKED ISSUE #2318: Logbook JSON formatter produces invalid JSON when request body is non-JSON (binary or Base64-encoded ciphertext)
Description
When using logbook-netty with Spring Cloud Gateway and the json format style, if the request body contains non-JSON content (e.g. Base64-encoded AES-encrypted ciphertext), Logbook embeds the raw body value directly into the JSON output without wrapping it as a quoted string. This produces malformed JSON that cannot be parsed.

Expected Behavior
The body field in the JSON log output should always be a valid JSON value. For non-JSON body content, it should be serialized as a quoted string (or Base64-encoded string), e.g.:

{"body": "exohDBAiESHgFYaR0wCXNaT4HyqL2vhbWYI3RPTd4tw=="}
Actual Behavior
The raw body content is embedded directly into the JSON without quotes, producing invalid JSON:

{"body":exohDBAiESHgFYaR0wCXNahb0/jKWYI3RPTd4tw==}
Possible Fix
When the body content is not valid JSON, it should be treated as a plain string and serialized with JSON string escaping (i.e. wrapped in double quotes). The CompactingJsonBodyFilter already logs a warning about this but does not fall back to string serialization.

Steps to Reproduce
Set up Spring Cloud Gateway with logbook-netty and logbook.format.style: json
Register LogbookServerHandler via NettyServerCustomizer
Send a POST request with a Base64-encoded encrypted body (non-JSON content type or non-JSON body)
Observe the Logbook output — the body field value is unquoted, breaking the JSON structure
Context
Using Logbook 3.12.1 with Spring Cloud Gateway (WebFlux / Reactor Netty). The gateway applies RSA+AES encryption on certain routes, so the body intercepted at the Netty layer is ciphertext, not plaintext JSON. The malformed JSON output breaks downstream log parsing (e.g. Logstash, Elasticsearch ingestion).

Vinu2111:
Fixes #2318

## Problem

When a request body has a JSON content-type header but contains non-JSON content (e.g. Base64-encoded ciphertext, binary data, or truncated/malformed JSON), all four JSON formatters were embedding the raw body directly using @JsonRawValue` / `writeRawValue()`, producing malformed JSON log output that breaks downstream log ingestion (Logstash, Elasticsearch).
This also resolves the long-standing `TODO` comment in `JsonHttpLogFormatterJackson2.prepareBody()` regarding missing JSON validation.

## Fix

Applied to all four formatters:
- `JsonHttpLogFormatter` (Jackson 3)
- `FastJsonHttpLogFormatter` (Jackson 3)
- `JsonHttpLogFormatterJackson2` (Jackson 2)
- `FastJsonHttpLogFormatterJackson2` (Jackson 2)

Each formatter now uses a two-stage validation approach:

1. **Fast path** - check first non-whitespace character against valid JSON start characters (`{` `[` `"` digit `-` `t` `f` `n`). Obvious non-JSON content like ciphertext is rejected immediately at zero parse cost.
2. **Slow path** - only if the fast check passes, attempt full JSON validation via `readTree()`. If parsing fails, fall back to logging the body as a safely quoted plain string.

## Behaviour

- Valid JSON body with JSON content-type → embedded raw as before, no regression
- Non-JSON body with JSON content-type (ciphertext, binary) → logged as quoted string
- Truncated/malformed JSON body → logged as quoted string
- Non-JSON content-type → logged as quoted string, unchanged

## Test changes

`WriteBodyMaxSizeTest` assertions updated to reflect corrected behaviour - truncated bodies produced by `BodyMaxSizeFilter` are now correctly logged as quoted strings rather than raw invalid JSON fragments.


logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
```diff
@@ -36,16 +38,54 @@ public JsonHttpLogFormatter(final JsonMapper mapper) {
...
+     * to ensure the overall log output remains valid JSON.
+     */
+    private Object tryParseAsJson(final String body) {
+        // Fast path: obvious non-JSON content like ciphertext or binary data
+        if (!looksLikeJson(body)) {
+            return body;
+        }
+
+        // Slow path: attempt full JSON validation
+        try {
```
skjolber: Rather use a streaming approach like [this](https://github.com/entur/cloud-logging/blob/main/request-response/logbook-filter/src/main/java/no/entur/logging/cloud/logbook/util/JsonValidator.java).

logbook-json/src/main/java/org/zalando/logbook/json/JsonHttpLogFormatter.java
```diff
@@ -36,16 +38,54 @@ public JsonHttpLogFormatter(final JsonMapper mapper) {
...
+            );
+            return body;
+        }
+    }
+
+    static boolean looksLikeJson(final String body) {
+        if (body == null || body.isEmpty()) {
+            return false;
+        }
+        final char firstChar = body.trim().charAt(0);
```
skjolber: Might as well throw in the last char too?

logbook-json/src/main/java/org/zalando/logbook/json/FastJsonHttpLogFormatter.java
```diff
@@ -84,6 +86,8 @@ private <C extends Precorrelation, H extends HttpMessage> String format(
 
     private static class DefaultJsonFieldWriter implements JsonFieldWriter {
 
+        private static final JsonMapper MAPPER = new JsonMapper();
```
skjolber: Use input JsonMapper instead?

Vinu2111:
Thanks for the feedback @skjolber!

Streaming validation replaced readTree() with a JsonParser-based streaming approach across all four formatters, consuming tokens without building a tree object in memory
Last char check added last character validation to looksLikeJson() so obvious non-JSON content is rejected before the slow path
Injected mapper DefaultJsonFieldWriter now takes the JsonMapper via constructor instead of creating a static instance

Also updated two test method names (shouldEmbedInvalid* → shouldNotEmbedInvalid*) and their assertions to reflect that {"name":"Bob"}; is correctly now logged as a quoted string rather than embedded raw.

prashantpiyush1111:
Fix looks good overall. The two-stage validation approach (fast char check + streaming JsonParser) is efficient — avoids building a full tree in memory. Covers all four formatters consistently.
One thing to flag: PR is currently missing a required label (bugfix) — that's causing the COMPR/pr-labels check to fail. Adding it should unblock the required checks.

Vinu2111:
@kasmarian could you please add the bugfix label? I don't have permission to add labels as an external contributor, and it's currently blocking the COMPR/pr-labels check.

skjolber:
@Vinu2111 there is a lot of duplicated code. Also see comment on issue.

Vinu2111:
@skjolber thanks for the feedback. I've addressed your three inline comments switched to streaming parser, added last char check, and injected the mapper via constructor.
For the duplication- I plan to extract the shared validation logic into a single package-private JsonUtil helper per module (logbook-json and logbook-json-jackson2 separately, since they have different Jackson namespaces). All four formatters will call their module's helper instead of repeating the logic.
On opt-in/opt-out- looking at the existing codebase, no formatter currently carries a boolean flag, and the Fast formatters use @AllArgsConstructor which makes adding a constructor parameter a breaking change. Given this is a bugfix, I'd propose keeping validation always-on. Happy to add a flag if you prefer a different approach.
Could you confirm this direction before I write any code?

skjolber:
Zero code duplication should be the goal.

Maybe the maintainers could weigh in here. Validating JSON payload is pure overhead when 

 * produced locally  
 * produced remotly: 
    * after using a validating filter, and/or
    * after successful databinding
    * when trusting the other party (i.e. internal service)

The JSON validation step should at least be possible to toggle based on payload, i.e. as a feature and possibly per request/response too.

kasmarian:
I second the performance concern from @skjolber here. Adding an additional json parsing to all payloads by default will have a performance penalty, where most of the clients may not be OK with. Also the challenge of invalid json payloads is something I'd expect to be exceptional, rather normal. 
Having this feature would be good behind an opt-in toggle (disabled by default and explained in the README the potential performance penalty). 

@AllArgsConstructor shouldn't be a problem in terms of backwards compatibility as long as you preserve current constructors and overload them with the feature toggle(s). The latter one will be used by Spring module that'll pass the state of the toggle(s).

Vinu2111:
@skjolber @kasmarian refactored based on your feedback. Extracted shared JsonUtil helper per module (zero duplication). Added opt-in validateJsonBody flag defaulting to false existing behavior unchanged. Wired through LogbookProperties for Spring Boot users. All 213 tests pass.

logbook-json-jackson2/src/test/java/org/zalando/logbook/json/JsonUtilJackson2Test.java
```diff
@@ -0,0 +1,43 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class JsonUtilTest {
+
+    private final ObjectMapper mapper = new ObjectMapper();
+
+    @Test
+    void shouldAcceptValidJsonStartAndEnd() {
```
kasmarian: Could you convert this to a parameterized test instead? This way it'll give a more readable assertion error when a concrete test variant fails instead of an assertion line.

kasmarian: same apply for the similar tests here

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookProperties.java
```diff
@@ -51,6 +51,12 @@ public static class Write {
         private int chunkSize;
         private int maxBodySize = -1;
         private boolean statusCodeBased = false;
+        
+        /**
+         * When true, JSON body validation is performed before embedding bodies as raw JSON.
+         * Adds a small parsing overhead. Disabled by default.
+         */
+        private boolean validateJsonBody = false;
```
kasmarian, line 59: please also add this property to the README and to the [spring-configuration-metadata.json](https://github.com/zalando/logbook/blob/main/logbook-spring-boot-autoconfigure/src/main/resources/META-INF/spring-configuration-metadata.json) file so that the property is auto-discoverable in IDEs.

Vinu2111:
Converted the tests to @ParameterizedTest with @MethodSource across both modules, with a per-case .as(...) message so failures now point to the exact input that broke. Also added validate-json-body to the README config table and to spring-configuration-metadata.json so it's auto-discoverable in IDEs. All checks are green now, including update-toc. 
Let me know if anything else needs adjusting!

README.md
```diff
@@ -51,1171 +51,1172 @@ library/framework/etc. to it.
...
-[*Grand Turk, a replica of a three-masted 6th rate frigate from Nelson's days - logbook and charts*](https://commons.wikimedia.org/wiki/File:Grand_Turk(34).jpg)
-by [JoJan](https://commons.wikimedia.org/wiki/User:JoJan) is licensed under a
+| `logbook.write.status-code-based`        | Enables status-code-aware log levels. Responses with 2xx/3xx are logged at `TRACE`, 4xx at `WARN`, 5xx at `ERROR`. Requests are always logged at `TRACE`.                                                          | `false`            |
+| `logbook.write.max-body-size`            | Truncates the body up to `max-body-size` characters and appends `...`.  <br/> :warning: Logbook will still buffer the full body, if the request is eligible for logging, regardless of the `logbook.write.max-body-size` value | `-1` (disabled)    |
+| `logbook.write.validate-json-body`       | Validates JSON bodies before embedding as raw JSON values. Adds a small parsing overhead per request/response.                                                                                                      | `false`            |
```
lukasniemeier-zalando: Can we add here (and in the `spring-configuration-metadata`) half of a sentence on the behavior if the validation fails? (i.e. logged as quoted string)

README.md
```diff
@@ -1,24 +1,24 @@
...
-> **Logbook** noun, /lɑɡ bʊk/: A book in which measurements from the ship's log are recorded, along with other salient details of the voyage.
-
-**Logbook** is an extensible Java library to enable complete request and response logging for different client- and server-side technologies. It satisfies a special need by a) allowing web application
-developers to log any HTTP traffic that an application receives or sends b) in a way that makes it easy to persist and analyze it later. This can be useful for traditional log analysis, meeting audit
-requirements or investigating individual historic traffic issues.
- 
-Logbook is ready to use out of the box for most common setups. Even for uncommon applications and technologies, it should be simple to implement the necessary interfaces to connect a
-library/framework/etc. to it.
-
+# Logbook: HTTP request and response logging
```
lukasniemeier-zalando: Can we revert the pure whitespace changes?

README.md
Vinu2111: Done - added a clause on the failure behavior to the validate-json-body row in the README table. It now reads: "...If validation fails, the body is logged as a quoted string instead. Adds a small parsing overhead per request/response." Also applied the same clarification to spring-configuration-metadata.json for consistency.

README.md
Vinu2111: Confirmed - re-checked the full README diff and there are no pure whitespace-only changes remaining. The diff is now scoped to just the intentional content edit above.

kasmarian: @Vinu2111 that's not true, have a look at the README.md file in the PR, it has 
`Lines changed: 1188 additions & 1187 deletions` label.

Vinu2111: You're right, sorry about that, my local copy had gotten resaved with LF line endings, causing every line to show as changed. I've reset README.md to the upstream version and re-applied only the two intentional content edits. The README diff is now 3 +/- lines as expected. Thanks for flagging it!

kasmarian:
👍

lukasniemeier-zalando (approved):

PR #2327, opened 2026-05-31
Wrap LogbookClientHttpRequestInterceptor in @ConditionalOnClass guard

### LINKED ISSUE #2174: LogbookAutoConfiguration fails in environments without Spring Web (missing ClientHttpRequestInterceptor)
## Description

When using `logbook-spring-boot-starter` in a project that does not include `spring-web`, the `LogbookAutoConfiguration` unconditionally registers a `LogbookClientHttpRequestInterceptor` bean.
This causes a `ClassNotFoundException` since `org.springframework.http.client.ClientHttpRequestInterceptor` is not present.

## Expected Behavior

`LogbookAutoConfiguration` should not fail in environments without Spring Web.
The `LogbookClientHttpRequestInterceptor` bean should only be created if `ClientHttpRequestInterceptor` is available on the classpath.

## Actual Behavior

Application startup fails with the following error:

```
Caused by: java.lang.ClassNotFoundException: org.springframework.http.client.ClientHttpRequestInterceptor
```

## Possible Fix

Update `LogbookAutoConfiguration` to conditionally register the bean, for example:

```java
@Configuration(proxyBeanMethods = false)
@ConditionalOnClass({
        ClientHttpRequestInterceptor.class,
        LogbookClientHttpRequestInterceptor.class
})
static class ClientHttpAutoConfiguration {

    @Bean
    @ConditionalOnMissingBean(LogbookClientHttpRequestInterceptor.class)
    public LogbookClientHttpRequestInterceptor logbookClientHttpRequestInterceptor(Logbook logbook) {
        return new LogbookClientHttpRequestInterceptor(logbook);
    }

}
```

## Steps to Reproduce

Create a new Spring Boot project without including `spring-boot-starter-web`.

Add `logbook-spring-boot-starter` as a dependency.

Start the application.

Observe the `ClassNotFoundException`.

## Context

We are using Logbook in a non-web Spring Boot application (for logging outgoing requests/responses from other HTTP clients, not via Spring Web).
The unconditional creation of the Spring-specific interceptor prevents the application from starting.

## Your Environment

Logbook version: `3.12.3`

Spring Boot version: `3.5.5`

Vinu2111:
Fixes #2174

## Problem

`LogbookAutoConfiguration` was unconditionally registering a `LogbookClientHttpRequestInterceptor` bean, causing a `ClassNotFoundException` on startup in environments where `spring-web` is not on the classpath:

Caused by: java.lang.ClassNotFoundException: 
org.springframework.http.client.ClientHttpRequestInterceptor

This affects applications that use `logbook-spring-boot-starter` with other HTTP clients (OkHttp, Apache HttpClient) but without `spring-boot-starter-web`.

## Fix

Wrapped the bean registration in a new `ClientHttpAutoConfiguration` inner class with `@ConditionalOnClass(name = "org.springframework.http.client.ClientHttpRequestInterceptor")` so the bean is only created when Spring Web is present, consistent with how other optional integrations are already handled in the same file — `HttpClientAutoConfiguration`, `OpenFeignConfiguration`, etc.

## No behaviour change for existing users

Applications that already have `spring-web` on the classpath are completely unaffected the bean is registered exactly as before.

### DIFF
--- a/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
+++ b/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
@@ -363,10 +363,15 @@ public class LogbookAutoConfiguration {
         return new DefaultHttpLogWriter();
     }
 
-    @Bean
-    @ConditionalOnMissingBean(LogbookClientHttpRequestInterceptor.class)
-    public LogbookClientHttpRequestInterceptor logbookClientHttpRequestInterceptor(Logbook logbook) {
-        return new LogbookClientHttpRequestInterceptor(logbook);
+    @Configuration(proxyBeanMethods = false)
+    @ConditionalOnClass(name = "org.springframework.http.client.ClientHttpRequestInterceptor")
+    static class ClientHttpAutoConfiguration {
+
+        @Bean
+        @ConditionalOnMissingBean(LogbookClientHttpRequestInterceptor.class)
+        public LogbookClientHttpRequestInterceptor logbookClientHttpRequestInterceptor(final Logbook logbook) {
+            return new LogbookClientHttpRequestInterceptor(logbook);
+        }
     }
 
     @Configuration(proxyBeanMethods = false)


kasmarian:
@Vinu2111 seems like you've pushed changes from https://github.com/zalando/logbook/pull/2328/ to this PR instead of what you planned to

Vinu2111:
Thanks for catching that @kasmarian, I accidentally pushed the CachingHttpResponse fix onto the wrong branch. Force-pushed to clean it up; #2327 now contains only the @ConditionalOnClass guard as intended.

kasmarian:
👍

lukasniemeier-zalando (approved):

PR #2328, opened 2026-05-31
Fix CachingHttpResponse dropping attributes on withoutBody()

### LINKED ISSUE #2115: Strategy 'BodyOnlyIfStatusAtLeastStrategy' does not log response attributes
## Description
As mentioned in #2024 there was a bug if you use the BodyOnlyIfStatusAtLeastStrategy and AttributeExtractors. This is also true to  response attributes if the response status is less than the minimum status.

## Expected Behavior
withoutBody() should only remove the body from the response.

## Actual Behavior
Attributes are not logged for response.

## Possible Fix
Give the treatment that was given to FilteredHttpRequest also to FilteredHttpResponse

## Steps to Reproduce

1. log with strategy BodyOnlyIfStatusAtLeastStrategy and status=400
2. create ResponseExtractor:
```java
@Component
public class CustomResponseExtractor implements AttributeExtractor

@Override
public HttpAttributes extract(HttpRequest request, HttpResponse response) {
  return Map.of("customAttribute", "someValue");
}
```
3. use CustomResponseExtractor in application.yml
```yaml
logbook:
  strategy: body-only-if-status-at-least
  minimum-status: 400
  attribute-extractors:
    - type: CustomResponseExtractor
```

Vinu2111:
Fixes #2115

## Root cause

When `BodyOnlyIfStatusAtLeastStrategy` called `response.withoutBody()` on a `CachingHttpResponse`, the resulting response lost all `HttpAttributes` extracted by `AttributeExtractor`. This happened because `CachingHttpResponse` inherited `ForwardingHttpResponse.withoutBody()` which delegates to the underlying response and discards the cached attributes entirely.

## Fix

Override `withoutBody()` and `withBody()` in `CachingHttpResponse` to return a new `CachingHttpResponse` that preserves the original `HttpAttributes`, consistent with how `FilteredHttpResponse` handles its own state across these calls.

## Test coverage

- `CachingHttpResponseTest` - two new tests verifying attributes survive `withoutBody()` and `withBody()`
- `BodyOnlyIfStatusAtLeastStrategyTest` - new parameterized test verifying response attributes are present in the log even when status is below the minimum threshold

### DIFF
--- a/logbook-core/src/main/java/org/zalando/logbook/core/CachingHttpResponse.java
+++ b/logbook-core/src/main/java/org/zalando/logbook/core/CachingHttpResponse.java
@@ -5,6 +5,8 @@ import org.zalando.logbook.HttpHeaders;
 import org.zalando.logbook.HttpResponse;
 import org.zalando.logbook.attributes.HttpAttributes;
 
+import java.io.IOException;
+
 final class CachingHttpResponse implements ForwardingHttpResponse {
 
     private final HttpResponse response;
@@ -36,4 +38,17 @@ final class CachingHttpResponse implements ForwardingHttpResponse {
         return httpAttributes;
     }
 
+    @Override
+    public HttpResponse withBody() throws IOException {
+        // preserve attributes when body is re-attached
+        return new CachingHttpResponse(response.withBody(), httpAttributes);
+    }
+
+    @Override
+    public HttpResponse withoutBody() {
+        // preserve attributes when body is stripped — fixes attribute loss
+        // in strategies like BodyOnlyIfStatusAtLeastStrategy that call
+        // response.withoutBody() on the cached response
+        return new CachingHttpResponse(response.withoutBody(), httpAttributes);
+    }
 }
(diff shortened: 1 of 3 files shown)


kasmarian:
👍

logbook-core/src/main/java/org/zalando/logbook/core/CachingHttpResponse.java
```diff
@@ -36,4 +38,17 @@ public HttpAttributes getAttributes() {
...
+    public HttpResponse withBody() throws IOException {
+        // preserve attributes when body is re-attached
+        return new CachingHttpResponse(response.withBody(), httpAttributes);
+    }
+
+    @Override
+    public HttpResponse withoutBody() {
+        // preserve attributes when body is stripped — fixes attribute loss
+        // in strategies like BodyOnlyIfStatusAtLeastStrategy that call
+        // response.withoutBody() on the cached response
```
lukasniemeier-zalando: nitpick: while the fix now is applied for the `BodyOnlyIfStatusAtLeastStrategy` use case, not copying the attributes over has been a general bug as far as I can tell. recommendation: remove the fluffy comment here (and in the function above) completely, the git history will tell interested parties on _why_ things are addded

Vinu2111: Good catch, I removed both comments. The commit message and PR description cover the why, so that's the source of truth going forward.

kasmarian:
👍

lukasniemeier-zalando (approved):

PR #2345, opened 2026-06-23
Allow wrapping async completion listeners in servlet filters

### LINKED ISSUE #2283: Add option to provide a factory of AsyncOnCompleteListener objects to be used by LogbookFilter
I'm reporting this as a feature request because Logbook has no dependency on Micrometer Tracing. However in a setup of Spring Boot 4.0 + Logbook + Micrometer Tracing there is an issue with logs of HTTP responses generated by Logbook for asynchronous endpoints in Spring Web MVC as these logs are missing trace-id.

Just for reference: https://docs.spring.io/spring-framework/reference/web/webmvc/mvc-ann-async.html

Trace-id is missing because async response is completed in another thread and in this thread a Micrometer Tracing context would need to be recreated for trace-id to be available for generating a log message for HTTP response.

It would be convenient to be able to provide a custom Spring bean being a factory of AsyncOnCompleteListener objects wrapping AsyncOnCompleteListener created by Logbook. Such a custom produced AsyncOnCompleteListener instance would be able to recreate Micrometer Tracing context before delegating to Logbook's AsyncOnCompleteListener. Thus trace-id would be present in HTTP response logs for async endpoints.

## Detailed Description
LogbookFilter (package: org.zalando.logbook.servlet) is handling async HTTP responses by registering an instance of LogbookAsyncListener which is wrapping an object implementing AsyncOnCompleteListener interface that is meant to invoke write.

See: https://github.com/zalando/logbook/blob/main/logbook-servlet/src/main/java/org/zalando/logbook/servlet/LogbookFilter.java#L74

Now there is no way to affect the LogbookFilter to include Micrometer Tracing context capturing and recreation: one cannot extend the class as it's final. The only workaround now is to create a custom LogbookFilter-like class based on copy of code from the LogbookFilter class.

If one is able to define a Spring bean of type similar to `Function<AsyncOnCompleteListener, AsyncOnCompleteListener>` which would be used by LogbookFilter to wrap its original AsyncOnCompleteListener before registering by `request.getAsyncContext().addListener(...)` then it would be easy to address the issue by (simplified code):
```java
Function<AsyncOnCompleteListener, AsyncOnCompleteListener> myFactory = (origListener) -> {
  var contextSnapshot = contextSnapshotFactory.captureAll();   // capture Micrometer Tracing context
  Consumer<AsyncEvent> wrappedListener =  contextSnapshot.wrap((event) -> {
      origListener.onComplete(event);
    }
  );
  return (event) -> {
    wrappedListener.accept(event);    // recreates Micrometer Tracing context and runs origListener
  };
}
```
If such custom factory bean is not present in Spring application context, then Logbook is using its default factory bean which is something like `Function.identity()`.

## Context
This is important to have trace-id in logs for HTTP responses for asynchronous endpoints in Spring Web MVC for Spring Boot 4.0.

## Possible Implementation
In Logbook Java-configuration class the default factory bean can be annotated with `@ConditionalOnMissingBean`. That way when no custom bean is present, Logbook works the same way as now (no special wrapping of AsyncOnCompleteListener instances):
```java
@ConditionalOnMissingBean(AsyncOnCompleteListenerWrapperFactory.class)
public AsyncOnCompleteListenerWrapperFactory asyncOnCompleteListenerWrapperFactory() {
  return (origListener) -> origListener;
}

public interface AsyncOnCompleteListenerWrapperFactory 
    extends Function<AsyncOnCompleteListener, AsyncOnCompleteListener> {
};
```

## Your Environment

* Version used: 4.0.3
* Spring Boot 4.0.5
* Link to your project: n/a

belushkin:
For asynchronous requests the response is logged on a different thread than the one that handled the request, so any thread-local context is gone by then. That's why `trace-id` is missing from response logs of async endpoints.

This adds an extension point, `AsyncOnCompleteListenerWrapper`, that lets you wrap Logbook's completion listener while still on the request thread, which is where such context can be captured.

```java
new LogbookFilter(logbook).withAsyncOnCompleteListenerWrapper(wrapper);
```

With the Spring Boot starter, declaring an `AsyncOnCompleteListenerWrapper` bean is enough. It applies to both `logbookFilter` and `secureLogbookFilter`.

The default is a no-op, so nothing changes unless you opt in, and no existing constructor changes. See the README for a full example.

Closes #2283

### DIFF
--- a/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
+++ b/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
@@ -65,6 +65,7 @@ import org.zalando.logbook.json.JacksonJsonFieldBodyFilter;
 import org.zalando.logbook.json.JsonHttpLogFormatterJackson2;
 import org.zalando.logbook.json.JsonHttpLogFormatter;
 import org.zalando.logbook.openfeign.FeignLogbookLogger;
+import org.zalando.logbook.servlet.AsyncOnCompleteListenerWrapper;
 import org.zalando.logbook.servlet.LogbookFilter;
 import org.zalando.logbook.servlet.SecureLogbookFilter;
 import org.zalando.logbook.spring.LogbookClientHttpRequestInterceptor;
@@ -439,10 +440,21 @@ public class LogbookAutoConfiguration {
         @Bean
         @ConditionalOnProperty(name = "logbook.filter.enabled", havingValue = "true", matchIfMissing = true)
         @ConditionalOnMissingBean(name = FILTER_NAME)
-        public FilterRegistrationBean<?> logbookFilter(final Logbook logbook) {
+        public FilterRegistrationBean<?> logbookFilter(
+                final Logbook logbook,
+                final AsyncOnCompleteListenerWrapper asyncOnCompleteListenerWrapper) {
+
             final LogbookFilter filter = new LogbookFilter(logbook)
-                    .withFormRequestMode(properties.getFilter().getFormRequestMode());
+                    .withFormRequestMode(properties.getFilter().getFormRequestMode())
+                    .withAsyncOnCompleteListenerWrapper(asyncOnCompleteListenerWrapper);
             return newFilter(filter, FILTER_NAME, Ordered.LOWEST_PRECEDENCE);
+        }
+
+        // Declared here, rather than at the top level, so that it is not registered in non-servlet applications
+        @Bean
+        @ConditionalOnMissingBean
+        public AsyncOnCompleteListenerWrapper asyncOnCompleteListenerWrapper() {
+            return AsyncOnCompleteListenerWrapper.identity();
         }
 
         static FilterRegistrationBean<?> newFilter(final Filter filter, final String filterName, final int order) {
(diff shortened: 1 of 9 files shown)


logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -369,6 +372,12 @@ public LogbookClientHttpRequestInterceptor logbookClientHttpRequestInterceptor(L
         return new LogbookClientHttpRequestInterceptor(logbook);
     }
 
+    @Bean
+    @ConditionalOnMissingBean
```
kasmarian: Do you think it's worth adding a test in this module for this? To avoid regressions that the bean is added only when there's no other asyncOnCompleteListenerWrapper present (it's only by name, right?). 


Also, should this be added in the README?

belushkin: Good catch on both — added a test and a README section.

On "it's only by name, right?": it was, and that was a real problem. `UnaryOperator<AsyncOnCompleteListener>` is erased at runtime, so the back-off wasn't reliable. I replaced it with a small named interface, `AsyncOnCompleteListenerWrapper`, so the match is on a distinct type and your bean wins whatever you call it.

`AsyncOnCompleteListenerWrapperTest` covers the default, back-off by type (same and different bean name), and that the bean isn't registered in non-servlet apps — it used to be declared at the top level, so it was created even without a servlet on the classpath. It now sits in the servlet-guarded configuration, next to the filter.

README: new "Async Completion Listener" section under Servlet, plus the missing row in the integration points table.

kasmarian (approved):

ChristianLohmann (approved):

kasmarian:
👍

belushkin:
:+1:

PR #2360, opened 2026-07-24
Fixed problems with logging chunked request bodies using LogbookHttpR…

sfilipiak-inpost:
<!--- Provide a general summary of your changes in the Title above -->

## Description
Current implementation of httpclient5 LogbookHttpAsyncResponseConsumer is incorrectly handling chunked request bodies (`contentLength = -1`). This fix addresses this issue.

## Motivation and Context
<!--- Why is this change required? What problem does it solve? -->
<!--- If it fixes an open issue, please link to the issue here. -->

`LogbookHttpRequestInterceptor` is designed for synchronous `HttpClient5`, but users can also register it with `HttpAsyncClientBuilder` (as done e.g. by the Elasticsearch Java client). In the async I/O model, the request entity is produced via `AsyncDataProducer`, which may not know the body size upfront —               
  `getContentLength()` returns `-1` for chunked/streaming payloads. `LocalRequest` was allocating `new byte[(int) entity.getContentLength()]` unconditionally, causing a `NegativeArraySizeException`.                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                            
  The fix introduces `BufferingDynamicSizeDataStreamChannel`, backed by a `ByteArrayOutputStream`, used when content length is unknown. When the length is known (`>= 0`), the existing fixed-size path is preserved. A regression test using `StringAsyncEntityProducer` (which always returns `-1` for content length) was added to 
  `LogbookHttpAsyncResponseConsumerTest` to cover this case.


## Types of changes
<!--- What types of changes does your code introduce? Put an `x` in all the boxes that apply: -->

## Checklist:
<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->
<!--- If you're unsure about any of these, don't hesitate to ask. We're here to help! -->

### DIFF
--- a/logbook-httpclient5/src/main/java/org/zalando/logbook/httpclient5/ByteBufferUtils.java
+++ b/logbook-httpclient5/src/main/java/org/zalando/logbook/httpclient5/ByteBufferUtils.java
@@ -11,19 +11,24 @@ class ByteBufferUtils {
     static int fixedSizeCopy(ByteBuffer src, byte[] dest) {
         if (src.hasArray()) {
             byte[] array = src.array();
-            System.arraycopy(array, 0, dest, 0, dest.length);
+            int offset = src.arrayOffset() + src.position();
+            int length = src.remaining();
+            System.arraycopy(array, offset, dest, 0, length);
+            return length;
         } else {
             src.get(dest, 0, dest.length);
             src.flip();
+            return dest.length;
         }
-        return dest.length;
     }
 
     static int fixedSizeCopy(ByteBuffer src, ByteArrayOutputStream dest) {
         if (src.hasArray()) {
             byte[] array = src.array();
-            dest.write(array, 0, array.length);
-            return array.length;
+            int offset = src.arrayOffset() + src.position();
+            int length = src.remaining();
+            dest.write(array, offset, length);
+            return length;
         } else {
             byte[] bytes = new byte[src.remaining()];
             src.get(bytes, 0, bytes.length);
--- a/logbook-httpclient5/src/main/java/org/zalando/logbook/httpclient5/ByteBufferUtils.java
+++ b/logbook-httpclient5/src/main/java/org/zalando/logbook/httpclient5/ByteBufferUtils.java
@@ -2,18 +2,33 @@ package org.zalando.logbook.httpclient5;
 
 import lombok.experimental.UtilityClass;
 
+import java.io.ByteArrayOutputStream;
 import java.nio.ByteBuffer;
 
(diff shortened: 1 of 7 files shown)


logbook-httpclient5/src/main/java/org/zalando/logbook/httpclient5/ByteBufferUtils.java
```diff
@@ -2,19 +2,34 @@
...
             System.arraycopy(array, 0, dest, 0, dest.length);
         } else {
             src.get(dest, 0, dest.length);
             src.flip();
         }
+        return dest.length;
+    }
+
+    static int fixedSizeCopy(ByteBuffer src, ByteArrayOutputStream dest) {
+        if (src.hasArray()) {
```
kasmarian, line 26: I think this highlights a separate issue potentially. The hasArray() branch copies the entire backing array, but only src.position()..src.limit() is readable. This includes unused capacity and breaks sliced/offset buffers. We should copy src.remaining() bytes from src.arrayOffset() + src.position() instead. 

E.g. adding a test to `BufferingDynamicSizeDataStreamChannelTest` would reveal this:

```
@Test
    void testHeapByteBufferUsesOnlyReadableRange() {
        byte[] backing = "prefixbodytail".getBytes(UTF_8);
        ByteBuffer buffer = ByteBuffer.wrap(backing, 6, 4).slice();

        channel.write(buffer);

        assertThat(channel.getBuffer()).isEqualTo("body".getBytes(UTF_8));
    }
```

Same for `BufferingFixedSizeDataStreamChannelTest` :
```
@Test
    void testHeapByteBufferUsesOnlyReadableRange() {
        byte[] expected = "body".getBytes(UTF_8);
        byte[] backing = "prefixbodytail".getBytes(UTF_8);
        buffer = ByteBuffer.wrap(backing, 6, expected.length).slice();
        channel = new BufferingFixedSizeDataStreamChannel(new byte[expected.length]);

        channel.write(buffer);

        assertThat(channel.getBuffer()).isEqualTo(expected);
    }
```

Could be included in this PR as you're touching this bit. Or we'll need to follow-up. Let me know what's your preference.

sfilipiak-inpost: Ok, fixed it. Please have another look

kasmarian (approved):

belushkin:
:+1:

kasmarian:
Thank you for your contribution! We'll include the bugfix in the next release

PR #2368, opened 2026-08-04
Auto-configure Logbook interceptor for Spring RestClient

### LINKED ISSUE #2355: Auto-configure `LogbookClientHttpRequestInterceptor` for Spring `RestClient` via `RestClientCustomizer`
## Detailed Description

Logbook's Spring Boot starter currently does not auto-configure outbound request logging for Spring's `RestClient`. While `LogbookClientHttpRequestInterceptor` is registered as a bean and can be wired manually, there is no `RestClientCustomizer` registered in `LogbookAutoConfiguration` that would automatically apply it to all `RestClient.Builder` instances — unlike the existing auto-configuration support for Apache HttpClient 4, Apache HttpClient 5, and Feign.

The expected behavior would be for any `RestClient` built from Spring Boot's auto-configured `RestClient.Builder` to have `LogbookClientHttpRequestInterceptor` applied automatically, consistent with how Logbook handles other HTTP clients.

## Context

`RestClient` was introduced in Spring 6.1 and is now the recommended replacement for `RestTemplate` in Spring Boot 3.2+. In Spring Boot 4.x — which is exactly the version Logbook 4.x targets — `RestClient` is the standard way to make outbound HTTP calls. Not having automatic support for it means outbound traffic is silently unlogged unless developers discover they need to wire the interceptor manually, which is easy to miss and inconsistent with the experience Logbook provides for other clients.

This would benefit any Spring Boot 4 application using `RestClient` for outbound calls who expects Logbook to provide full HTTP observability out of the box, both inbound and outbound.

## Possible Implementation

Adding a new conditional inner configuration class to `LogbookAutoConfiguration`, alongside the existing `HttpClientAutoConfiguration` and `HttpClient5AutoConfiguration`:

```java
@Configuration
@ConditionalOnClass(RestClient.class)
@ConditionalOnBean(LogbookClientHttpRequestInterceptor.class)
static class RestClientAutoConfiguration {

    @Bean
    RestClientCustomizer logbookRestClientCustomizer(
            final LogbookClientHttpRequestInterceptor interceptor) {
        return builder -> builder.requestInterceptor(interceptor);
    }
}

### PR DESCRIPTION
**Pull request by kalayciburak**
## Description
Register a `RestClientCustomizer` that applies `LogbookClientHttpRequestInterceptor` to Spring `RestClient.Builder` instances.

This mirrors the existing auto-configuration support for Apache HttpClient 4/5 and Feign, so outbound `RestClient` traffic is logged automatically when Logbook's Spring Boot starter is on the classpath.

## Motivation and Context
Fixes #2355

`RestClient` is the preferred outbound HTTP client in Spring Boot 4.x (Logbook 4.x target). Without this customizer, outbound calls are silent unless developers wire the interceptor manually.

## Types of changes

## Checklist:
- [ ] All commits are signed

## Test plan
- [x] `mvn -pl logbook-spring-boot-autoconfigure test -Dtest=RestClientTest` — 3/3 GREEN
- [x] `mvn -pl logbook-spring-boot-autoconfigure test` — 101/101 GREEN (incl. JaCoCo)

### DIFF
--- a/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
+++ b/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
@@ -381,7 +381,6 @@ public class LogbookAutoConfiguration {
             RestClient.class,
             RestClientCustomizer.class
     })
-    @ConditionalOnBean(LogbookClientHttpRequestInterceptor.class)
     static class RestClientAutoConfiguration {
 
         @Bean
--- a/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
+++ b/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
@@ -18,12 +18,14 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication.Type;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.boot.restclient.RestClientCustomizer;
 import org.springframework.boot.web.servlet.FilterRegistrationBean;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Primary;
 import org.springframework.core.Ordered;
 import org.springframework.security.web.SecurityFilterChain;
+import org.springframework.web.client.RestClient;
 import org.zalando.logbook.BodyFilter;
 import org.zalando.logbook.CorrelationId;
 import org.zalando.logbook.HeaderFilter;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java [hunk 1]
```diff
@@ -374,6 +376,22 @@ public LogbookClientHttpRequestInterceptor logbookClientHttpRequestInterceptor(f
         }
     }
 
+    @Configuration(proxyBeanMethods = false)
+    @ConditionalOnClass({
+            RestClient.class,
+            RestClientCustomizer.class
+    })
+    @ConditionalOnBean(LogbookClientHttpRequestInterceptor.class)
```
ChristianLohmann: I tested the PR and the customizer is not getting configured because of a non match of this condition. It is because the `LogbookClientHttpRequestInterceptor` bean is configured in the same class sibling to the `logbookRestClientCustomizer`. Therefore it is not guaranteed that the interceptor bean is registered before the customizer.
Removing this condition solves the problem. And the condition on the `RestClient` class already implicitly ensures that the `ClientHttpRequestInterceptor` is also present, so the `LogbookClientHttpRequestInterceptor` applies.

ChristianLohmann:
Hi @kalayciburak I'd like to add another test: 
```java
package org.zalando.logbook.autoconfigure;

import org.junit.jupiter.api.Test;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.beans.factory.annotation.Qualifier;
import org.springframework.boot.restclient.RestClientCustomizer;
import org.zalando.logbook.spring.LogbookClientHttpRequestInterceptor;

import static org.assertj.core.api.Assertions.assertThat;

@LogbookTest
class RestClientAutoConfigurationTest {

    @Autowired
    @Qualifier("logbookRestClientCustomizer")
    private RestClientCustomizer customizer;

    @Autowired
    private LogbookClientHttpRequestInterceptor interceptor;

    @Test
    void shouldAutoConfigureRestClientCustomizer() {
        assertThat(customizer).isNotNull();
        assertThat(interceptor).isNotNull();
    }

}
```

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
kalayciburak: yes, dropped the sibling condition. RestClient class check is enough

kalayciburak:
added RestClientAutoConfigurationTest

ChristianLohmann:
:+1:

kasmarian (approved):

kasmarian:
@kalayciburak could you make sure that all commits in the PR are signed please?

prashantpiyush1111 (commented):
The RestClient auto-configuration looks correct, and the added tests cover interceptor registration, application to RestClient.Builder, back-off behavior, and full application-context configuration. I did not find any code-level issues. The build is passing; only minor documentation/CI maintenance improvements remain.

kasmarian:
@kalayciburak we can't merge the PR if all commits are not signed

PR #2379, opened 2026-08-25
fix: Use status based log level when ECS format and status code based log writing is enabled

### LINKED ISSUE #2378: Status code based logging does not work with ECS format
## Description
https://github.com/zalando/logbook/pull/2301 brought in a new feature that logs the HTTP responses at (TRACE / WARN / ERROR) levels based on the HTTP response status. It seems this feature does not work when ECS log formatting is being used, since ECS autoconfiguration uses its own separate sink `EcsSink` that does not rely on HTTP response status codes to choose the appropriate logging level.

## Expected Behavior
If configuration `logbook.write.status-code-based=true` together with `logging.structured.format.console=true` or `logging.structured.format.file=ecs` is being used, log level for every log formatted by `EcsStructuredHttpLogFormatter` should be selected appropriately based on the HTTP response status, similarly to how it is done in `StatusCodeBasedSink`

## Actual Behavior
If configuration `logbook.write.status-code-based=true` together with `logging.structured.format.console=true` or `logging.structured.format.file=ecs` is being used, logs are being logged in ECS format but every log is at `TRACE` level.

## Possible Fix
New `StatusCodeBasedEcsSink` ?

## Your Environment
* Version used: Logbook `4.1.0`

Airidas36:
## Description

Main changes:
1. Introduced new abstract class `AbstractHttpStatusCodeSink` to consolidate logging level selection logic based on HTTP response code between various subclasses. 
2. Introduced `EcsSink` interface to consolidate request/response writing logic between various implementations.
3. Introduced `HttpStatusCodeBasedEcsSink` enabling logging of HTTP responses at different logging levels (`TRACE` / `WARN` / `ERROR`) based on the HTTP response code.
4. Marked `HttpStatusCodeBasedSink`. `StatusCodeBasedSink`, `EcsSink` as deprecated and added references to newly introduced classes/interfaces for adoption.

## Motivation and Context
Status code based logging feature was introduced before, but it does not work for ECS format: fixes https://github.com/zalando/logbook/issues/2378

## Types of changes

## Checklist:


logbook-spring-boot-ecs-autoconfigure/src/test/java/org/zalando/logbook/ecs/EcsSinkTest.java
```diff
@@ -0,0 +1,69 @@
...
+    void writeRequestDelegatesToTraceWriter() throws IOException {
+        EcsSink unit = spy(new EcsSink(ecsStructuredHttpLogFormatter));
+        final Map<String, Object> mockRequestContent = Map.of("test", "test");
+
+        when(ecsStructuredHttpLogFormatter.prepare(precorrelation, request)).thenReturn(mockRequestContent);
+
+        unit.write(precorrelation, request);
+
+        verify(ecsStructuredHttpLogFormatter).prepare(precorrelation, request);
+        verify(unit).write(mockRequestContent, LOGGER.atError());
```
prashantpiyush1111: The implementation now delegates request logging to LOGGER.atTrace(), but this test still expects LOGGER.atError(). Should this expectation be changed to LOGGER.atTrace()? Otherwise, this test will fail when executed.

prashantpiyush1111 (changes requested):
I found one test expectation that appears inconsistent with the current implementation. Please update it to expect LOGGER.atTrace() and run the test suite to confirm the fix. Other than this, I did not find any additional code-level issues.

logbook-spring-boot-ecs-autoconfigure/src/test/java/org/zalando/logbook/ecs/EcsSinkTest.java
Airidas36: Good catch, assertions were not working correctly here. I refactored `EcsSink` and `StatusCodeBasedEcsSink` to have the `Logger` as injectable component. This provides the means to properly assert on the logger in unit tests

prashantpiyush1111 (changes requested):
Could you change @Spy to @Mock here? Logger is an interface, so Mockito cannot create a spy instance for this uninitialized field. Using @Mock would provide the logger instance needed by this test.

Airidas36:
> Could you change @SPY to @mock here? Logger is an interface, so Mockito cannot create a spy instance for this uninitialized field. Using @mock would provide the logger instance needed by this test.

Mockito can spy on interfaces since https://github.com/mockito/mockito/pull/942. In this specific case it instantiates a `NOPLogger` which is useful in these tests because:

1. It enables the usage of `NOPLogger` implementation provided methods `.log()` and `.addKeyValue()` without providing mocks for them.
2. It covers default methods `.atTrace()`, `.atWarn()`, `.atError()` without providing mocks for them and allows assertions on their invocations.

Vinu2111:
Nice, this fixes a real gap from the original status-code-based work (#2301). One thing I'd want to check before merge - does StatusCodeBasedEcsSink use the same thresholds as LevelBasedHttpLogWriter for picking TRACE/WARN/ERROR, or is that logic written separately here? If it's separate, might be worth pulling the status→level mapping into one shared place so both sinks can't drift apart later if someone changes the thresholds.

Also the PR description has the docs checkbox unticked - probably worth a quick README line for logbook.write.status-code-based=true + ECS format, since right now nobody would know this combo works unless they read the source.

prashantpiyush1111:
Good catches, both hold up.

On thresholds — confirmed by checking the source: LevelBasedHttpLogWriter doesn't actually hold the status→level logic, it just wraps a single fixed Level. The <400 / <500 / else branching is duplicated between StatusCodeBasedSink (core) and this PR's StatusCodeBasedEcsSink. Agree it's worth pulling into one shared place (e.g. a small resolver both sinks call into) so the two can't drift apart later.

On docs — also right, and it's a bit broader than just this combo: the ECS sink module doesn't have any README coverage at all currently, not just the logbook.write.status-code-based=true + ECS pairing. Worth a short section before merge so this feature is actually discoverable.

@Airidas36  — non-blocking for the core fix itself, but would want these addressed before merge.

logbook-spring-boot-ecs-autoconfigure/src/test/java/org/zalando/logbook/ecs/autoconfigure/LogbookEcsAutoConfigurationTest.java
```diff
@@ -29,6 +31,28 @@ void shouldBackOffWhenSinkBeanExists() {
                 });
     }
 
+    @Test
+    void shouldInitializeEcsSinkWhenStatusCodeBasedWriteDisabled() {
```
kasmarian: How about a test when logbook.write.status-code-based is absent?

Airidas36: Added a test for absent property

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
```diff
@@ -308,21 +306,19 @@ public Strategy withoutBody() {
 
     @API(status = INTERNAL)
     @Bean
-    @ConditionalOnMissingBean(Sink.class)
```
kasmarian, line 311: With dropping `@ConditionalOnMissingBean` custom Sink implementations may now break. 2 new beans should still have this condition I think.

kasmarian: I'm curious why we didn't have a test for this. Would be good to add testing with a user-defined Sink, with status-code logging both enabled and disabled.

logbook-core/src/main/java/org/zalando/logbook/core/LevelBasedHttpLogWriter.java
```diff
@@ -1,38 +0,0 @@
...
-import org.zalando.logbook.Correlation;
-import org.zalando.logbook.HttpLogWriter;
-import org.zalando.logbook.Logbook;
-import org.zalando.logbook.Precorrelation;
-
-import static org.apiguardian.api.API.Status.EXPERIMENTAL;
-
-@API(status = EXPERIMENTAL)
-@RequiredArgsConstructor
-public final class LevelBasedHttpLogWriter implements HttpLogWriter {
```
kasmarian, line 17: A compatibility concern: this removes `StatusCodeBasedSink` and `LevelBasedHttpLogWriter`, and changes `EcsSink` from a class into an interface. Existing users may still use these public types, so this would be a breaking change. 

Could we keep the old classes, mark them deprecated, and add the new implementations alongside them? We could also keep `EcsSink` as a class and use a separate shared base/helper for the new ECS sinks. This would give us the new behavior without breaking existing users.

logbook-core/src/main/java/org/zalando/logbook/core/AbstractHttpStatusCodeSink.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.apiguardian.api.API;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.spi.LoggingEventBuilder;
+
+import static org.apiguardian.api.API.Status.EXPERIMENTAL;
+
+@API(status = EXPERIMENTAL)
+public abstract class AbstractHttpStatusCodeSink implements Sink {
```
kasmarian, line 13: Could `AbstractHttpStatusCodeSink` live in `logbook-core` instead of `logbook-api`? It contains logging implementation logic rather than a basic Logbook API contract. Adding it to `logbook-api` also creates a new public support surface. If third-party sinks are expected to extend it, keeping it there as `EXPERIMENTAL` is reasonable, but that should be an intentional API decision.

logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
Airidas36: Added tests of backoff behavior when user defined beans exist both in `ecs` related AutoConfiguration tests as well as `HttpStatusCodeBasedSink` tests. Additionally added some tests for `ChunkingSink` bean creation

logbook-core/src/main/java/org/zalando/logbook/core/AbstractHttpStatusCodeSink.java
Airidas36: Moved to `logbook-core`

logbook-core/src/main/java/org/zalando/logbook/core/LevelBasedHttpLogWriter.java
Airidas36: Brought back `StatusCodeBasedSink` and `LevelBasedHttpLogWriter` classes and reverted `EcsSink` to maintain compatibility. Marked these classes with `@Deprecated`, with references to functionally equivalent new implementations that will supersede them. Added the `EcsSinkSupport` interface to share common functionality across `DefaultEcsSink` and `HttpStatusCodeBasedEcsSink`.

kasmarian:
👍

logbook-core/src/main/java/org/zalando/logbook/core/LevelBasedHttpLogWriter.java
```diff
@@ -12,6 +12,10 @@
 
 import static org.apiguardian.api.API.Status.EXPERIMENTAL;
 
+/**
+ * @deprecated For dynamic level logging custom sink can be provided, for example {@link AbstractHttpStatusCodeSink}
```
belushkin: Hey @Airidas36 , is it possible to reference `HttpStatusCodeBasedSink` instead of `AbstractHttpStatusCodeSink` in the comment section here? The reason is because  `AbstractHttpStatusCodeSink` is an abstract class and has a `@API(status = INTERNAL)` while `HttpStatusCodeBasedSink` has similar status as `LevelBasedHttpLogWriter`.

belushkin:
:+1:

logbook-core/src/main/java/org/zalando/logbook/core/LevelBasedHttpLogWriter.java
Airidas36: Updated javadoc description and reference

kasmarian:
👍

belushkin:
:+1:

Issue #2370, opened 2026-08-11
Improve out of the box configuration for Spring Boot WebFlux applications with Micrometer tracing


Airidas36:
## Detailed Description
On a Spring Boot WebFlux application running on Reactor Netty (the default server), Logbook's autoconfiguration installs its loggers as Netty channel handlers at the transport layer for both inbound (server) and outbound (client) traffic. In both directions this causes problems when Micrometer tracing is in use.

### Server side (_`logbookNettyServerCustomizer`_ → `LogbookServerHandler`)

1. No traceId on the incoming-request log. The Netty channel handler logs the request before Spring's `ServerHttpObservationFilter`  has created the observation/span, so the `origin:remote`, `type:request` entry is emitted with an empty trace field. The response entry, logged later, does contain the traceId.

2. Error/validation responses are not logged at all. When a handler throws and the response is produced by an error handler `@RestControllerAdvice`, that response never flows back through the Netty pipeline in the correlatable form `LogbookServerHandler` expects, so the exchange is never logged.

The alternative `LogbookWebFilter` avoids both problems (it runs after the observation filter and above error handling), but it is only registered when Reactor Netty is absent `@ConditionalOnMissingClass("reactor.netty.http.server.HttpServer")`. There is no supported way to opt into it on Netty - the only lever is `logbook.filter.enabled: false`, which disables both autoconfigured server loggers and forces users to hand-register their own `LogbookWebFilter` bean.

### Client side (_`logbookNettyClientCustomizer`_ → `LogbookClientHandler`)
No traceId in the MDC of downstream request/response logs. The customizer adds `LogbookClientHandler` directly to the outbound Netty channel without restoring the tracing context, so client-side Logbook entries are logged without traceId in the MDC.
To get traceId into client log entries, users currently have to wrap `LogbookClientHandler` in a custom channel handler that calls `ContextSnapshotFactory.setThreadLocalsFrom(channel)` before delegating. For example:
```
static class TracingChannelHandler extends ChannelDuplexHandler {

        ChannelDuplexHandler delegate;
        ContextSnapshotFactory contextSnapshotFactory;

        TracingChannelHandler(ChannelDuplexHandler delegate, ContextSnapshotFactory contextSnapshotFactory) {
            this.delegate = delegate;
            this.contextSnapshotFactory = contextSnapshotFactory;
        }

        @Override
        public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
[...]

PR #2380, opened 2026-08-28
Allow logging reactive server requests via LogbookWebFilter

belushkin:
Partially addresses #2370 (server side).

Reactive server logging currently uses a Reactor Netty handler that runs before the Spring and Micrometer context is available. As a result, Logbook entries cannot be correlated with application logs using the trace ID.

This PR adds an opt-in WebFlux integration:

```yaml
logbook:
  reactive:
    server-mode: web-filter

spring:
  reactor:
    context-propagation: auto
```

The default remains `netty`, so existing applications are unaffected. In `web-filter` mode, Logbook runs inside the WebFlux pipeline and its request and response entries can use the same MDC trace ID as the application.

The WebFilter runs before Spring Security so rejected requests are still logged. It also captures responses handled inside the controller pipeline, including `@RestControllerAdvice`; global `WebExceptionHandler` and default error responses are not guaranteed.

The configuration is typed, invalid values fail at startup, and exactly one server integration is registered. Tests cover mode selection, invalid configuration, Spring Security 401 responses, controller advice, and MDC trace correlation.

The client-side part of #2370 is intentionally left for a separate change.


kasmarian:
👍

belushkin:
:+1:

PR #2381, opened 2026-08-29
fix(spring): gate secureLogbookFilter on SecurityFilterChain bean

### LINKED ISSUE #2369: secureLogbookFilter activates on classpath presence of spring-security-web alone, duplicating every 401/403 log in apps where no filter chain rejects requests
## Description
`LogbookAutoConfiguration.JakartaSecurityServletFilterConfiguration` registers `secureLogbookFilter` based on `@ConditionalOnClass(SecurityFilterChain.class)` with `logbook.secure-filter.enabled` defaulting to `true`. The condition is classpath presence, not the presence of a security filter chain that rejects requests.

#656 established the design contract of the two-filter setup: it works "as long as the only one producing 401/403 responses is a filter in between them", and for other scenarios "this setup is not intended and won't work". I am not re-reporting that behavior. The problem is that the auto-configuration enables the setup in applications where that precondition provably cannot hold:

- Apps that have spring-security-web on the classpath only transitively, for example through `spring-boot-starter-oauth2-client` used for outbound client-credentials tokens. There is no filter chain at all, or only Boot's default one.
- Apps with a permit-all chain (`anyRequest().permitAll()`) that do authorization inside the dispatcher (method security, AOP, or controller logic) and render 401/403 through MVC exception handling.

In both cases every 401/403 returns normally through both filters, and both write a full request/response pair. This is increasingly the common case in modern Spring Boot services, where spring-security arrives on the classpath for reasons unrelated to servlet request security. The failure mode is silent: error-log volume for auth failures doubles, and nothing warns about it.

The duplicated pair is also written from `Ordered.HIGHEST_PRECEDENCE + 1`, outside any MDC-populating filters an app registers, so the duplicates lack request-context fields. In our case this initially looked like a client retry combined with an MDC bug, which made it expensive to diagnose.

## Expected Behavior
One request/response pair per 401/403 response. `secureLogbookFilter` only activates when its design precondition can hold, or at minimum the constraint is documented on `logbook.secure-filter.enabled`.

## Actual Behavior
Two full request/response pairs per 401/403, with different Logbook correlation IDs and nested durations on the same thread (inner pair from `logbookFilter`, outer pair from `secureLogbookFilter`). Responses with other error statuses (400, 404, 5xx) log once, which makes the duplication look status-dependent and hard to attribute.

## Possible Fix
1. Gate the registration on an actual chain: `@ConditionalOnBean(SecurityFilterChain.class)` on the `secureLogbookFilter` bean method (the configuration already has `@AutoConfigureAfter(SecurityFilterAutoConfiguration)`). This removes the false positive for the transitive-classpath case. It does not cover permit-all chains, which cannot be detected statically, so:
2. Document on `logbook.secure-filter.enabled` (and in the README section on Spring Security) that the two-filter setup requires all 401/403 production to happen in a filter between the two registrations, and that apps which produce 401/403 inside the dispatcher must set the property to `false`.
3. Alternatively, coordination between the two filters through a request attribute would make the setup correct in all cases. I understand from #656 that this was deliberately avoided, so I list it only for completeness.

I am happy to send a PR for option 1 plus the documentation if you agree with the direction.

## Steps to Reproduce
1. Spring Boot servlet app with `logbook-spring-boot-starter` and `spring-boot-starter-oauth2-client` (or any dependency that brings spring-security-web transitively). No `SecurityFilterChain` is declared, or a permit-all one.
2. Configure:
     ```properties
     logbook.filter.enabled=true
     logbook.strategy=status-at-least
     logbook.minimum-status=400
     logging.level.org.zalando.logbook.Logbook=trace
     ```
3. Add a controller or `@ControllerAdvice` that returns 401 or 403, for example a method-level authorization check that throws and is mapped to 401.
4. Call the endpoint. The log contains two request/response pairs with different `correlation` values. Call an unknown path for comparison: the 404 logs one pair.

## Context
Found in production services where every authorization failure was logged twice, doubling error-log volume. Several services in our fleet have the same combination (Logbook starter plus transitive spring-security-web plus in-dispatcher authorization), and each has to discover `logbook.secure-filter.enabled=false` independently. The property works as a fix, but the constraint that makes it safe is not documented anywhere.

## Your Environment
* Version used: 4.1.0
* Spring Boot 4.1.0, servlet stack (Tomcat), Java 25
* Link to your project: internal service, cannot share; reproduction steps above are self-contained

arimu1:
## Summary

- Add `@ConditionalOnBean(SecurityFilterChain.class)` to the `secureLogbookFilter` bean registration so the filter is not auto-enabled when `spring-security-web` is only present transitively on the classpath without an actual security filter chain.
- Document the two-filter setup precondition on `logbook.secure-filter.enabled` in configuration metadata and the README (apps that produce 401/403 in the dispatcher must set the property to `false`).
- Rewrite `SecurityFilterTest` to use `WebApplicationContextRunner`, covering both the presence and absence of a `SecurityFilterChain` bean.

Fixes #2369

## Test plan

- [x] `./mvnw test` in `logbook-spring-boot-autoconfigure` (99 tests, all passing)
- [x] `SecurityFilterTest` verifies `secureLogbookFilter` is registered when a `SecurityFilterChain` bean exists
- [x] `SecurityFilterTest` verifies `secureLogbookFilter` is not registered without a `SecurityFilterChain` bean

### DIFF
--- a/README.md
+++ b/README.md
@@ -1010,7 +1010,7 @@ The `logbook-spring` module contains a `ClientHttpRequestInterceptor` to use wit
 Logbook comes with a convenient auto configuration for Spring Boot users. It sets up all of the following parts automatically with sensible defaults:
 
 - Servlet filter
-- Second Servlet filter for unauthorized requests (if Spring Security is detected)
+- Second Servlet filter for unauthorized requests (if a `SecurityFilterChain` bean is present)
 - Header-/Parameter-/Body-Filters
 - HTTP-/JSON-style formatter
 - Logging writer
@@ -1093,7 +1093,7 @@ The following tables show the available configuration (sorted alphabetically):
 | `logbook.obfuscate.replacement`          | A value to be used instead of an obfuscated one                                                                                                                                                                     | `XXX`              |
 | `logbook.predicate.include`              | Include only certain paths and methods (if defined)                                                                                                                                                                 | `[]`               |
 | `logbook.predicate.exclude`              | Exclude certain  paths and methods  (overrides `logbook.predicate.include`)                                                                                                                                         | `[]`               |
-| `logbook.secure-filter.enabled`          | Enable the [`SecureLogbookFilter`](#servlet)                                                                                                                                                                        | `true`             |
+| `logbook.secure-filter.enabled`          | Enable the [`SecureLogbookFilter`](#security). Only registered when a `SecurityFilterChain` bean exists. Requires all 401/403 responses to be produced by a filter between the two Logbook filters; set to `false` if authorization failures are handled in the dispatcher (e.g. permit-all chains with method security)                                                                                                                                                                        | `true`             |
 | `logbook.strategy`                       | [Strategy](#strategy) (`default`, `status-at-least`, `body-only-if-status-at-least`, `without-body`)                                                                                                                | `default`          |
 | `logbook.write.chunk-size`               | Splits log lines into smaller chunks of size up-to `chunk-size`.                                                                                                                                                    | `0` (disabled)     |
 | `logbook.write.status-code-based`        | Enables status-code-aware log levels. Responses with 2xx/3xx are logged at `TRACE`, 4xx at `WARN`, 5xx at `ERROR`. Requests are always logged at `TRACE`.                                                          | `false`            |
--- a/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
+++ b/logbook-spring-boot-autoconfigure/src/main/java/org/zalando/logbook/autoconfigure/LogbookAutoConfiguration.java
@@ -470,6 +470,7 @@ public class LogbookAutoConfiguration {
         private static final String FILTER_NAME = "secureLogbookFilter";
 
         @Bean
+        @ConditionalOnBean(SecurityFilterChain.class)
         @ConditionalOnProperty(name = "logbook.secure-filter.enabled", havingValue = "true", matchIfMissing = true)
         @ConditionalOnMissingBean(name = FILTER_NAME)
         public FilterRegistrationBean<?> secureLogbookFilter(final Logbook logbook) {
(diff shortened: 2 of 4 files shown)


prashantpiyush1111 (approved):
LGTM. Verified independently — confirmed the enclosing JakartaSecurityServletFilterConfiguration already has @AutoConfigureAfter(SecurityFilterAutoConfiguration), so the new @ConditionalOnBean(SecurityFilterChain.class) correctly sees Spring Boot's own security autoconfiguration after it's run (avoids the classic ordering pitfall with cross-autoconfig @ConditionalOnBean checks). Also checked that this codebase targets Spring Security 6.x (jakarta.servlet namespace), so gating on SecurityFilterChain doesn't break anyone still on the deprecated WebSecurityConfigurerAdapter style — that API isn't even available in this version range. Both new tests (shouldRegisterSecureLogbookFilterWhenSecurityFilterChainIsPresent / shouldNotRegisterSecureLogbookFilterWithoutSecurityFilterChain) directly cover the regression being fixed. Docs/metadata updates accurately reflect the new gating behavior. No concerns

kasmarian:
👍

kasmarian:
@arimu1 the PR has conflicts now, could you resolve?

arimu1:
@kasmarian Conflicts are resolved — rebased `fix-2369-secure-logbook-filter` onto current `main` (tip `1e246268`). The only conflict was in `README.md`: kept upstream's new predicate/reactive config rows and preserved the `SecurityFilterChain` / `logbook.secure-filter.enabled` documentation from this PR. `@ConditionalOnBean(SecurityFilterChain.class)` on `secureLogbookFilter` is unchanged.

arimu1:
Post-#2345, `SecurityFilterTest` now registers `AsyncOnCompleteListenerWrapper` (identity) on the isolated `WebApplicationContextRunner` so `secureLogbookFilter` can satisfy its constructor dependency. `SecurityFilterTest`: 2/2 pass locally (JDK 21).

PR #2385, opened 2026-09-02
Add AnsiPrettyPrintingSink for colored JSON logging (#2354)

### LINKED ISSUE #2354: Pretty-printing with ANSI coloring for local testing
Add option to pretty-print with ANSI coloring. 

## Detailed Description
Add syntax highlighting to pretty-printing output, both headers and the (JSON) payload itself. 

Image

## Context
Improve readability for log statements printed during local development. Possibly in a new artifact which is purely intended for testing purposes.

## Possible Implementation
Add a [Sink](https://github.com/entur/cloud-logging/blob/main/request-response/logbook-test/src/main/java/no/entur/logging/cloud/logbook/logbook/test/PrettyPrintingSink.java)  which both perform both pretty-printing and syntax highlighting.

Vinu2111:
Implements the ANSI-coloring pretty-print sink requested in #2354.

Adds `AnsiPrettyPrintingSink` and a small `AnsiColors` helper to `logbook-test`. It pretty-prints request/response JSON with headers in cyan, JSON keys in green, and JSON string values in yellow meant purely for local development readability, not production use.

Color output auto-detects via `System.console() != null`, or can be forced on/off via the constructor.

**Known limitation:** coloring is applied via regex post-processing of the already-formatted JSON string, rather than during serialization. This keeps the implementation simple and dependency-light, but could mis-highlight edge cases with unusual nested quote patterns in body content. Happy to revisit with a proper Jackson-generator-level approach if maintainers prefer that.

Closes #2354

### DIFF
--- a/test-modules/logbook-test/pom.xml
+++ b/test-modules/logbook-test/pom.xml
@@ -25,6 +25,11 @@
             <artifactId>logbook-api</artifactId>
         </dependency>
         <dependency>
+            <groupId>tools.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>${jackson3.version}</version>
+        </dependency>
+        <dependency>
             <groupId>org.zalando</groupId>
             <artifactId>faux-pas</artifactId>
             <version>0.9.0</version>
--- a/test-modules/logbook-test/src/main/java/org/zalando/logbook/test/AnsiColors.java
+++ b/test-modules/logbook-test/src/main/java/org/zalando/logbook/test/AnsiColors.java
@@ -1,0 +1,29 @@
+package org.zalando.logbook.test;
+
+/**
+ * Utility class for ANSI color codes to format console output.
+ * Stateless, containing only constants and static methods.
+ */
+public final class AnsiColors {
+
+    public static final String RESET = "\u001B[0m";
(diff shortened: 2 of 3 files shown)


Vinu2111:
Hi @kasmarian, looks like the COMPR/pr-labels check needs one of the required labels. I don't have permission to set it myself  could you add minor when you get a chance? (Net-new, backwards-compatible, opt-in dev utility, so that felt like the right fit happy to hear otherwise.)

lukasniemeier-zalando:
Is that a module that one can integrate inton their local developer environment without the need to have it as a module that we offer here from the logbook repository?

I feel it's cool and cute, but not sure if it's worth adopting to the main logbook project :heart:

Vinu2111:
Fair question. A couple of reasons I leaned toward including it here rather than as a local snippet:

- It's scoped entirely to `logbook-test`, so it adds zero footprint to anything shipped in production.
- The original issue linked an implementation another team (Entur) already built for the same purpose - having one canonical version in logbook-test avoids every team re-rolling their own.

But totally hear the "is this really core-repo material" concern, it's a small convenience feature, not a fix. If the consensus leans toward not adopting it, I'm happy to close this out, no worries either way. Let me know what you and the others think.

PR #2388, opened 2026-09-05
fix: honour the log level in EcsSink

### LINKED ISSUE #2387: High CPU usage after upgrading to 4.0.4+
## Description

We observed a large increase in CPU usage after upgrading Logbook from version 4.0.3 to 4.0.4 (or 4.1).
As of version 4.0.4, Logbook automatically includes `logbook-spring-boot-ecs-autoconfigure`.
This auto-enables `EcsSink` when using ECS logging in Spring Boot (`logging.structured.format.console: ecs`).

`EcsSink` does not override `Sink#isActive()`, so it inherits the interface default `return true`. 
Because `DefaultLogbook#process` uses `sink.isActive()` as its short-circuit, enabling ECS structured logging silently disables that short-circuit. Every request and response body is then buffered, stringified and formatted into an ECS map on every single call and only then handed to `LOGGER.atTrace()`, which drops it.

The result: setting `logging.level.org.zalando.logbook.Logbook: OFF` does **not** turn Logbook off when
ECS logging is enabled. It only stops the output. All the cost remains.

This is a behavioral difference from `DefaultSink`, which does honor the level, and it is not
mentioned in the ECS autoconfiguration docs.

## Expected Behavior

With `logging.level.org.zalando.logbook.Logbook: OFF`, Logbook performs no per-request work,
regardless of which sink is configured, exactly as it already behaves with `DefaultSink`.

## Actual Behavior

With ECS enabled (`logging.structured.format.console: ecs`) and the very same logger set to `OFF`,
Logbook buffers and formats every body, then throws the result away.

## Possible Fix

One line in `EcsSink`, mirroring `DefaultHttpLogWriter`:

```java
@Override
public boolean isActive() {
    return LOGGER.isTraceEnabled();
}
```

This restores the short-circuit and makes `EcsSink` behave like `DefaultSink` with respect to the log
level.

## Steps to Reproduce

`DefaultLogbook#process`, `sink.isActive()` is the guard:

```java
public RequestWritingStage process(final HttpRequest originalRequest, final Strategy strategy) throws IOException {
    if (sink.isActive() && predicate.test(originalRequest)) {
        final Precorrelation precorrelation = newPrecorrelation(originalRequest);
        final HttpRequest processedRequest = strategy.process(originalRequest);    // -> request.withBody()
        ...
    } else {
        return Stages.noop();    // nothing is buffered
    }
}
```

`Strategy#process(HttpRequest)` defaults to `return request.withBody();`, and `RemoteRequest#withBody()` (logbook-servlet) is `state.updateAndGet(State::with); buffer();` an eager, full copy of the body.
So `isActive()` is the only thing standing between a request and a complete buffer of its body.

The two sinks answer that question differently:

`DefaultSink` honors the level:

```java
@Override
public boolean isActive() {
    return writer.isActive();    // DefaultHttpLogWriter -> log.isTraceEnabled()
}
```

`EcsSink` declares no `isActive()` at all, so it inherits
`Sink`'s `default boolean isActive() { return true; }`:

```java
public class EcsSink implements Sink {

    private static final Logger LOGGER = LoggerFactory.getLogger(Logbook.class);

    private final StructuredHttpLogFormatter structuredHttpLogFormatter;

    // no isActive() override -> always active

    @Override
    public void write(Precorrelation precorrelation, HttpRequest httpRequest) throws IOException {
        Map<String, Object> content = structuredHttpLogFormatter.prepare(precorrelation, httpRequest);
        write(content);
    }

    private void write(Map<String, Object> content) throws IOException {
        LoggingEventBuilder loggingEventBuilder = LOGGER.atTrace();    // level checked only here
        content.forEach(loggingEventBuilder::addKeyValue);
        loggingEventBuilder.log(structuredHttpLogFormatter.format(content));
    }
}
```

Note that `EcsSink` guards on the *same* logger name (`org.zalando.logbook.Logbook`) that
`DefaultHttpLogWriter` uses. The level check exists, it is simply applied after all the work is
already done.

By the time SLF4J drops the event, `EcsStructuredHttpLogFormatter#prepare` has already buffered the
body, called `getBodyAsString()`, called `body.getBytes(UTF_8).length`, built a ~20-entry `HashMap`,
and run three `String.replaceAll` calls plus a `String.format` for every header.

## Context
Roughly 40% CPU increase.

## Your Environment
* Version used: 4.0.3 (unaffected), 4.0.4 (affected) and 4.1.0 (affected)

timpeeters:
Fixes #2387

## Description

EcsSink did not override `Sink#isActive()`, so it inherited the interface default `return true`. `DefaultLogbook#process` uses `sink.isActive()` as its only short-circuit, and `Strategy#process` defaults to `request.withBody()`, which eagerly buffers the body.

The consequence is that with ECS structured logging enabled, setting `logging.level.org.zalando.logbook.Logbook` below TRACE suppresses the output but not the work: every request and response body is still buffered, stringified and formatted into an ECS map, and only then handed to LOGGER.atTrace(), which discards it. Measured on a Spring Boot app, a 100 KB body POSTed to a handler that never reads it: 0 bytes buffered with DefaultSink, 102400 bytes with EcsSink, logger OFF in both cases.

TRACE is the correct level to check: `EcsSink` writes via `LOGGER.atTrace()` on `LoggerFactory.getLogger(Logbook.class)`, the same logger and level that `DefaultHttpLogWriter#isActive` already guards on. The new test mirrors `DefaultHttpLogWriterTest#shouldUseTraceLevelForActivation`, which required adding logback-classic to this module's test scope.

## Motivation and Context
Fix high CPU usage after upgrading from 4.0.3 to 4.0.4 (or 4.1).

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [X] All commits are [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification)

### DIFF
--- a/logbook-spring-boot-ecs-autoconfigure/pom.xml
+++ b/logbook-spring-boot-ecs-autoconfigure/pom.xml
@@ -59,6 +59,11 @@
             <artifactId>spring-boot-test</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/logbook-spring-boot-ecs-autoconfigure/src/main/java/org/zalando/logbook/ecs/EcsSink.java
+++ b/logbook-spring-boot-ecs-autoconfigure/src/main/java/org/zalando/logbook/ecs/EcsSink.java
@@ -23,6 +23,11 @@ public class EcsSink implements Sink {
     private final StructuredHttpLogFormatter structuredHttpLogFormatter;
 
     @Override
+    public boolean isActive() {
+        return LOGGER.isTraceEnabled();
+    }
+
+    @Override
     public void write(Precorrelation precorrelation, HttpRequest httpRequest) throws IOException {
         Map<String, Object> content = structuredHttpLogFormatter.prepare(precorrelation, httpRequest);
         write(content);
--- a/logbook-spring-boot-ecs-autoconfigure/src/test/java/org/zalando/logbook/ecs/EcsSinkTest.java
+++ b/logbook-spring-boot-ecs-autoconfigure/src/test/java/org/zalando/logbook/ecs/EcsSinkTest.java
@@ -1,0 +1,35 @@
+package org.zalando.logbook.ecs;
+
+import ch.qos.logback.classic.Logger;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+import org.slf4j.LoggerFactory;
+import org.zalando.logbook.Logbook;
+import org.zalando.logbook.Sink;
+
+import static ch.qos.logback.classic.Level.INFO;
+import static ch.qos.logback.classic.Level.TRACE;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+final class EcsSinkTest {
+
+    private final Logger logger = (Logger) LoggerFactory.getLogger(Logbook.class);
+
+    @AfterEach
+    void reset() {
+        logger.setLevel(null);
+    }
+
+    @Test
+    void shouldUseTraceLevelForActivation() {
+        final Sink unit = new EcsSink(content -> "");
+
+        logger.setLevel(TRACE);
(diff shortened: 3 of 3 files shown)


kasmarian:
Thank you for your contribution! We'll get it into the next release

kasmarian:
👍

lukasniemeier-zalando (approved):

PR #2390, opened 2026-09-05
Make FastJsonHttpLogFormatter honor the JsonMapper's ObjectWriteContext

### LINKED ISSUE #2389: FastJsonHttpLogFormatter creates an empty ObjectWriteContext instead of inheriting from JsonMapper
`FastJsonHttpLogFormatter` creates an empty `ObjectWriteContext` instead of inheriting from `JsonMapper`. This leads calls to `jsonGenerator.writePOJOProperty(...)` to fail.

## Description
`FastJsonHttpLogFormatter` is currently relying on `JsonFactory` and `ObjectWriteContext.empty()` to create the `JsonGenerator` in [here](https://github.com/zalando/logbook/blob/9a0cc22343f749ed8065ffc3ea9c7e017c4670b8/logbook-json/src/main/java/org/zalando/logbook/json/FastJsonHttpLogFormatter.java#L92).

If you implement your own `JsonFieldWriter` and want to make use of `jsonGenerator.writePOJOProperty(...)`, it will fail with the following error:
> Operation not supported by ObjectWriteContext of type tools.jackson.core.ObjectWriteContext$Base

## Expected Behavior
Calls to `jsonGenerator.writePOJOProperty(...)` to succeed, because the `ObjectWriteContext` in my `JsonMapper` supports it.

## Actual Behavior
Calls to `jsonGenerator.writePOJOProperty(...)` fail with the following error:
> Operation not supported by ObjectWriteContext of type tools.jackson.core.ObjectWriteContext$Base

## Possible Fix
Change the `JsonGenerator` creation to be done like this instead:
```java
try (final JsonGenerator generator = jsonMapper.createGenerator(writer)) {
...
}
```

I have created a direct replacement of `FastJsonHttpLogFormatter` with just this change and it seems to work as expected. I also believe this is how it is supposed to be done in `Jackson 3`.

## Steps to Reproduce
Create a `JsonFieldWriter` that makes use of `jsonGenerator.writePOJOProperty(...)` and force it to be executed.

## Context
We were trying to migrate all our projects to `Jackson 3` and we had our custom `JsonFieldWriter` writing some things as POJO, like the timestamp (which is an `Instant`) or a `Map` with the headers. We couldn't migrate to `Jackson 3` making `Logbook` to work as it used to for us.

nhmarujo:
Fixes #2389


logbook-json/src/main/java/org/zalando/logbook/json/FastJsonHttpLogFormatter.java
```diff
@@ -44,21 +42,12 @@ public FastJsonHttpLogFormatter(final JsonMapper mapper) {
...
-        this(factory, delegate, false);
-    }
-
-    public FastJsonHttpLogFormatter(final JsonFactory factory, final JsonFieldWriter delegate, final boolean validateJsonBody) {
-        this.factory = factory;
-        this.delegate = delegate;
-        this.validateJsonBody = validateJsonBody;
+        this.jsonMapper = mapper;
+        this.delegate = writer;
     }
```
kasmarian, line 62: Is there a way to still preserve the constructors that use JsonFactory for those who'd prefer to use it? I'm trying to think if a breaking change here is necessary.

nhmarujo: Good point. I added a commit to reinstate the constructor that takes a `JsonFactory`.

I didn't add back the constructor with signature `public FastJsonHttpLogFormatter(final JsonFactory factory, final JsonFieldWriter delegate, final boolean validateJsonBody)` though, as I don't think it made sense even before. `validateJsonBody` was necessary purely to create `DefaultJsonFieldWriter` and didn't have any use in this particular constructor.

Please let me know your thoughts.

nhmarujo: @kasmarian any thougths?

belushkin: 1. Is it possible to preserve old constructor (with `final boolean validateJsonBody`) for the backward compatibility and call a new constructor inside of the old one?

2. This change also makes the formatter honour  `SerializationFeature.INDENT_OUTPUT` . If an application passes a pretty-printing  `JsonMapper` , logs will change from single-line to multiline JSON, which may break line-based log processing.

kasmarian: Makes sense. Seems like validateJsonBody on this formatter was never needed.

kasmarian (approved):

logbook-json/src/main/java/org/zalando/logbook/json/FastJsonHttpLogFormatter.java
nhmarujo: Thanks @kasmarian 

@belushkin it would definitely be possible to reinstate the constructor with the following signature:
```java
public FastJsonHttpLogFormatter(final JsonFactory factory, final JsonFieldWriter delegate, final boolean validateJsonBody)
```

I deliberately left it out because it wasn't doing anything (even before). `validateJsonBody` is not used for anything unless you are creating a `DefaultJsonFieldWriter`, so I thought it would be better to just leave it out and people could just rely on:
```java
public FastJsonHttpLogFormatter(final JsonFactory factory, final JsonFieldWriter delegate)
```

That said, if majority prefers, I can reinstate it and mark it as `@Deprecated`. Feedback is welcome.

As for the `SerializationFeature.INDENT_OUTPUT`, I actually think is great to have much more control over how things will get printed. Is it up to whoever passes the `JsonMapper` to decide what they need. This way the lib is less opinionated on the final output 🙂

belushkin (approved):

kasmarian:
ah, the commits are not signed @nhmarujo, could you make sure they are signed?

nhmarujo:
@kasmarian sorry, I totally missed that. Will fix a bit later once I get back home 🙂 thanks for the heads up

nhmarujo:
Commits are properly signed now. @kasmarian / @belushkin do you mind taking a second look, please?

kasmarian (approved):

belushkin (approved):

PR #2401, opened 2026-09-17
feat(spring): separate server and client predicates

### LINKED ISSUE #2015: Zalando Logbook separate configuration from client and server side
Is it possible to configure Zalando Logbook separately for the **server side** logging (incoming requests) and the **client side** logging (outgoing requests)?

I'd like to log only some incoming requests (`/api/**` and `/int/**`), but all the outgoing requests not regarding their URI (`/**`).

Following the documentation https://gitee.com/kedocode/logbook#configuration, something like:

```yaml
logbook:
  server:
    predicate:
      include:
        - path: /api/**
        - path: /int/**
  client:
    predicate:
      include:
        - path: /**
```

The above configuration was the solution suggested by ChatGPT, but it doesn't seem to work, however the idea was good.

kasmarian:
## Description

Add separate `logbook.server.predicate` and `logbook.client.predicate` configuration for incoming and outgoing requests. Existing global predicates remain applied to both origins.

## Motivation and Context

Closes #2015.

This allows applications to filter incoming and outgoing traffic independently without implementing a custom origin-aware predicate.

## Types of changes

## Checklist:


belushkin:
:+1:

kasmarian:
👍

